PR #3, opened 2015-10-29
request and response classes and related hooks of old client protocol are deleted

sancar:


ihsandemir:
:+1:  We shall commit the actual protocol changes on top of this PR.

PR #10, opened 2015-11-26
Fixing release scripts

sancar:
HAZELCAST_LIBRARY_TYPE is removed since it is duplicate.
HZ_LIB_TYPE is used where needed. Release and test scripts
are made sync.


ihsandemir:
:+1:

PR #11, opened 2015-11-30
New client protocol implementation

ihsandemir:
Implemented the new client protocol. Implemented the Map putAll/getAll changes. Added the predicate implementation. Removed DataSerializable.

The two commits shall be merged into a single commit after review.


ihsandemir:
verify

sancar:
There are classes related to old protocol. Can you clean them up ? 
And also I see other unused classes, parameters and duplicate methods. Can you take another look to your pr with this in mind ?

sancar:
I see std::auto_ptr<Data> usages. Data itself is using auto_ptr inside. When it is copied it will transfer the ownership. We don't need another std::auto_ptr around it.

sancar:
There are two new flags in CMake HZ_PLATFORM_WINDOWS and HZ_PLATFORM_DARWIN. 
 HZ_PLATFORM_WINDOWS is not used in codebase. And HZ_PLATFORM_DARWIN. can be removed with a couple of casts where needed(DataEntryViewCodec).

sancar:
There are two methods for getTime in TransactionOptions class. One is introduced in this pr. It is better to make the conversion where needed instead of introducing a new method. We may add a util method for conversion if there lots of usages.

sancar:
We should move common::container classes to util namespace. All of our utility classes are there. ManagedArray is just re-implementation of std::vector. And ManagedArray is not used, we can remove it.

ihsandemir:
verify

ihsandemir:
verify

ihsandemir:
verify

sancar:
:+1:

PR #13, opened 2015-12-10
Uncommented the test code and resend future set correction

ihsandemir:
Uncommented the mistakenly commented out test code and a minor correction on a new test. Eliminated the warning during the invocation resend operations by fixing the promise future into correct state before making the re-invocation of the request.


cangencer:
:+1:

PR #14, opened 2015-12-14
Correlation id size change

ihsandemir:
Changed the correlation id from 4 byte unsigned int to 8 byte unsigned int representation.


sancar:
:+1:

PR #16, opened 2015-12-23
Code samples for C++ - First part

ihsandemir:
Code samples are added. The java code samples repository is used as a reference. This is the initial part of the code samples. The rest shall be added later.

There are also some minor corrections in the library.

### DIFF
--- a/examples/distributed-map/entry-listener/main.cpp
+++ b/examples/distributed-map/entry-listener/main.cpp
@@ -1,0 +1,80 @@
+/*
+ * Copyright (c) 2008-2015, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+//
+// Created by İhsan Demir on 21/12/15.
+//
+#include <hazelcast/client/HazelcastClient.h>
+
+class MyEntryListener : public hazelcast::client::EntryListener<std::string, std::string> {
+
+public:
+    void entryAdded(const hazelcast::client::EntryEvent<std::string, std::string> &event) {
+        std::cout << "[entryAdded] " << event << std::endl;
+    }
+
+    void entryRemoved(const hazelcast::client::EntryEvent<std::string, std::string> &event) {
+        std::cout << "[entryRemoved] " << event << std::endl;
+    }
+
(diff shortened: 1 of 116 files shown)


ihsandemir:
verify

ihsandemir:
verify

ihsandemir:
verify

examples/monitoring/cluster/main.cpp
```diff
@@ -0,0 +1,102 @@
...
+        std::cout << "The following are members in the cluster:" << std::endl;
+        for (std::vector<hazelcast::client::Member>::const_iterator it = members.begin(); it != members.end(); ++it) {
+            std::cout << it->getAddress() << std::endl;
+        }
+        
+        MyMemberListener memberListener;
+        cluster.addMembershipListener(&memberListener);
+        
+        MyInitialMemberListener initialMemberListener;
+        cluster.addMembershipListener(&initialMemberListener);
```
sancar: For the sake of showing correct usage, we should add removeMembershipListener to these tests.
Otherwise it causes a memory leak

examples/distributed-map/entry-listener/main.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+        hazelcast::client::IMap<std::string, std::string> map = hz.getMap<std::string, std::string>("somemap");
+
+        MyEntryListener listener;
+
+        map.addEntryListener(listener, true);
+
+        std::cout << "EntryListener registered" << std::endl;
+
+        // wait for modify map executable to run
```
sancar: Is it clearly documented somewhere that there should be another executable to show this behaviour ?
If not, we can show this behaviour in same process with a thread easily. I think, it will be a more straight-forward example for user. S/he will see all logic in same code.

examples/distributed-map/locking/PessimisticUpdate.cpp
```diff
@@ -0,0 +1,92 @@
...
+        for (int k = 0; k < 1000; k++) {
+            map.lock(key);
+            try {
+                map.get(key);
+                hazelcast::util::sleepmillis(10);
+                v.amount++;
+                map.put(key, v);
+
+                map.unlock(key);
+            } catch (...) {
```
sancar: I think we should not use universal catch in our examples. We should not promote it. Our users should only expect a well defined Exception. All our exceptions extends hazelcast::exception::IException, we should use that instead.

And this code block should have been as follows with `finally`

            map.lock(key);
             try {
                 map.get(key);
                 hazelcast::util::sleepmillis(10);
                 v.amount++;
                 map.put(key, v);
            } finally {
                       map.unlock(key);
             }

examples/distributed-primitives/lock/RaceFree.cpp
```diff
@@ -0,0 +1,77 @@
...
+                    long n2 = number2.get();
+                    if (n1 - n2 != 0) {
+                        std::cout << "Datarace detected at" << k << " !" << std::endl;
+                    }
+                } else {
+                    number1.incrementAndGet();
+                    number2.incrementAndGet();
+                }
+            } catch (...) {
+                lock.unlock();
```
sancar: Who unlocks the lock in successful case ?
Should we have a `finally` instead of `catch`.

PR #19, opened 2016-01-06
Test machine runs out of disk space

ihsandemir:
Added cleanup to test scripts to prevent build machine from running out of disk space. The test related folders are removed when test is successful.


ihsandemir:
We decided to clean the test folders using the jenkins workspace deletion step. Hence, we don't need this PR for now.

PR #20, opened 2016-01-07
Code samples 

ihsandemir:


ihsandemir:
verify

examples/serialization/portable/main.cpp
```diff
@@ -54,7 +54,7 @@ class Person : public hazelcast::client::serialization::Portable {
     std::auto_ptr<std::string> name;
 };
 
-namespace std{
+namespace std {
```
sancar: This usage does not seem right. You don't need to put your `operator<<` in `namespace std`  . There is ADL (Argument Dependent Lookup) just to be able to do this. see
https://en.wikipedia.org/wiki/Argument-dependent_name_lookup

ihsandemir: OK, changed.

sancar:
:+1:

PR #22, opened 2016-01-08
SimpleMapTest throughput is degraded after a while

ihsandemir:
Fixes the performance degradation after a while problem in SimpleMapTest.

### DIFF
--- a/hazelcast/include/hazelcast/util/Atomic.h
+++ b/hazelcast/include/hazelcast/util/Atomic.h
@@ -50,7 +50,7 @@ namespace hazelcast {
                 return v++;
             }
 
-            T &operator++() {
+            T operator++() {
                 LockGuard lockGuard(mutex);
                 return ++v;
             }
@@ -65,7 +65,7 @@ namespace hazelcast {
                 return v;
             }
 
-            T &operator--() {
+            T operator--() {
                 LockGuard lockGuard(mutex);
                 return --v;
             }
--- a/hazelcast/src/hazelcast/client/spi/InvocationService.cpp
+++ b/hazelcast/src/hazelcast/client/spi/InvocationService.cpp
@@ -35,6 +35,8 @@
 #include "hazelcast/client/spi/ServerListenerService.h"
 #include "hazelcast/client/serialization/pimpl/SerializationService.h"
 
+#include <assert.h>
+
 namespace hazelcast {
     namespace client {
         namespace spi {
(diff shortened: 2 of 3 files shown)


sancar:
Can you add the information about cause of the bug to the commit message for future reference ?

hazelcast/include/hazelcast/util/Atomic.h
```diff
@@ -50,7 +50,7 @@ namespace hazelcast {
                 return v++;
             }
 
-            T &operator++() {
+            T operator++() {
```
sancar: I think we also want to change `operator--`for same reason.

hazelcast/include/hazelcast/util/Future.h
```diff
@@ -94,7 +94,11 @@ namespace hazelcast {
                 }
                 time_t endTime = time(NULL) + timeInSeconds;
                 while (!(resultReady || exceptionReady) && endTime > time(NULL)) {
-                    conditionVariable.waitFor(mutex, endTime - time(NULL));
+                    if (!conditionVariable.waitFor(mutex, endTime - time(NULL))) {
+                        std::ostringstream out;
+                        out << "Future::get(timeInSeconds) timed out. Timeout: " << timeInSeconds;
+                        hazelcast::util::ILogger::getLogger().warning(out.str());
```
sancar: I think we should not log here. We will tell the timeout to user via TimeoutException.

hazelcast/src/hazelcast/client/spi/InvocationService.cpp
```diff
@@ -267,7 +275,17 @@ namespace hazelcast {
                 }
 
                 boost::shared_ptr<connection::CallPromise> promise = deRegisterCall(connection, correlationId);
-                if (promise.get() == NULL) {
+                if (NULL == promise.get()) {
```
sancar: If I don't remember wrong, it  is possible to get null from deregister in some cases . But I need to focus on that more to figure it out. Can you reconsider this part if there could be another thread already deregistered in some cases(exception, connection close .. etc) ?

hazelcast/include/hazelcast/util/Atomic.h
ihsandemir: yes, corrected.

hazelcast/src/hazelcast/client/spi/InvocationService.cpp
ihsandemir: I checked on caller paths for deregister, the access to the callpromisemap for the connection is only in 3 places: registerCall, deregisterCall and cleanReources. None of them removes the entry except the deregisterCall. deregisterCall is called from handleMessage. The only way I can think of in which case the promise is synchronousy removed from the map (so that one of the deregisterCall's getting null response) is if the server sends response with the same correlation id two times and this should be an implementation error.

If you still think, this case may occur in production which may cause too many warning messages, I can revert it, and we can open an issue to investigate further.

hazelcast/include/hazelcast/util/Future.h
ihsandemir: Reverted the log.

ihsandemir:
verify

ihsandemir:
verify

sancar:
:+1:

PR #24, opened 2016-01-12
Optional C++11 enabling

ihsandemir:
Adds the HZ_USE_C11 flag for cmake. With this fag the user can compile the library for c++11 if desired. It is experimental for now and it is only an option.


cangencer:
:+1:

PR #37, opened 2016-02-03
CustomSerialized classes don't have to implement getTypeId inside class

sancar:
getTypeId method should be provided as free function. This way, user
can serialize classes that it can not change.

CHECK_NULL in SerializationService header moved to source file
since MACRO's can collide with users MACRO when defined in header.

A example code is provided for custom serialization.

Related test codes are modified for custom serialization.

For threads safety, instance of SerializationConstants is created in
global space.

### DIFF
--- a/hazelcast/src/hazelcast/client/serialization/pimpl/SerializationConstants.cpp
+++ b/hazelcast/src/hazelcast/client/serialization/pimpl/SerializationConstants.cpp
@@ -18,36 +18,16 @@
 //
 
 #include "hazelcast/client/serialization/pimpl/SerializationConstants.h"
+#include "hazelcast/util/Util.h"
+#include "hazelcast/util/ILogger.h"
+#include "hazelcast/client/exception/IClassCastException.h"
 
 namespace hazelcast {
     namespace client {
         namespace serialization {
             namespace pimpl {
-                SerializationConstants *SerializationConstants::instance = NULL;
-
                 SerializationConstants::SerializationConstants()
-                : CONSTANT_TYPE_NULL(0)
-                , CONSTANT_TYPE_PORTABLE(-1)
-                , CONSTANT_TYPE_DATA(-2)
-                , CONSTANT_TYPE_BYTE(-3)
-                , CONSTANT_TYPE_BOOLEAN(-4)
-                , CONSTANT_TYPE_CHAR(-5)
-                , CONSTANT_TYPE_SHORT(-6)
-                , CONSTANT_TYPE_INTEGER(-7)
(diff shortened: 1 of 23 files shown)


hazelcast/src/hazelcast/client/serialization/pimpl/SerializationConstants.cpp
```diff
@@ -18,12 +18,15 @@
...
+#include "hazelcast/util/Util.h"
+#include "hazelcast/util/ILogger.h"
+#include "hazelcast/client/exception/IClassCastException.h"
 
 namespace hazelcast {
     namespace client {
         namespace serialization {
             namespace pimpl {
-                SerializationConstants *SerializationConstants::instance = NULL;
+                SerializationConstants *SerializationConstants::instance = &getInstance();
```
mdogan: why not `SerializationConstants *SerializationConstants::instance = new SerializationConstants();`?

sancar: I was thinking `SerializationConstants`, because constructor is private.But I just tried and I can call the constructor here. It looks like this part counts as inside of SerializationConstants scope. 
And I was taking this link as reference
http://stackoverflow.com/questions/17924688/meyers-singleton-thread-safe-with-c-98

Now reading carefully, the discussion about thread safety is not clear. We may have to use locks here. And double checking lock guarantee is not available in c++98. I can reason that new implementation in this prd safer, but I am not sure if this is safe. See following,
http://www.aristeia.com/Papers/DDJ_Jul_Aug_2004_revised.pdf

Or better way, remove the singleton approach completely.

ihsandemir:
Can you also update http://docs.hazelcast.org/docs/3.6/manual/html-single/index.html#c-client-code-examples?

ihsandemir:
This change shall also cause existing custom serialization users' code to fail (until they add the needed getTypeId function), hence we have to emphasize this in release notes so that they are informed before using the released code.

ihsandemir:
After this change, we can get the type id for any of our serializable objects if we add the implementation of these functions for the primitives as well. This way we can unify the serialization of all types and we can easily support concepts like global serializer as in Java client. We can also solve the problem of getAll returning types of different types. Furthermore, this may even allow us template free API.

ihsandemir:
Let's change method signature from int getHazelcastTypeId(const SomeType &) to int getHazelcastTypeId(const SomeType *) in case we may need to work with null objects. 

Except this all is OK :+1:

sancar:
Ok changing the method signature

PR #42, opened 2016-02-08
Backported fixes from master(#37)

sancar:
CHECK_NULL in SerializationService header moved to source file
since MACRO's can collide with users MACRO when defined in header.

A example code is provided for custom serialization.

Related test codes are modified for custom serialization.

For threads safety, instance of SerializationConstants is created in
global space.

ClassCastException for portable identified and custom serializable
is added.

These fixes are backported from #37

Remove singleton usage of SerializationConstants to avoid thread safety
issues.


ihsandemir:
:+1:

Issue #35, opened 2016-02-01
Usage of released includes and libraries


ihsandemir:
We need tests to make sure that the released zip folder is correct before it is released. For this purpose, we may use the examples built with the to-be-release folder. This is important for detecting any issues with the release in an earlier stage.

PR #44, opened 2016-02-09
Missing verification step in release scripts

ihsandemir:
Added the verification scripts to verify that the released files work correctly. We compile the examples using different versions of the libraries to achieve this. If the builds succeed, then we assume that the release is proper and correct. This will avoid any missing file problem during releases.

Fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/35


sancar:
:+1:

PR #50, opened 2016-02-18
Add release method to obtain raw pointer from shared_ptr

ihsandemir:
Added a utility method to obtain the raw pointer from a shared pointer. It is not a desired way but in case the raw pointer is needed without keeping the shared_ptr this may be a work around.


ihsandemir:
verify

ihsandemir:
This solution causes memory leak due to internal dynamic memory allocations in boost::shared_ptr such as reference count object.

PR #52, opened 2016-02-22
Backport of #53: When directory path includes spaces windows release verification script fails

ihsandemir:
Fix for release script error when there is space in directory path.


sancar:
:+1:

PR #63, opened 2016-02-25
Backport #62 Revert the thread usage for partition refreshing

ihsandemir:
Reverted the thread usage for partition refreshing. It should be handled better by using a mechanism such as ExecutorService. The destructor of the thread was blocking on thread join after performing thread_cancel(cancel was not affecting the update) which caused the update to be blocking anyway. The refresh is being called through the cluster listener thread hence it does not block the IOSelector thread.

Added the additional addresses to java server member for some manual tests.


ihsandemir:
verify

PR #70, opened 2016-03-22
PartitionService shutdown causes deadlock in windows

ihsandemir:
Removed the lock usage in PartitionService shutdown, since it was causing deadlock in Windows environment maintenance nightly builds ( https://hazelcast-l337.ci.cloudbees.com/view/C++/job/cpp_windows%20-%20maintenance%20-%20nightly/65/console ).

Steps that cause deadlock:

1. shutdown gets the lock.
2. refreshPartitions needs the lock to update the partition list before the thread terminates and the deadlock occurs.


ihsandemir:
verify

PR #71, opened 2016-03-22
Backport of PR #70

ihsandemir:
Removed the lock usage in PartitionService shutdown, since it was causing deadlock in Windows environment maintenance nightly builds.


ihsandemir:
verify

Issue #85, opened 2016-04-06
ClientQueueTest.testTake sometimes causes segmentation fault


ihsandemir:
ClientQueueTest.testTake sometimes causes segmentation fault on linux and windows.
Log:
https://hazelcast-l337.ci.cloudbees.com/job/cpp-linux-nightly-64-STATIC-Release/69/console
https://hazelcast-l337.ci.cloudbees.com/job/cpp-linux-nightly-64-SHARED-Release/69/console
https://hazelcast-l337.ci.cloudbees.com/job/cpp-linux-nightly-64-STATIC-Release/69/console

06:03:26 [ RUN      ] ClientQueueTest.testTake
06:03:26 Apr 06, 2016 12:49:27 PM CppClientListener
06:03:26 INFO: SHUTDOWN for instance 39 is completed.
06:03:26 START command received: NEW INSTANCE OPEN 
06:03:26 Apr 06, 2016 12:49:27 PM com.hazelcast.instance.DefaultAddressPicker
06:03:26 INFO: [LOCAL] [dev] [3.7-SNAPSHOT] Interfaces is disabled, trying to pick one address from TCP-IP config addresses: [127.0.0.1]
06:03:26 Apr 06, 2016 12:49:27 PM com.hazelcast.instance.DefaultAddressPicker
06:03:26 INFO: [LOCAL] [dev] [3.7-SNAPSHOT] Picked Address[127.0.0.1]:5701, using socket ServerSocket[addr=/0:0:0:0:0:0:0:0,localport=5701], bind any local is true
06:03:26 Apr 06, 2016 12:49:27 PM com.hazelcast.system
06:03:26 INFO: [127.0.0.1]:5701 [dev] [3.7-SNAPSHOT] Hazelcast 3.7-SNAPSHOT (20160405 - 68d2790) starting at Address[127.0.0.1]:5701
06:03:26 Apr 06, 2016 12:49:27 PM com.hazelcast.system
06:03:26 INFO: [127.0.0.1]:5701 [dev] [3.7-SNAPSHOT] Copyright (c) 2008-2016, Hazelcast, Inc. All Rights Reserved.
06:03:26 Apr 06, 2016 12:49:27 PM com.hazelcast.system
06:03:26 INFO: [127.0.0.1]:5701 [dev] [3.7-SNAPSHOT] Configured Hazelcast Serialization version : 1
[...]

PR #86, opened 2016-04-06
Queue testTake causes segmentation fault after test passes

ihsandemir:
Fix for issue https://github.com/hazelcast/hazelcast-cpp-client/issues/85 . Waits for the thread to finish when the test is successfull.


ihsandemir:
verify

PR #89, opened 2016-04-18
Predicates implementation

ihsandemir:
Implements the builtin predicates for C++. Added the tests and the examples as well.

Changes the DataArray and EntryArray API to handle for already serialized entries.


ihsandemir:
verify

PR #90, opened 2016-04-19
Git commit date format correction for client starting log

ihsandemir:
Change commit date from 2016-04-20 to 20160420.


ihsandemir:
verify

ihsandemir:
verify

PR #92, opened 2016-04-21
Added the git commit date and commit id to the client starting log.

sancar:
Example new log:

`Apr 21, 2016 01:42:16 PM INFO: [HazelcastCppClient3.7-SNAPSHOT] [dev] [4298317824] 
(2016-04-06:0ef9654) LifecycleService::LifecycleEvent STARTING`


hazelcast/src/hazelcast/client/spi/LifecycleService.cpp
```diff
@@ -89,8 +97,12 @@ namespace hazelcast {
                 util::ILogger &logger = util::ILogger::getLogger();
                 switch (lifecycleEvent.getState()) {
                     case LifecycleEvent::STARTING :
-                        logger.info("LifecycleService::LifecycleEvent STARTING");
+                    {
+                        char msg[100];
+                        util::snprintf(msg, 100, "(%s:%s) LifecycleService::LifecycleEvent STARTING", HAZELCAST_GIT_COMMIT_DATE, HAZELCAST_GIT_COMMIT_ID);
```
ihsandemir: Is the date format as YYYYMMDD? PRD wants it this way. You can just convert it. See https://github.com/hazelcast/hazelcast-cpp-client/pull/90/files#diff-25e2207206ddcc5a7f4063ce1e5c2853R94

sancar: I did not think it as a strong requirement. And adding code just for that seemed unnecessary to me. That is why I leave it that way.

PR #95, opened 2016-04-24
CountDownLatch::awaitMillis fix for the tests.

ihsandemir:
Resets the elapsed time at the start of the await for early returns from the method without any sleep.


ihsandemir:
Verify

PR #100, opened 2016-04-27
Eliminate server restarts during tests if not needed to speedup the test runs

ihsandemir:
Speeds up the test runs by eliminating the server restarts during the each test case.


ihsandemir:
verify

PR #124, opened 2016-05-05
Add continous query(listener with predicate) to RawPointerMap.

ihsandemir:
Added continous query(listener with predicate) to RawPointerMap.


ihsandemir:
The branch incorrectly set under origin, Re-doing the PR (opened PR #126 ) using branch at ihsandemir.

PR #127, opened 2016-05-07
Basic travis CI integration

maxbeutel:
I'm working on C bindings for Hazelcast (wrapping the C++ client) and for my project I created a travis CI integration (https://travis-ci.org). As there are no .deb packages available for the Hazelcast C++ client, I do a build of a selected Hazelcast C++ client anyway during my project.

I decided to split that part out and contribute it back to you guys, are you interested? 

- Currently I run the build against two compilers, clang and g++. We can easily add more versions of the two. 
- I think it also makes sense to run the unit tests on travis, but I haven't looked into that yet, if you also think that's a good idea, I'm happy to add that to this PR. 
- In order for this to work, you need to enable travis CI integration in this Github project, as described here: https://docs.travis-ci.com/user/getting-started/

How a build looks like can be seen here: https://travis-ci.org/maxbeutel/hazelcast-cpp-client

The build is triggered with every commit to any branch. 

**NOTE** the Build Status Icon shows the status of my forked repository, so you can get an idea of how it will look like. The URL should be changed when merging into master to point to `hazelcast/hazelcast-cpp-client` repository. 

**EDIT** I haven't signed the contributors agreement yet, can do so on Monday if this PR has a chance of being merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,35 @@
+sudo: required
+dist: precise
+language: cpp
+
+matrix:
+  include:
+    - compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - george-edison55-precise-backports # cmake 3.2.3 / doxygen 1.8.3
+          packages:
+            - g++-4.9
+            - cmake
+            - cmake-data
+      env: COMPILER=g++-4.9
+    - compiler: clang
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-precise-3.7
+            - george-edison55-precise-backports # cmake 3.2.3 / doxygen 1.8.3
+          packages:
+            - clang-3.7
+            - cmake
+            - cmake-data
+      env: COMPILER=clang++-3.7
+
+before_install:
+  - sudo apt-get update -qq
+
+script:
+   - (mkdir build && cd build && cmake .. -DHZ_LIB_TYPE=STATIC -DHZ_BIT=64 -DCMAKE_BUILD_TYPE=Release && make)
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,14 @@ sudo: required
 dist: precise
 language: cpp
 
+# Handle git submodules yourself
+git:
+    submodules: false
+# Use sed to replace the SSH URL with the public URL, then initialize submodules
+before_install:
+    - sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
+    - git submodule update --init --recursive
+
 matrix:
   include:
     - compiler: gcc
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Hazelcast C++ Client
 
+[![Build Status](https://travis-ci.org/maxbeutel/hazelcast-cpp-client.svg?branch=upstream-pr-travis-ci)](https://travis-ci.org/maxbeutel/hazelcast-cpp-client)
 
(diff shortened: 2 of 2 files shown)


ihsandemir:
Hello Max, thanks for contributing to our C++ client. I am very excited with this effort. 

As it regards to the Travis CI, we already have a CI in place and it has a lot of builds for testing the C++ client. It is running on Jenkins. We can integrate your project into our build system by creating a build for your project, hence you will not need the Travis integration at all unless you have some good reason to use it. Does this sound good?

Please go ahead and fill the contributor agreement. 

We can also get your C-client become more visible to the community by putting it into our incubator program if you want. The details are here: http://blog.hazelcast.com/incubator/ , https://github.com/hazelcast-incubator

maxbeutel:
Hello @ihsandemir 

thanks a lot for your feedback. I wasn't aware that you already have a build infrastructure in place. As I understand it's also quite a bit more powerful than my travis CI integration. 

The upside of the travis CI integration is of course that the build status can be made available to the public and you can also get contributions to improve/change the build. 

But I can also understand that you don't want to maintain another build infrastructure besides your jenkins. Feel free to close this PR then. :)

Regardless of what happens with this PR, I can still sign the contributors agreement, just in case. 
===

With regards to the C client, I would probably create a separate issue for collaboration on this issue. Integrating my project into your build would be of course super helpful.

The current state (which is not yet impressive) can be seen at https://github.com/maxbeutel/hazelcast-c-client - I just wrapped a few basic `Map` operations, which worked quite nicely. 

Currently I'm still thinking about how to wrap the C++ API in a nice and user-friendly way. I summed up my points in a Stackoverflow question http://programmers.stackexchange.com/questions/317851/idiomatic-wrapping-of-c-template-type-api-in-c maybe you also got an idea?

As for featuring it in the community, that's very good idea, but I think it's too early, I would like to have a basic, working implementation and a general direction decided first, currently there is not that much to show yet, haha. 

Thanks a lot so far!

ihsandemir:
Hello @maxbeutel 

When we setup your build, you will be able to see the build status (people in github system will also see it. You can check hazelcast builds and PR details to see jenkins build status). Therefore, you should be able to get the same contribution from the community since your project will be public.

I have some ideas how to make the C API, i will write as an answer to your stackoverflow question briefly but we can also discuss and come up with a user friendly, easy to use API. API design is really not easy. We spend a lot of time to make it as intuitive and as easy as possible to use. It is a good design if you can make it so that it will be difficult for a user to misuse a method.

Regarding the community, it is OK to have an incomplete version initially, it will just make people aware of your project and you may get very useful feedbacks and may get some contribution. I looked at your project, it is in initial phase but that is no problem at all for putting it in the incubator. We can even come up with a roadmap (such as versions 0.1, 0.2, 0.3, etc.) and you can work on a roadmap which we do for other new client languages. Therefore, I would suggest just to add it to the incubator.

maxbeutel:
Hello @ihsandemir 

Thank you very much for your offers. 

I think it makes sense to close this PR and rather integrate any C bindings with the existing build infrastructure.

Milestones/Roadmap would also pretty great, because currently I can't yet oversee which functionality should be implemented for the minimal viable version. 

I read your reply on Stackoverflow, providing an API to store binary sounds like an excellent idea, I wouldn't have come up with that. I will try to make this work this week to try it out. 

How should we proceed to discuss collaboration? Shall I open an issue here in GitHub? I can also be reached via Skype at max.beutel maybe we ca have a chat whenever time allows for you.

Thanks a lot for taking the time to provide that much feedback! Much appreciated.

ihsandemir:
@maxbeutel Yes, you can open issue here to discuss but it may be better to discuss directly on the project. This is also why I told you that we can put it to inside the incubator (https://github.com/hazelcast-incubator), if we have the project at incubator, then we can just open the issues directly on your project under incubator and put the roadmap there. We can also have the implementation reviewed directly there. That would be just like any other client that we develop, and other people from team may have very useful feedbacks as well.

I will add you to my Skype. Surely, we can have a chat to get you started.

PR #135, opened 2016-05-20
Implement Ringbuffer

ihsandemir:
Added Ringbuffer container implementation and a simple test

Added exception factory implementation.


ihsandemir:
verify

PR #140, opened 2016-05-25
Alidugan patch 1 stlport compile

alidugan:
required modifications for compilation with stlport 5.2.1

### DIFF
--- a/hazelcast/src/hazelcast/client/spi/InvocationService.cpp
+++ b/hazelcast/src/hazelcast/client/spi/InvocationService.cpp
@@ -36,6 +36,7 @@
 #include "hazelcast/client/exception/IllegalStateException.h"
 
 #include <assert.h>
+#include <string>
 
 namespace hazelcast {
     namespace client {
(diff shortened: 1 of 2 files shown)


ihsandemir:
verify

hasancelik:
verify

ihsandemir:
@alidugan we can setup a nightly build for stlport based compilation, did you just put the stlport directory in the INCLUDE_DIRECTORY path? Can you also share how you built? We run the tests using a docker image, the image needs to include the stlport. If you want, you can setup your docker image with the same settings as your environment (same Linux version etc.) and share it with us.

alidugan:
I have modified CMakeList.txt to add following lines for Linux compilation;
link_directories(path_to_stlport_lib)
include_directories(path_to_stlport_includes)
link_libraries(stlport_lib)

ihsandemir:
We probably need a conditional flag in CMakeLists.txt such as STLPORT_INSTALL_DIR to provide the stlport install dir as a variable to make it compile with stlport conditionally. If this directory is provided, then use the stlport, otherwise just use the compiler provided stl. @alidugan Can you add those flags and modify the CMakeLists.txt so that we can compile the project this way?

cmake .. -DHZ_LIB_TYPE=SHARED -DHZ_BIT=64 -DCMAKE_BUILD_TYPE=Debug **-DSTLPORT_INSTALL_DIR=<install dir>**

PR #143, opened 2016-06-01
Core stack when test causes core dump

ihsandemir:
Added a log for taking stack trace of any core if existed after the test run.


hasancelik:
👍

PR #146, opened 2016-06-08
Reliable topic implementation

ihsandemir:
This PR implements the reliable topic. It only supports DISCARD_OLDEST topic overload policy. The executor is not configurable which will be enhanced later.

A fix in SerializationService is done for handling null objects.

Topic::MessageListener interface is added.

Updates the boost dependency to 1.61 version. Also added date_time header only lib as dependency for calculation of milliseconds since epoch as needed for ReliableTopic.


ihsandemir:
verify

ihsandemir:
verify

PR #164, opened 2016-06-28
Map put with ttl does not work as expected for server configured map ttl 

ihsandemir:
Fixes the map put API so that it works correctly with the server configured ttl for the map. Previously, we were sending 0 for ttl which meant infinite time, but we now send -1 to mean that it shall use the map configuration for ttl.

Originally reported at http://stackoverflow.com/questions/38064997/hazelcast-c-client-map-and-ttl/38079955#38079955


ihsandemir:
verify

ihsandemir:
verify

PR #169, opened 2016-07-12
IOHandler handleSocketException leaks memory

ihsandemir:
Release memory to prevent memory leak.


maxbeutel:
I applied the patch, that fixed the memory leak. Thanks for the quick fix!

PR #171, opened 2016-07-22
Fixes for examples and sample server configurations

oguzdemir:
Fixes were done for examples.
Required server files for some examples are added and explained in comments.


java/src/main/java/distributedMapCustomAttribute/CarAttributeServer.java
```diff
@@ -0,0 +1,120 @@
...
+
+        final Config config = new XmlConfigBuilder().build();
+        config.getSerializationConfig().addDataSerializableFactory(1, new DataSerializableFactory() {
+            @Override
+            public IdentifiedDataSerializable create(int typeId) {
+                return new Car();
+            }
+        });
+
+        /*
```
mustafaiman: why commented out section here?

examples/distributed-collections/itemlisteners/CollectionChanger.cpp
```diff
@@ -18,6 +18,7 @@
 //
 #include <hazelcast/client/HazelcastClient.h>
 
+//This program should be executed on sleep of ItemListener to record queue events on listener.
```
ihsandemir: This is not very clear, can you elaborate?

examples/distributed-collections/itemlisteners/ItemListener.cpp
```diff
@@ -60,7 +60,8 @@ int main() {
     std::cout << "Registered the listener with registration id:" << registrationId <<
     "Waiting for the listener events!" << std::endl;
 
-    hazelcast::util::sleep(5);
+    //Sleep until the execution of CollectionChanger is finished.
```
ihsandemir: We probably need to put a README in the test folder to tell the user which one to start first and do what, step by step if we are having these kinds of requests. Just assume that the user knows nothing. Furthermore, we need to make it as esy as possible for the user to try an example, therefore let's put a runExample.sh and runExample.bat scripts to run the script and let the script do the sleep, what do you think?

examples/adaptor/run-rawlist.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+$1/rawlist
```
ihsandemir: how do you start server?

ihsandemir: We also need the same scripts for windows as well probably with .bat extension.

examples/distributed-map/locking/RacyUpdate.cpp
```diff
@@ -72,6 +76,7 @@ int main() {
 
         boost::shared_ptr<Value> oldValue = map.get(key);
         hazelcast::util::sleepmillis(10);
+        (*oldValue).amount++;
```
ihsandemir: oldValue->amount++

java/attributeServer.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
ihsandemir: these scripts should be under examples folder.

java/pom.xml
```diff
@@ -27,6 +27,26 @@
...
+                <version>2.8</version>
+                <executions>
+                    <execution>
+                        <id>copy-dependencies</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>copy-dependencies</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/lib</outputDirectory>
```
ihsandemir: Did not understand why we need these lines.

sancar:
@oguzdemir Closing this one since it is not updated. f you want to merge this, please address/answer the questions asked on comments.

PR #172, opened 2016-07-23
Heartbeat fail does not close non-owner connection

ihsandemir:
It was reported that when member is powered off the request is frozen. When examined the client log we see that the client detect heartbeat failure but it does not close the connection. It closes it only after 10 seconds during socket recv. The connection close were for some reason commented out, i uncommented the connection close of non-owner connection when heartbeat fails. Need to check if any side affects of this.


sancar:
We should not close the connection. Client close the connection only if it is owner connection. And it closes connection to others only when they are not in memberlist anymore(meaning kicked out of cluster by master).

ihsandemir:
Closing the issue since this was the behaviour as expected, it is also same at the Java client side.

Issue #173, opened 2016-08-05
C++ client connection problems when node is powered off


burakgokmen:
There are several Hazelcast servers running on different nodes in our test environment.

When I power off the node where one of the Hazelcast servers (3.6.2) is running, I occasionally encounter with two problems:

1) The C++ client (3.6.3) might get stuck in printing the following message:
Future.set_exception should not be called twice : details  ExceptionMessage (Wait is timed out) at Future::get(timeInSeconds)

2) The C++ client might get stuck in hazelcast::client::IMap::keySet(). The stack output:
Thread1:
- stlp_std::ostreambuf_iterator...:copy() 
- stlp_std::ostreambuf_iterator...:put_integer() 
- stlp_std::num_put::do_put(...)
- stlp_std::priv::_put_number()
- hazelcast::client::operator<<(...)
- hazelcast::client::connection::CallFuture::get(long)
- hazelcast::client::connection::CallFuture::get()
- hazelcast::client::proxy::ProxyImpl:invoke(...)
- hazelcast::client::proxy::IMapImpl:keySetData(...)
- hazelcast::client::IMap<...>:keySet(...)

Thread3:
- hazelcast::client::Socket:::send()
- hazelcast::client::connection::OutputSocketStream::write()
- hazelcast::client::connection::Connection::init()
- hazelcast::client::connection::ConnectionManager::authenticate()
- hazelcast::client::connection::ConnectionManager::connectTo()
- hazelcast::client::connection::OwnerConnectionFuture::createNew()
- hazelcast::client::connection::ConnectionManager::createOwnerConnection()
- hazelcast::client::spi::ClusterService::connectToOne()
- hazelcast::client::connection::ClusterListenerThread::run()
- hazelcast::client::connection::ClusterListenerThread::staticRun()
- hazelcast::util::Thread::controlledThread()

C++ client connection parameters:
- ClientProperties::PROP_HEARTBEAT_INTERVAL=1
- ClientProperties::PROP_HEARTBEAT_TIMEOUT=2
- ClientProperties::PROP_REQUEST_RETRY_COUNT=0
- ClientProperties::PROP_REQUEST_RETRY_WAIT_TIME=1
- ConnectionTimeout=500
- ConnectionAttemptLimit=0
- AttemptPeriod=0

The OS is RHEL 6.6, the machine is virtual.

ihsandemir:
[...]

PR #174, opened 2016-08-12
Request retry when future exception occurs may cause warning for future being set twice when heartbeat to server is lost

ihsandemir:
Fixes the future set twice problem during server poweroff. Fixes issue https://github.com/hazelcast/hazelcast-cpp-client/issues/173


ihsandemir:
verify

PR #180, opened 2016-08-24
Connection initiation should obey the connection timeout that is set in the config

ihsandemir:
Added timeout during connection initiation.


ihsandemir:
No need for this PR any more.

Issue #190, opened 2016-11-28
LoadTest may fail with core dump signal 6 (abort)


ihsandemir:
A failing run: https://hazelcast-l337.ci.cloudbees.com/job/cpp-linux-maintenance/841/consoleFull

Symptom: When the threads are analysed I saw this stack trace:

12:34:33 #4  0xf6e61c36 in __assert_fail () from /lib/libc.so.6
12:34:33 No symbol table info available.
12:34:33 #5  0x08378834 in hazelcast::util::LittleEndianBufferWrapper::setBytes (this=0xe1a00a64, value=0xe9500678 "", len=828782182) at /home/jenkins/jenkins/workspace/cpp-linux-maintenance/hazelcast/include/hazelcast/util/LittleEndianBufferWrapper.h:210
12:34:33         __PRETTY_FUNCTION__ = "void hazelcast::util::LittleEndianBufferWrapper::setBytes(const hazelcast::byte*, int32_t)"
12:34:33 #6  0x083787b1 in hazelcast::util::LittleEndianBufferWrapper::set (this=0xe1a00a64, value=) at /home/jenkins/jenkins/workspace/cpp-linux-maintenance/hazelcast/include/hazelcast/util/LittleEndianBufferWrapper.h:206
12:34:33         len = 828782182
12:34:33 #7  0xf73120d7 in hazelcast::client::protocol::ClientMessage::setNullable<std::string> (this=0xe1a00a60, value=0xe9500638) at /home/jenkins/jenkins/workspace/cpp-linux-maintenance/hazelcast/include/hazelcast/client/protocol/ClientMessage.h:180
12:34:33         isNull = false
12:34:33 #8  0xf730f6d8 in hazelcast::client::protocol::ClientMessage::set (this=0xe1a00a60, value=0xe9500638) at /home/jenkins/jenkins/workspace/cpp-linux-maintenance/hazelcast/src/hazelcast/client/protocol/ClientMessage.cpp:128
12:34:33 No locals.
[...]

PR #191, opened 2016-11-28
ConnectionManager principal member object is not thread safe

ihsandemir:
Makes the principal object thread safe in the ConnectionManager. This object is accessed via owner and non-owner connection authentications which can be done through different threads.

Also enhanced the logging statements. Made the members list logging the same as the java client.

Hoping that this fixes the failure reported at issue: https://github.com/hazelcast/hazelcast-cpp-client/issues/190


sancar:
👍

PR #193, opened 2016-12-02
Initial Near Cache Implementation Interfaces

ihsandemir:
Introduces the proxymanager (for map only) as used at java client.
Implemented the empty classes that will form the interfaces for the near cache implementation. (A lot of TODOs exist at this version which shall be completed in later PRs). Existing map functionalities are preserved.


hazelcast/include/hazelcast/client/HazelcastClient.h
```diff
@@ -705,17 +702,27 @@ namespace hazelcast {
...
+                return proxyManager.getOrCreateProxy<FACTORY>(serviceName, name, factory);
+            }
+
             ClientConfig clientConfig;
             ClientProperties clientProperties;
             spi::ClientContext clientContext;
             spi::LifecycleService lifecycleService;
             serialization::pimpl::SerializationService serializationService;
             connection::ConnectionManager connectionManager;
+            std::auto_ptr<internal::nearcache::NearCacheManager> nearCacheManager;
```
ihsandemir: At java side, it uses a method to create it: createNearCacheManager which returns a pointer which uses a client extension which we do not have a cpp client. But i preferred using the same code as java does. Provided that its constructor parameters are already present, it can be done both ways. if createNearCacheManager does more complex things in the future, it may be better to keep it as a pointer.

PR #194, opened 2016-12-20
Near cache implementation with dummy data store and no eviction

ihsandemir:
Implementation of near cache with dummy no-op data stores and dummy no-op eviction. Added a near cached version of the ClientMapTest.


CMakeLists.txt
```diff
@@ -136,8 +136,8 @@ IF(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
 
 	message(STATUS "${HZ_BIT} Bit")
 
-	set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -Wall -Werror ${HZ_BIT_FLAG} ${HZ_CODE_COVERAGE_COMPILE_FLAGS} ${HZ_VALGRIND_COMPILE_FLAGS} ${HZ_C11_FLAGS}")
-	set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -Wall ${HZ_BIT_FLAG} ${HZ_CODE_COVERAGE_COMPILE_FLAGS} ${HZ_VALGRIND_COMPILE_FLAGS} ${HZ_C11_FLAGS}")
+	set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g -Wall -Werror -D__STDC_LIMIT_MACROS -D__STDC_CONSTANT_MACROS ${HZ_BIT_FLAG} ${HZ_CODE_COVERAGE_COMPILE_FLAGS} ${HZ_VALGRIND_COMPILE_FLAGS} ${HZ_C11_FLAGS}")
```
sancar: I removed because they were causing compile problems . And you are adding them back .
Do we really need these ? Because I remember removing these usages. 
https://github.com/ihsandemir/hazelcast-cpp-client/commit/6c5027f67b4aac723e7cad625ce88f2a84a6ff4c

hazelcast/include/hazelcast/client/ClientConfig.h
```diff
@@ -395,10 +405,10 @@ namespace hazelcast {
 
             std::map<std::string, config::ReliableTopicConfig> reliableTopicConfigMap;
 
-            std::map<std::string, config::NearCacheConfig> nearCacheConfigMap;
+            util::SynchronizedMap<std::string, config::NearCacheConfig> nearCacheConfigMap;
```
sancar: Why do we need synchronisation here ?. Isn't this populated at the start only ? After that, it should read only.

ihsandemir: It is used as ConcurrentHashMap at Java side. If not needed then we can remove after removing from Java implementation.

CMakeLists.txt
ihsandemir: Needed for using INT32_MAX and IMT64_MAX. Will keep the flags now and if we come with a compilation problem, then we will look at it again.

PR #208, opened 2017-01-13
Header includes missing from HazelcastAll.h (e.g. RawPointerMap)

### LINKED ISSUE #195: variable m has initializer but incomplete type Error
Trying the below code ins Client 3.6.3 stable version, during compilation below code:
`hazelcast::client::IMap<std::string, std::string> m = hz.getMap<std::string, std::string>("map");
hazelcast::client::adaptor::RawPointerMap<std::string, std::string> map(m);`

variable m has initializer but incomplete type error is thrown any reason why this happens?

ihsandemir:
Added the missing headers to the list of headers.

Fixes #195

### DIFF
--- a/hazelcast/include/hazelcast/client/HazelcastAll.h
+++ b/hazelcast/include/hazelcast/client/HazelcastAll.h
@@ -17,12 +17,16 @@
 // Created by sancar koyunlu on 29/01/14.
 //
 
-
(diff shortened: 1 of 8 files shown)


ihsandemir:
run-lab-run

ihsandemir:
verify

PR #214, opened 2017-01-18
Client map putTTL test fails against server 3.7.4

ihsandemir:
Corrected the usages of ASSERT_NULL_EVENTUALLY.

Also terminates the clients gracefully at the BasicClientNearCacheTest.


ihsandemir:
verify

PR #217, opened 2017-01-24
Near cache performance test

ihsandemir:
Added the near-cache performance test and the HdrHistogram library dependency (only for MacOs and Linux) to be used when the cmake flag -DHZ_NEARCAHCE_PERFORMANCE=ON exists. Coordinated omission is accounted for.


ihsandemir:
verify

ihsandemir:
No need for this test PR.

PR #246, opened 2017-03-21
testListenersWhenClusterDown test causes invalid owner member uuid error

ihsandemir:
Java server rejects the non-owner authentication request with: Member having uuid 294fcb20-33ed-42ec-93bf-c45cd3440889 is not part of the cluster. Client Authentication rejected.

Added ensureOwnerConnectionAvailable functionality to connection manager.


ihsandemir:
verify

PR #253, opened 2017-03-28
remove boost shared ptr dependency from api and internal 

sancar:


sancar:
Development parked to revisited in 4.0

PR #257, opened 2017-03-30
Add isDone method to Future API

ihsandemir:
Added the isDone method to Future API.


hazelcast/include/hazelcast/client/connection/CallFuture.h
```diff
@@ -56,6 +56,8 @@ namespace hazelcast {
                 int64_t getCallId() const;
 
                 const Connection &getConnection() const;
+
+                const bool isDone() const;
```
sancar: Return type should be `bool` instead of `const bool`

PR #260, opened 2017-03-30
Implement std::future like api

ihsandemir:
Implemented the std::future like class which is used by IMap::submitToKey.


ihsandemir:
Reviewed by @sancar

PR #261, opened 2017-03-31
Edit example for Entry Listener at README file

Zenikolas:


README.md
```diff
@@ -498,33 +498,46 @@ int main() {
...
 using namespace hazelcast::client;
 
-class SampleEntryListener {
+class SampleEntryListener : public hazelcast::client::EntryListener<std::string, std::string> {
   public:
-
-  void entryAdded( EntryEvent<std::string, std::string> &event ) {
-    std::cout << "entry added " <<  event.getKey() << " "
-        << event.getValue() << std::endl;
+  void entryAdded( const EntryEvent<std::string, std::string> &event ) override {
```
ihsandemir: Can you remove override keywords since it is only supported since c++11 (http://en.cppreference.com/w/cpp/language/override )? We want to also support older standards as well.

Zenikolas: Ok!

PR #265, opened 2017-04-02
Examples cmake minimum required version can be changed to 2.6.4

ihsandemir:
Change examples minimum required cmake version to 2.6.4 as this is the case for the master CMakeLists.txt.


ihsandemir:
verify

PR #268, opened 2017-04-04
Type stack_st_SSL_CIPHER does not exist at earlier openssl versions (e.g. 0.9.8e)

ihsandemir:
Use a more portable ssl type name when retrieving the ciphers. This is needed when compiling the library on CentOS 5 which comes with openssl 0.9.8.

The STACK_OF(SSL_CIPHER) macro is more portable.


ihsandemir:
verify

PR #277, opened 2017-04-22
Nightly release builds are failing

ihsandemir:
HttpsClientTest should be disabled if tls/ssl is not enabled.


ihsandemir:
verify

PR #278, opened 2017-04-24
Integrate AWS discovery and private to public ip translation to the client

ihsandemir:
Added aws address translator to be used for private to public ip conversion.

Integrated the cloud discovery using AWSClient to client.

Made the tagSet and groupSet reads from AWS response to be optional since for some instances they may not exist.

Minor fix to CppClientListener to obtain the license key from environment rather than as a command line parameter.

### DIFF
--- a/hazelcast/include/hazelcast/client/ClientProperties.h
+++ b/hazelcast/include/hazelcast/client/ClientProperties.h
@@ -80,6 +80,7 @@ namespace hazelcast {
 
             const ClientProperty& getRetryWaitTime() const;
 
+            const ClientProperty& getAwsMemberPort() const;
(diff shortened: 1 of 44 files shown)


ihsandemir:
verify

sancar:
I think we should confine `HZ_BUILD_WITH_SSL` micro to one or two classes. 
Seeing in connection manager rings some bells. It is to much used and all over the place. I see similar usage with for SSL feature also. 

DescribteInstance or SyncHttpClient will use `HZ_BUILD_WITH_SSL` . And they will not be guarded in class level but in method level,  so that it will be invisible from outside.

PR #285, opened 2017-05-12
Client does not shut down properly when cluster connection attempts all fail

ihsandemir:
Fix for shutting down the client properly if the cluster connection attempts all fail.

### DIFF
--- a/hazelcast/include/hazelcast/client/connection/Connection.h
+++ b/hazelcast/include/hazelcast/client/connection/Connection.h
@@ -68,7 +68,7 @@ namespace hazelcast {
 
                 void connect(int timeoutInMillis);
 
-                void close();
+                void close(const char *closeReason = NULL);
 
                 void write(protocol::ClientMessage *message);
 
--- a/hazelcast/include/hazelcast/client/exception/ProtocolExceptions.h
+++ b/hazelcast/include/hazelcast/client/exception/ProtocolExceptions.h
@@ -33,8 +33,9 @@ namespace hazelcast {
         namespace exception {
             class HAZELCAST_API ProtocolException : public IException {
             public:
-                ProtocolException(const std::string& source, const std::string& message, int32_t errorNo, int32_t causeCode)
-                        : IException(source, message), errorCode(errorNo), causeErrorCode(causeCode) {
+                ProtocolException(const std::string& message, const std::string& details, int32_t errorNo,
+                                  int32_t causeCode)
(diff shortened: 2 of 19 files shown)


ihsandemir:
verify

ihsandemir:
verify

ihsandemir:
verify

ihsandemir:
verify

sancar:
I did not get why shutdown needs a countdown latch to wait. 
Can you describe failing scenario ? 
First thing comes to my mind that a thread called shutdown and running shutdown method. In the meantime, destructor called shutdown. Checked that it is not active and returned immediately causing to delete everything while another thread accessing the fields of it. 
But in this scenario, it means that HazelcastInstance shared among threads without a sharedPtr. 

I am trying to understand if this is a problem we need to solve, or is it just a illegal usage.

ihsandemir:
@sancar Yes, that is exactly the case. I saw it test https://github.com/hazelcast/hazelcast-cpp-client/blob/master/hazelcast/test/src/cluster/ClientConnectionTest.cpp#L48

So, the test finishes and hence the client destructor is executed before the shutdown is actually completed and hence during the shutdown illegal memory access occurs. I don't think it is an illegal use. 

Yes, java uses shared references which eliminates the problem. It may also be solved by keeping an internally shared pointer and just returning a wrapper object HazelcastClient to the user, and also keeping a reference inside the clientcontext object which is being used between the threads. 

In any case, we need to guarantee that the shutdown is finished and all threads are closed before exiting the destructor.

sancar:
I still don't get it. How is the HazelcastClient shared between two threads in this test? If it is a raw pointer and one is calling destructor. This usage is illegal. 
If it was shared via a shared_ptr<HazelcastCLient> then since the one calling shutdown is not decremented the shared_ptr count, no one can call destructor before it finishes.

ihsandemir:
@sancar Scenario: ClusterListenerThread at this line: https://github.com/ihsandemir/hazelcast-cpp-client/blob/shutdownFix/hazelcast/src/hazelcast/client/connection/ClusterListenerThread.cpp#L88
and HazelcastClient passes this line: https://github.com/hazelcast/hazelcast-cpp-client/blob/master/hazelcast/src/hazelcast/client/HazelcastClient.cpp#L56

But I think I can also fix it by changing the lines 87 and 88 https://github.com/ihsandemir/hazelcast-cpp-client/blob/shutdownFix/hazelcast/src/hazelcast/client/connection/ClusterListenerThread.cpp#L87 .

ihsandemir:
verify

PR #290, opened 2017-06-01
Visual Studio compiler warnings - getenv and exception specification ignored

ihsandemir:
Fixes the following warnings:

- Preconditions.h(83): warning C4290: C++ exception specification ignored except to indicate a function is not __declspec(nothrow)

- DescribeInstancesTest.cpp(58): warning C4996: 'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead.

Visual studio C++ compiler does not want an exception definition in method signature. Also, suppressed the insecure warning (4996) for the getenv usage.


ihsandemir:
verify

Issue #300, opened 2017-06-21
cpp-linux-nightly-32-STATIC-Debug aborted after testAddInterceptor


ihsandemir:
We see a linux build is aborted following the testAddInterceptor test:

01:45:20 [0;32m[       OK ] [mClientMapTest/0.testAddInterceptor (13 ms)
01:45:20 Jun 20, 2017 10:45:20 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [-150066816] LifecycleService::LifecycleEvent SHUTTING_DOWN
01:45:20 Jun 20, 2017 10:45:20 PM WARNING: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [-150066816] Closing connection (id:2) to Address[127.0.0.1:5702] with socket id 9. Hazelcast client is shutting down
01:45:20 Jun 20, 2017 10:45:20 PM WARNING: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [-150066816] Closing connection (id:3) to Address[127.0.0.1:5701] with socket id 10. Hazelcast client is shutting down
01:45:20 Jun 20, 2017 10:45:20 PM WARNING: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [-150066816] Closing connection (id:1) to Address[127.0.0.1:5701] with socket id 4 as the owner connection.Closing owner connection to Address[127.0.0.1:5701]
01:45:20 Jun 20, 2017 10:45:20 PM WARNING: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [-525345936] hz.heartbeater is cancelled 
01:45:20 Jun 20, 2017 10:45:20 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [-516953232] LifecycleService::LifecycleEvent CLIENT_DISCONNECTED
01:45:20 Jun 20, 2017 10:45:20 PM com.hazelcast.nio.tcp.TcpIpConnection
01:45:20 INFO: [127.0.0.1]:5701 [dev] [3.8.2-SNAPSHOT] Connection[id=2, /127.0.0.1:5701->/127.0.0.1:44614, endpoint=[127.0.0.1]:44614, alive=false, type=CPP_CLIENT] closed. Reason: Connection closed by the other side
01:45:20 Jun 20, 2017 10:45:20 PM com.hazelcast.client.ClientEndpointManager
[...]

PR #301, opened 2017-06-21
Remove interceptor at the end of test testAddInterceptor

ihsandemir:
Removed interceptor at the end of test testAddInterceptor.

Decreased the lock scope when cancelling the thread at linux.

Related to issue #300


ihsandemir:
verify

ihsandemir:
verify

Issue #299, opened 2017-06-21
Windows build unexpectedly end at the end of test interceptSSLBasic


ihsandemir:
During some nightly builds, we observe that the build terminated unexpectedly. The log is:

00:38:16 [ RUN      ] SocketInterceptorTest.interceptSSLBasic
00:38:18 Jun 20, 2017 10:38:14 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4308] (20170620:8e3e18f) LifecycleService::LifecycleEvent STARTING
00:38:18 Jun 20, 2017 10:38:14 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4292] Connected and authenticated by Address[127.0.0.1:5701]. Connection id:1 , socket id:796 as owner connection. Client uuid: 18611324-1ad7-4e1c-a1f2-4fffd7b0892e. Owner member uuid: 25e236fe-c44a-436e-84d0-fca78caf8177
00:38:18 Jun 20, 2017 10:38:14 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4292] LifecycleService::LifecycleEvent CLIENT_CONNECTED
00:38:18 Jun 20, 2017 10:38:14 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4292] 
00:38:18 Members [1]  {
00:38:18 	Member[127.0.0.1]:5701 - 25e236fe-c44a-436e-84d0-fca78caf8177
00:38:18 }
00:38:18 
00:38:18 Jun 20, 2017 10:38:14 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4308] Connected and authenticated by Address[127.0.0.1:5701]. Connection id:2 , socket id:644. Client uuid: 18611324-1ad7-4e1c-a1f2-4fffd7b0892e. Owner member uuid: 25e236fe-c44a-436e-84d0-fca78caf8177
00:38:18 Jun 20, 2017 10:38:14 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4308] LifecycleService::LifecycleEvent STARTED
00:38:18 Jun 20, 2017 10:38:14 PM INFO: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4308] LifecycleService::LifecycleEvent SHUTTING_DOWN
[...]

PR #302, opened 2017-06-21
ClusterService shutdown may not work correctly if it was called when client was disconnected

ihsandemir:
Remove the unneeded active flag for ClusterService, since it may cause the shutdown code being not executed.

Fix done while analysing #299

### DIFF
--- a/hazelcast/include/hazelcast/client/connection/ClusterListenerThread.h
+++ b/hazelcast/include/hazelcast/client/connection/ClusterListenerThread.h
@@ -56,8 +56,6 @@ namespace hazelcast {
 
                 void setThread(util::Thread *);
 
-                const util::Thread *getThread() const;
-
                 static void staticRun(util::ThreadArgs &args);
 
                 void run(util::Thread *currentThread);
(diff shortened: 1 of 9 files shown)


hazelcast/src/hazelcast/client/connection/ClusterListenerThread.cpp
```diff
@@ -85,8 +81,16 @@ namespace hazelcast {
...
-                                isStartedSuccessfully = false;
-                                clientContext.getLifecycleService().shutdown();
+                                /**
+                                 * We can not call shutdown if the client is already shutting down.
+                                 * If we do this, then it will cause deadlock since shutdown waits for
+                                 * until another in-progress shutdown is finished. Hence, if the shutdown
+                                 * was called and it was waiting for cluster thread termination but the following
+                                 * line was reached at cluster thread, then this will be a deadlock.
+                                 */
+                                if (clientContext.getLifecycleService().isRunning()) {
```
sancar: So there is a deadlock but we are not solving with it with this check. There is a race. isRunning and shutdown calls are not atomic. It is possible that `shutdown` is called after `isRunning` check.

hazelcast/include/hazelcast/client/connection/ClusterListenerThread.h
```diff
@@ -56,8 +56,6 @@ namespace hazelcast {
 
                 void setThread(util::Thread *);
```
sancar: Can we also move the initialization of this thread from ClusterService to ClusterListenerThread and remove setThread  also. And also make staticRun and run methods private.

hazelcast/src/hazelcast/client/connection/ClusterListenerThread.cpp
ihsandemir: OK. Reverted the change which made the client shutdown blocking. Instead, now we only block (waits for latch) on LifecycleService destructor.

hazelcast/include/hazelcast/client/connection/ClusterListenerThread.h
ihsandemir: OK, done.

PR #304, opened 2017-06-21
Closing socket writes warning "Receive failed. An operation was attempted on something that is not a socket"

ihsandemir:
I saw the following log for a closing connection at some builds:
Jun 21, 2017 10:49:29 PM WARNING: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [4044] Closing connection (id:3) to Address[127.0.0.1:5702] with socket id 584.
Jun 21, 2017 10:49:29 PM WARNING: [HazelcastCppClient3.8.2-SNAPSHOT] [dev] [3780] [IOHandler::handleSocketException] Closing socket to endpoint 127.0.0.1:5702, Cause:ExceptionMessage {Receive failed. An operation was attempted on something that is not a socket.
} at TcpSocket::receive

The second log line can actually be avoided and we should not see it for a closing socket.
Process socket data only if the connection is not closed.

### DIFF
--- a/hazelcast/src/hazelcast/client/connection/Connection.cpp
+++ b/hazelcast/src/hazelcast/client/connection/Connection.cpp
@@ -106,14 +106,18 @@ namespace hazelcast {
                 if (!_isOwnerConnection) {
                     readHandler.deRegisterSocket();
                 }
-                socket->close();
-                if (_isOwnerConnection) {
-                    return;
+
+                if (!_isOwnerConnection) {
+                    /**
+                     * Remove connection and socket from the list before closing the socket
+                     * in order to prevent the use of closed socket descriptor.
+                     */
+                    clientContext.getConnectionManager().onConnectionClose(serverAddr, socketId);
+
+                    clientContext.getInvocationService().cleanResources(*this);
                 }
 
-                clientContext.getConnectionManager().onConnectionClose(serverAddr, socketId);
-
-                clientContext.getInvocationService().cleanResources(*this);
+                socket->close();
             }
 
 


hazelcast/src/hazelcast/client/connection/InSelector.cpp
```diff
@@ -66,7 +66,7 @@ namespace hazelcast {
                             sleepingSocket->receive(&wakeUpSignal, sizeof(int));
                         } else {
                             boost::shared_ptr<Connection> conn = connectionManager.getConnectionIfAvailable(fd);
-                            if (conn.get() != NULL) {
+                            if (conn.get() != NULL && conn->live) {
```
sancar: Connection live is closed and then it is removed from this list. 
As long as we don't change this order, it is possible that connection is closed after your `live` check here. 
It think fix should be fixing the order of 
`socket->close();`
with followings . 
```
socketConnections.remove(socketId);
connections.remove(address);
```

ihsandemir: Current situation: https://github.com/hazelcast/hazelcast-cpp-client/blob/master/hazelcast/src/hazelcast/client/connection/Connection.cpp#L114  removal is after socket.close.  Yes, reversing the order may also help decrease this possibility, but what about already scheduled sockets waiting on select call. May be I can do this: In addition to this ordering, also trigger IOSelector wakeup after the removal from the list.

sancar: I don't think another wakeup call is necessary. 
`select` call already handles the socket close. 
`EBADF` is checked explicitly for that in our code.

ihsandemir: Yes, but I want to wakeup after removing from the list just before I close the connection. By the way, this is only for explicit close of the connection usually during shutdown.

sancar: Why invent a second way when it is already handled ? What will we gain ?

ihsandemir: OK, I understand now. i think it is not EBADF but errorfds will be filled for the close case (if the fd was initally valid (i.e. open socket) and closed during select wait). No need for wakeup.

ihsandemir:
verify

ihsandemir:
verify

ihsandemir:
verify

PR #305, opened 2017-06-23
Missing support for PartitionAware as in Java client

ihsandemir:
Implemented PartitionAware support.

### DIFF
--- a/hazelcast/include/hazelcast/client/ClientConfig.h
+++ b/hazelcast/include/hazelcast/client/ClientConfig.h
@@ -296,7 +296,7 @@ namespace hazelcast {
             *
             *  @return serializationConfig
             */
-            SerializationConfig const &getSerializationConfig() const;
+            SerializationConfig &getSerializationConfig();
 
             /**
             * SerializationConfig is used to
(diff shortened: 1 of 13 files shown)


ihsandemir:
verify

hazelcast/include/hazelcast/client/ClientConfig.h
```diff
@@ -296,7 +296,7 @@ namespace hazelcast {
             *
             *  @return serializationConfig
             */
-            SerializationConfig const &getSerializationConfig() const;
+            SerializationConfig &getSerializationConfig();
```
sancar: What was the reason behind removing this two `const`'s ?

ihsandemir: if it is const, then we can not do getSerializationConfig().registerSerializer(). The only way then to registerSerializer is the ClientConfig.setSerializationConfig(). This was the motivation behind it. I was to use it in a test but I later did not use it but I thought we better fix this as well.

ihsandemir:
verify

PR #312, opened 2017-07-11
Add an example for PartitionAware usage

ihsandemir:
Added example for PartitionAware usage. Also updated the HazelcastAll.h.


ihsandemir:
verify

PR #320, opened 2017-07-28
The previously connected server address needs to be put at the end of the server adress list

ihsandemir:
The previously connected server address needs to be put at the end of the server list when reconnecting. Current code puts the address twice and it may put the previous address earlier in the list.


ihsandemir:
verify

PR #327, opened 2017-08-07
Mac OS X release script needs to generate the doxygen files

ihsandemir:
Change mac OS X release script to generate the doxygen files.


ihsandemir:
verify

PR #336, opened 2017-10-23
Mixed type support implementation - MixedMapProxy

ihsandemir:
Implementation for supporting the mixed types in Hazelcast data structures.


hazelcast/src/hazelcast/client/serialization/pimpl/SerializationService.cpp
```diff
@@ -424,278 +420,72 @@ namespace hazelcast {
...
+                    // let usage of static member.
+                    DataInput dataInput(data->toByteArray(), 4);
 
-                    DataInput dataInput(data.toByteArray(), Data::DATA_OFFSET);
+                    ObjectDataInput objectDataInput(dataInput, getSerializerHolder());
 
-                    int32_t typeId = data.getType();
+                    if (SerializationConstants::CONSTANT_TYPE_DATA == type.typeId ||
+                            SerializationConstants::CONSTANT_TYPE_PORTABLE == type.typeId) {
+                        assert(type.typeId == objectDataInput.readInt());
```
sancar: Be careful here. You need to get this readInt out of assert. In `run` mode, readInt will not be called.

ihsandemir:
verify

ihsandemir:
verify

ihsandemir:
verify

hazelcast/include/hazelcast/client/serialization/pimpl/SerializationService.h
```diff
@@ -55,30 +57,62 @@ namespace hazelcast {
...
                         DataOutput output;
 
-                        ObjectDataOutput dataOutput(output, portableContext);
+                        SerializerHolder &serializerHolder = getSerializerHolder();
+                        ObjectDataOutput dataOutput(output, &serializerHolder);
 
                         writeHash<T>(object, output);
 
-                        dataOutput.writeObject<T>(object);
+                        if (NULL == object) {
```
sancar: `writeObject` seems like doing the same thing. Can we double check if it is the case so that we could remove code duplication ?

hazelcast/include/hazelcast/client/serialization/pimpl/SerializationService.h
```diff
@@ -55,30 +57,62 @@ namespace hazelcast {
 
         namespace serialization {
             namespace pimpl {
+                struct HAZELCAST_API ObjectType {
+                    ObjectType() : typeId(0), factoryId(-1), classId(-1) {}
```
sancar: It looks like we should not need an empty constructor. 
We may event do this private ?

hazelcast/include/hazelcast/client/serialization/pimpl/SerializationService.h
```diff
@@ -106,12 +140,15 @@ namespace hazelcast {
                             return std::auto_ptr<T>();
                         }
 
-                        // Constant 4 is Data::TYPE_OFFSET. Windows DLL export does not
+                        int32_t typeId = data.getType();
+                        
+                        // Constant 8 is Data::DATA_OFFSET. Windows DLL export does not
                         // let usage of static member.
-                        DataInput dataInput(data.toByteArray(), 4);
+                        DataInput dataInput(data.toByteArray(), 8);
```
sancar: As far as I understand SerializationConstants.checkClassType is not used anymore. 
Can we add that that check here ?

hazelcast/include/hazelcast/client/HazelcastClient.h
```diff
@@ -503,6 +504,13 @@ namespace hazelcast {
                 return IMap<K, V>(proxy);
             }
 
+            map::MixedMapProxy getMixedMap(const std::string &name) {
```
sancar: I think API name should be `MixedMap` rather than `MixedMapProxy`

sancar: May be we can use `adaptor`  approach as we do RawPointerMap ? 
Is there a specific reason why we choose a different approach ?

hazelcast/include/hazelcast/client/map/MixedMapProxy.h
```diff
@@ -0,0 +1,1136 @@
...
+                void evictAll();
+
+                /**
+                * Returns the entries for the given keys.
+                *
+                * @param keys keys to get
+                * @return map of entries
+                */
+                template <typename K>
+                std::vector<std::pair<TypedData, TypedData> > getAll(const std::set<K> &keys) {
```
sancar: Isn't this gonna cause problems ? 
We have a class in vector that does not copy properly -> When copied old data is useless.

hazelcast/include/hazelcast/client/serialization/pimpl/SerializationService.h
ihsandemir: It is actually being used from SerializationService::getObjectType.

hazelcast/include/hazelcast/client/serialization/pimpl/SerializationService.h
ihsandemir: Ok, reused the ObjectDataOutput::writeObject.

hazelcast/include/hazelcast/client/map/MixedMapProxy.h
ihsandemir: yes, that may be a problem in the future, since it is a user api. I changed the copy constructor of TypedData to perform deep copy of bytes.

hazelcast/include/hazelcast/client/HazelcastClient.h
ihsandemir: Adaptor could not apply in this concept, because what scan we adopt? IMap requires type information. Yes, I will put MixedMap instead of MixedMapProxy for the user API.

hazelcast/include/hazelcast/client/MixedMap.h
```diff
@@ -0,0 +1,1134 @@
...
+
+            /**
+            * Returns the <tt>EntryView</tt> for the specified key.
+            *
+            *
+            * @param key key of the entry
+            * @return <tt>EntryView</tt> of the specified key
+            * @see EntryView
+            */
+            EntryView <TypedData, TypedData> getEntryView(const TypedData &key);
```
sancar: Why key is TypedData instead of directly key itself ?

hazelcast/include/hazelcast/client/MixedMap.h
```diff
@@ -0,0 +1,1134 @@
...
+            * with same classId and factoryId.
+            *
+            * @tparam EntryProcessor type of entry processor class
+            * @tparam ResultType that entry processor will return
+            * @param entryProcessor that will be applied
+            * @param key of entry that entryProcessor will be applied on
+            * @return result of entry process.
+            */
+            template<typename K, typename ResultType, typename EntryProcessor>
+            boost::shared_ptr<ResultType> executeOnKey(const K &key, EntryProcessor &entryProcessor) {
```
sancar: Can we return the result of EntryProcessor as TypeData ?

hazelcast/include/hazelcast/client/MixedMap.h
```diff
@@ -0,0 +1,1134 @@
...
+
+            template <typename K>
+            EntryVector getAllInternal(
+                    const std::map<int, std::vector<std::pair<const K *, boost::shared_ptr<serialization::pimpl::Data> > > > &partitionToKeyData) {
+
+                /**
+                 * This map is needed so that we do not deserialize the response data keys but just use
+                 * the keys at the original request
+                 */
+                std::map<boost::shared_ptr<serialization::pimpl::Data>, const K *> dataKeyPairMap;
```
sancar: This dataKeyPairMap is not used .

hazelcast/src/hazelcast/client/serialization/TypeIDS.cpp
```diff
@@ -66,6 +66,46 @@ namespace hazelcast {
...
+            int32_t getHazelcastTypeId(const std::vector<float> *object) {
+                return pimpl::SerializationConstants::CONSTANT_TYPE_FLOAT_ARRAY;
+            }
+
+            int32_t getHazelcastTypeId(const std::vector<double> *object) {
+                return pimpl::SerializationConstants::CONSTANT_TYPE_DOUBLE_ARRAY;
+            }
+
+            int32_t getHazelcastTypeId(const std::vector<std::string *> *object) {
+                return pimpl::SerializationConstants::CONSTANT_TYPE_STRING_ARRAY;
```
sancar: std::vector<std::string *> and  std::vector<std::string> serializers should have different id's. Otherwise you will get illegal cast here
https://github.com/hazelcast/hazelcast-cpp-client/pull/336/files#diff-8c0e1e3a5025fea24f26381129c53b1eR248

hazelcast/include/hazelcast/client/MixedMap.h
ihsandemir: Corrected.

hazelcast/include/hazelcast/client/MixedMap.h
ihsandemir: Corrected.

hazelcast/include/hazelcast/client/MixedMap.h
ihsandemir: Deleted unused variable.

ihsandemir:
verify

ihsandemir:
verify

ihsandemir:
verify

hasancelik:
verify

ihsandemir:
verify

PR #354, opened 2017-11-28
Near cache implementation for MixedMap

ihsandemir:
Implements the near cache support for mixed type support for all data structures.


ihsandemir:
verify

hazelcast/src/hazelcast/client/MixedMap.cpp
```diff
@@ -1,297 +0,0 @@
...
-            return protocol::codec::MapExecuteOnKeyCodec::ResponseParameters::decode(response).response;
-        }
-
-        void
-        MixedMap::putAllInternal(const std::map<int, EntryVector> &entries) {
-            proxy::IMapImpl::putAllData(entries);
-        }
-
-        boost::shared_ptr<serialization::pimpl::Data>
-        MixedMap::toShared(const serialization::pimpl::Data &data) {
```
sancar: This looks redundant since you already have the same method in ProxyImpl. 
And same method also can be deleted from ClientMapProxy

PR #355, opened 2017-11-29
Change library version to 3.9-SNAPSHOT - API const fix and some cleanup

ihsandemir:
Changed library version to 3.9-SNAPSHOT. Also made some method params as const reference to allow for stack allocated parameter usage.

Also a correction for an incorrect namespace forward declaration.


ihsandemir:
verify

PR #364, opened 2017-12-12
Missing example and documentation for global serializer

ihsandemir:
Added globalSerializerExample example for global serializer usage.

Updated javadoc for global serializer.

Added the global serializer documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
 * [Serialization Support](#serialization-support)
   * [Custom Serialization](#custom-serialization) 
   * [Polymorphic Types Serialization](#polymorphic-types-serialization) 
+  * [Global Serializer](#global-serializer) 
 * [Raw Pointer API](#raw-pointer-api)
 * [Mixed Object Types Supporting HazelcastClient](#mixed-object-types-supporting-hazelcastclient)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -367,6 +368,19 @@ If you want to use custom objects which are polymorphic, you need to register bo
             boost::shared_ptr<serialization::StreamSerializer>(new Derived1CustomSerializer));
 ```
 
+## Global Serializer
+Sometimes you may want to use your own serialization for some objects and you may not want to define the Hazelcast serialization for that object or you just do not want to define the getHazelcastTypeId free function for that object, then you can use the global serializer configuration for this purpose. You set the global serializer and you manage how you serialize and deserialize the objects. Here is how you register a global serializer:
```
Serdaro: Sometimes you may want to use your own serialization for some objects and do not want to define the Hazelcast serialization for those objects. Or, you may not want to define the `getHazelcastTypeId` free function for those objects. In these cases, you can use the global serializer configuration for this purpose. You set the global serializer and manage how you serialize and deserialize the objects. Here is how you register a global serializer:

ihsandemir: OK, changed.

Issue #328, opened 2017-08-16
Inefficient ObjectDataInput::readByteArray() 


RomanB0:
Hi,

I'm using custom IdentifiedDataSerializable class with simple integer and byte array fields.
Using profiling tools (callgrind) I found that the **ObjectDataInput::readByteArray()** is very inefficient causing huge CPU load of the whole binary. 

The problem is, that readArray()  [hazelcast/include/hazelcast/client/serialization/pimpl/DataInput.h] does read data one by one in a loop, while for each byte it calls readByte() [hazelcast/src/hazelcast/client/serialization/pimpl/DataInput.cpp] which does moreover an extra buffer size check checkAvailable(), although this check was already done at the beginning of readArray().

Problem is present in 3.8.3 version and older.

**Workaround**
It is possible to replace readByteArray() with a sequence of readInt() to get the byte array length and then readFully() to populate the prearranged buffer with the given size.

Thank you,
Roman

ihsandemir:
Yes, I think we can change this for all primitive array reads. Thanks for reporting.

RomanB0:
Thank you very much, that would be great! 

I found also one more place, which seems to consume too much CPU and might be worth reviewing.
The ResponseParameters() which was called roughly x100.000 times, calls the ClientMessage::get(<Data>) several billion times x1.000.000.000 and takes 10% CPU in my test case.

hazelcast::client::protocol::codec::MapGetCodec::ResponseParameters::ResponseParameters(hazelcast::client::protocol::ClientMessage&)

hazelcast::client::serialization::pimpl::Data hazelcast::client::protocol::ClientMessage::get<hazelcast::client::serialization::pimpl::Data>() 

Thank you.

PR #365, opened 2017-12-13
Inefficient ObjectDataInput array reads and ClientMessage Data get

ihsandemir:
Makes the array reads more efficient.

Makes array writes work similar way as the array reads.

Makes the client message data read more efficient by minimising the size checks and reading the full byte array at a single api call.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/328


ihsandemir:
verify

ihsandemir:
verify

Issue #369, opened 2017-12-19
Windows builds fail on client shutdown


ihsandemir:
Windows tests sometimes fail due to memory access during client shutdown. The problem is isolated down to ClusterService shutdown and particularly to the deletion of the cluster listener thread. With some added logs we see the following:
```
12:09:20 Dec 19, 2017 09:09:16 AM INFO: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] LifecycleService::LifecycleEvent SHUTTING_DOWN
12:09:20 Dec 19, 2017 09:09:16 AM INFO: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] Shutting down InvocationService
12:09:20 Dec 19, 2017 09:09:16 AM INFO: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] Shutting down ConnectionManager
12:09:20 Dec 19, 2017 09:09:16 AM WARNING: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] Closing connection (id:2) to Address[127.0.0.1:5702] with socket id 144. Hazelcast client is shutting down
12:09:20 Dec 19, 2017 09:09:16 AM WARNING: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] Closing connection (id:3) to Address[127.0.0.1:5701] with socket id 160. Hazelcast client is shutting down
12:09:20 Dec 19, 2017 09:09:16 AM WARNING: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] Closing connection (id:1) to Address[127.0.0.1:5701] with socket id 340 as the owner connection.Closing owner connection to Address[127.0.0.1:5701]
12:09:20 Dec 19, 2017 09:09:16 AM WARNING: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [448] hz.heartbeater is cancelled
12:09:20 Dec 19, 2017 09:09:16 AM INFO: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] Shutting down PartitionService
12:09:20 Dec 19, 2017 09:09:16 AM INFO: [HazelcastCppClient3.9.1-SNAPSHOT] [dev] [5024] Shutting down ClusterService
[...]

PR #370, opened 2017-12-20
Client proper shutdown fix (windows)

ihsandemir:
Removed connection resets. Added connection alive check to be used in addition to the null check at the ClusterListenerThread thread loop.

The problem case:
1. Due to connection reset, the connection is destroyed and the underlying receiveBuffer is deleted.
2.  ClusterListenerThread is still at listenMembershipEvents call blocked at the readBlocking call which uses the deleted receiveBuffer of the connection. The receiveBuffer is accessed after the blocking read and this is an illegal memory access.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/369

Attached is the logs that I added which were used at windows to determine where the problem was.
[windows_failure_with_logs_added_issue_369.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/1579628/windows_failure_with_logs_added_issue_369.txt)

### DIFF
--- a/hazelcast/src/hazelcast/client/connection/ClusterListenerThread.cpp
+++ b/hazelcast/src/hazelcast/client/connection/ClusterListenerThread.cpp
@@ -57,7 +57,7 @@ namespace hazelcast {
                 spi::LifecycleService &lifecycleService = clientContext.getLifecycleService();
                 while (lifecycleService.isRunning()) {
                     try {
-                        if (conn.get() == NULL) {
+                        if (conn.get() == NULL || !conn->live) {
                             try {
                                 conn = clientContext.getClusterService().connectToOne(previousConnectionAddrPtr);
                                 previousConnectionAddr = conn->getRemoteEndpoint();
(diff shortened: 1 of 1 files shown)


hazelcast/src/hazelcast/util/Thread.cpp
```diff
@@ -61,8 +61,8 @@ namespace hazelcast {
             if (!isJoined) {
                 cancel();
                 join();
+                CloseHandle(thread);
```
sancar: Should not we close the handle if thread is already joined when destructor called ?

ihsandemir: If the thread is joined already this means that the above code inside if statement executed, hence the handle is closed. Closing the handle twice it may cause as problem as indicated here: https://msdn.microsoft.com/en-us/library/windows/desktop/ms724211(v=vs.85).aspx
"the function will throw an exception if it receives either a handle value that is not valid or a pseudo-handle value. This can happen if you close a handle twice"

PR #381, opened 2018-03-06
Can not write objects to logger without using ostringstream

ihsandemir:
This PR extends the logger with stream operator support so that you can easily log different objects and it is appendable.

### DIFF
--- a/hazelcast/include/hazelcast/client/ClientConfig.h
+++ b/hazelcast/include/hazelcast/client/ClientConfig.h
@@ -286,10 +286,18 @@ namespace hazelcast {
             *  set SEVERE to see only serious errors
             *
             *  Default log level is INFO
+            * @param loggerLevel The log level to be set.
             * @return itself ClientConfig
             */
-            ClientConfig &setLogLevel(LogLevel loggerLevel);
+            ClientConfig &setLogLevel(LoggerLevel::Level loggerLevel);
 
+            /**
+             * @deprecated Please use setLogLevel(LoggerLevel::Level loggerLevel) instead.
+             *
+             * @param loggerLevel The log level to be set.
+             * @return The configured client configuration for chaining.
+             */
+            ClientConfig &setLogLevel(LogLevel loggerLevel);
 
             /**
             *
--- a/hazelcast/include/hazelcast/client/ClientConfig.h
+++ b/hazelcast/include/hazelcast/client/ClientConfig.h
@@ -280,22 +280,14 @@ namespace hazelcast {
             ClientConfig &setLoadBalancer(LoadBalancer *loadBalancer);
 
             /**
-            *  enum LogLevel { SEVERE = 100, WARNING = 90, INFO = 50 };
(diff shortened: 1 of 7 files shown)


hazelcast/src/hazelcast/util/ILogger.cpp
```diff
@@ -115,7 +130,47 @@ namespace hazelcast {
...
+        LeveledLogger::LeveledLogger(ILogger &logger, client::LoggerLevel::Level logLevel) : logger(logger) {
+            if (logger.isEnabled(logLevel)) {
+                logger.lockMutex.lock();
+                logger.printMessagePrefix(logLevel);
+            }
+        }
+
+        LeveledLogger::~LeveledLogger() {
+            std::cout << std::endl;
+            logger.lockMutex.unlock();
```
sancar: The constructor takes the lock if only enabled.
The destructor always unlocks it.

hazelcast/include/hazelcast/client/ClientConfig.h
```diff
@@ -286,10 +286,18 @@ namespace hazelcast {
             *  set SEVERE to see only serious errors
             *
             *  Default log level is INFO
+            * @param loggerLevel The log level to be set.
             * @return itself ClientConfig
             */
-            ClientConfig &setLogLevel(LogLevel loggerLevel);
+            ClientConfig &setLogLevel(LoggerLevel::Level loggerLevel);
```
sancar: Is this new method really necessary? 
From the code, it looks like old one works just fine. Why are we deprecating that and introducing a new one?

hazelcast/src/hazelcast/util/ILogger.cpp
ihsandemir: good finding 👍 Fixed it.

hazelcast/include/hazelcast/client/ClientConfig.h
ihsandemir: Because when using the old one it is directly at the hazelcast::client namespace, hence you use for example isEnabled(WARNING), this seems like a global variable but I find it better looking usage like this: isEnabled(LoggerLevel::WARNING). It is really a matter of taste.

ihsandemir: Keeping the ClientConfig api the same but changing the internal ILogger api and removing the deprecated one since it is an internal API.

hazelcast/include/hazelcast/client/ClientConfig.h
```diff
@@ -280,17 +280,12 @@ namespace hazelcast {
...
             /**
-            *  enum LogLevel { SEVERE = 100, WARNING = 90, INFO = 50 };
-            *  set INFO to see every log.
-            *  set WARNING to see only possible warnings and serious errors.
-            *  set SEVERE to see only serious errors
-            *
-            *  Default log level is INFO
-            * @return itself ClientConfig
-            */
+             *
```
sancar: Don't forget to keep the javadoc here. I think it is deleted by mistake.

PR #383, opened 2018-03-07
Missing random uuid generator and address generator classes as defined at Java client

ihsandemir:
Added UuidUtil and AddressHelper classes. UuidUtil lets you to generate random uuid. AddressHelper gets possible address list for a provided hostname. These will be used by the client services.


ihsandemir:
verify-linux

PR #384, opened 2018-03-09
Future::get is not throwing the expected exception as derived but the base exception - Missing stream like constructor for an exception

ihsandemir:
Added tests so that the future.get throws the expected protocol exception correctly. I had to introduce the IException::clone method to create derived exception when setting the future.set_exception so that when it is thrown it throws the derived exception.

Implemented the ExceptionBuilder class for easier setting of the exception message.

### DIFF
--- a/hazelcast/include/hazelcast/client/exception/IException.h
+++ b/hazelcast/include/hazelcast/client/exception/IException.h
@@ -85,7 +85,7 @@ namespace hazelcast {
 
                 const boost::shared_ptr<IException> &getCause() const;
 
-                friend std::ostream &operator<<(std::ostream &os, const IException &exception);
+                friend std::ostream HAZELCAST_API &operator<<(std::ostream &os, const IException &exception);
 
                 virtual bool operator==(const IException &rhs) const;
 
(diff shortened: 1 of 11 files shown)


hazelcast/include/hazelcast/client/exception/IException.h
```diff
@@ -48,36 +54,76 @@ namespace hazelcast {
...
+                 * We need this method to clone the specific derived exception when needed. The exception has to be the
+                 * derived type so that the exception rethrowing works as expected by throwing the derived exception.
+                 * Exception throwing internals works by making a temporary copy of the exception.
+                 * @return The copy of this exception
+                 */
+                virtual std::auto_ptr<IException> clone() const;
+
+                const boost::shared_ptr<IException> &getCause() const;
+
+                friend std::ostream HAZELCAST_API &operator<<(std::ostream &os, const IException &exception);
```
sancar: After introducing `ExceptionBuilder`, It looks like we should not need this method anymore, am I right?

ihsandemir: This the stream operator used when you want to print the exception, which is equivalent of toString method at Java. I think we may still need it. The operator ExceptionBuilder uses is serving another purpose to compose the exception message.

PR #386, opened 2018-03-14
Missing executor for C++ client

ihsandemir:
Implemented simple executor. Introduced Runnable and Callable. Implemented startable thread.
Separated the windows and non-windows specific code to appropriate inline files.
Changed the windows thread to work with the InterruptedException rather than a custom exception.
Other util changes.


hazelcast/src/hazelcast/util/impl/Executor.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+            void SimpleExecutorService::startWorkers() {
+                // `maximumQueueCapacity` is the given max capacity for this executor. Each worker in this executor should consume
+                // only a portion of that capacity. Otherwise we will have `threadCount * maximumQueueCapacity` instead of
+                // `maximumQueueCapacity`.
+                int32_t perThreadMaxQueueCapacity = static_cast<int32_t>(ceil(
+                        (double) 1.0 * maximumQueueCapacity / threadCount));
+                for (int i = 0; i < threadCount; i++) {
+                    workers[i].reset(new Worker(threadNamePrefix, perThreadMaxQueueCapacity, live, logger));
+                    workers[i]->start();
```
sancar: From the current usage, it looks like there is no need for StartedThread and Thread differentiation. It looks like we can use old Thread class without any change because constructing a thread with is always together with starting it.  
What is the reason for introducing `StartedThread`?

ihsandemir: For some of the spi services, there is a start method where I start a thread but I construct the thread at the service constructor.

My executor implementation is missing the management of workers as being done at Java where the workers are being started on need basis. I want to also change it to this version later on.

We can always achieve the same purpose by using smart pointers and with some tricks but I find it better usage starting the the thread at another call which is the way chosen by Java. 

At this PR we provide both ways of usage, hence it should not be a problem. Name StartedThread implies better that the thread is started by construction.

hazelcast/include/hazelcast/util/Future.h
```diff
@@ -111,13 +193,28 @@ namespace hazelcast {
                 resultReady = false;
                 exceptionReady = false;
             }
-        private:
+
+            void andThen(const boost::shared_ptr<client::impl::ExecutionCallback<T> > &callback, util::Executor &executor) {
+                LockGuard guard(mutex);
+                callbacks.push_back(CallbackInfo(callback, executor));
```
sancar: Implementation seems missing here. What if the response is already set to the future when andThen is called?

hazelcast/include/hazelcast/util/Future.h
```diff
@@ -40,20 +45,72 @@ namespace hazelcast {
...
+            }
+
             void set_value(T& value) {
                 LockGuard guard(mutex);
                 if (exceptionReady || resultReady) {
                     util::ILogger::getLogger().warning(std::string("Future.set_value should not be called twice"));
                 }
                 sharedObject = value;
                 resultReady = true;
+                onComplete();
```
sancar: What is the use of onComplete? 
On java, this abstraction is necessary because the same invocationFuture is used by both clients and members? 
Here we don't really need that.

hazelcast/include/hazelcast/util/Future.h
```diff
@@ -90,7 +172,7 @@ namespace hazelcast {
 
             /**
              *
-             * @return true if result or exception is ready. false otherwise when timeout expires.
+             * @return true if result or exception is ready. git pufalse otherwise when timeout expires.
```
sancar: git pufalse :)

hazelcast/include/hazelcast/util/impl/AbstractThread.h
```diff
@@ -0,0 +1,79 @@
...
+
+#if  defined(WIN32) || defined(_WIN32) || defined(WIN64) || defined(_WIN64)
+#pragma warning(push)
+#pragma warning(disable: 4251) //for dll export
+#endif
+
+namespace hazelcast {
+    namespace util {
+        namespace impl {
+            class HAZELCAST_API AbstractThread : public Runnable {
```
sancar: I think Thread should not implement Runnable. In Java, it is that way because of backward compatibility.
See https://stackoverflow.com/questions/18305953/why-does-thread-implement-runnable

sancar: I just realized that Thread being Runnable is used in AbstractThread::start.

hazelcast/include/hazelcast/util/impl/AbstractThread.h
```diff
@@ -0,0 +1,79 @@
...
+            public:
+                AbstractThread(const std::string &name);
```
sancar: Is it possible to remove this constructor?

hazelcast/include/hazelcast/util/Future.h
ihsandemir: Java calls callIdSequence.complete(); on completion of ClientInvocationFuture, this is why we need it.

hazelcast/include/hazelcast/util/impl/AbstractThread.h
ihsandemir: I will remove the Runnable interface from Thread. I will also remove StartedThread and make everything Thread which accepts shared_ptr<Runnable>.

hazelcast/include/hazelcast/util/impl/AbstractThread.h
ihsandemir: Since I am removing the Runnable interface from Thread, yes, I may be able to remove it since Runnable already has a name.

ihsandemir:
verify-linux

hazelcast/include/hazelcast/util/Future.h
ihsandemir: Corrected and added tests.

PR #399, opened 2018-04-25
Update client release version since Hazelcast 3.10 is released

ihsandemir:
Changed client version from 3.9.2 to 3.10. Also updated the test server to 3.10.

Changed ClientSemaphoreTest so that the behaviour change for Hazelcast 3.10 for negative permits is allowed and the test passes for this case.


ihsandemir:
verify-linux

Issue #227, opened 2017-02-14
Improvements for Invocation and Retry Mechanisms - LoadTest may sometimes fail


ihsandemir:
C++ client should be able to retry invocations which can not be sent to the server side if it is retryable.

Change C++ invocation design to match for Java invocation design (implement Smart and Non-Smart InvocationService)

Some test runs show failures for LoadTest:

Stacktrace

/home/jenkins/jenkins/workspace/cpp-centos7-nightly-32-STATIC-Debug/hazelcast/test/src/faulttolerance/LoadTest.cpp:61
Expected: val = map->get(i) doesn't throw an exception.
  Actual: it throws.

Build: https://hazelcast-l337.ci.cloudbees.com/job/cpp-centos7-nightly-32-STATIC-Debug/41/testReport/junit/(root)/LoadTest/testIntMapSmartClientServerRestart/

When examined the problem, we see that the invocation system does not retry the failures when no connection exist to the target when config->setRedoOperation(true) is used. 

Example: Possible exception when InvocationService::invokeOnPartitionOwner called is not being retried.

Solution: Change the invocation system to Java invocations and use the same retry mechanism.

ihsandemir:
fixed by https://github.com/hazelcast/hazelcast-cpp-client/pull/392

PR #406, opened 2018-05-07
Enable the load test

ihsandemir:
Enabled the load test since https://github.com/hazelcast/hazelcast-cpp-client/pull/392 fixes the problem and also decreased runtime by decreasing number of operations down to 1000 per thread.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/227


sancar (dismissed):

ihsandemir:
no more valid.

PR #407, opened 2018-05-08
Increase test coverage

ihsandemir:
Removed some unused codecs and added more tests for coverage. Also some fixes found during tests.


ihsandemir:
verify-linux

PR #412, opened 2018-05-17
Race on client message buffer on retries

ihsandemir:
Ports the following fix from Java client https://github.com/hazelcast/hazelcast/pull/13072


ihsandemir:
verify-windows

Issue #4, opened 2015-10-30
Decrement Pending packets for events also


sancar:
An issue found in java client.
https://github.com/hazelcast/hazelcast/pull/6608
Needs to be checked if same issue is here.

ihsandemir:
This is most probably fixed with test PR https://github.com/hazelcast/hazelcast-cpp-client/pull/392

Will duplicate the Java test testPendingEventPacketsWithEvents to verify.

ihsandemir:
We decided to remove pendingPacketCount completely from Java client as well. The removal PR is: https://github.com/hazelcast/hazelcast/pull/13221/files

Hence, we will do the same for cpp client as well.

PR #419, opened 2018-05-30
Decrements the connection pending packet count on event receive

ihsandemir:
Decrements the connection pending packet count as expected when events are received.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/4


sancar:
Pending count is no longer is used on java client. It seems these parts are leftover. I think we should consider removing all logic. Wdyt ?

ihsandemir:
verify-linux

ihsandemir:
Decided to remove this field completely. See https://github.com/hazelcast/hazelcast/pull/13221

PR #420, opened 2018-05-31
Remove pending packet count from client connection since it has no use

### LINKED ISSUE #4: Decrement Pending packets for events also

ihsandemir:
Removing pending packet count from client connection since it has no use anymore.

fixes #4

### DIFF
--- a/hazelcast/src/hazelcast/client/spi/impl/AbstractClientInvocationService.cpp
+++ b/hazelcast/src/hazelcast/client/spi/impl/AbstractClientInvocationService.cpp
@@ -279,15 +279,11 @@ namespace hazelcast {
 
                 void AbstractClientInvocationService::ResponseThread::process(
                         const AbstractClientInvocationService::ClientPacket &packet) {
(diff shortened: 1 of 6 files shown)


hazelcast/src/hazelcast/client/spi/impl/AbstractClientInvocationService.cpp
```diff
@@ -279,15 +279,11 @@ namespace hazelcast {
 
                 void AbstractClientInvocationService::ResponseThread::process(
                         const AbstractClientInvocationService::ClientPacket &packet) {
```
sancar: Can you completely remove the packet and pass only ClientMessage here ?

ihsandemir: Ok, removed ClientPacket and used ClientMessage.

PR #426, opened 2018-06-13
testReconnectModeASYNCTwoMembers test may fail with map.get throwing TargetNotMemberException

ihsandemir:
Problem definition: testReconnectModeASYNCTwoMembers test may fail with map.get throwing TargetNotMemberException.

Two fixes:
1. We need to check RetryableHazelcastException more properly for the derived classes such as TargetNotMemberException. Hence, I throw and catch to determine if the exception is derived from RetryableHazelcastException.
2. ClientPartitionServiceImpl::RefreshTaskCallback::onResponse is fixed for correct null checking.


hazelcast/src/hazelcast/client/spi/impl/ClientPartitionServiceImpl.cpp
```diff
@@ -35,7 +35,7 @@ namespace hazelcast {
 
                 void ClientPartitionServiceImpl::RefreshTaskCallback::onResponse(
                         const boost::shared_ptr<protocol::ClientMessage> &responseMessage) {
-                    if (responseMessage.get()) {
+                    if (!responseMessage.get()) {
```
sancar: Old code seems correct. If it is null, it should return.

PR #438, opened 2018-07-18
Implement CRDT PN Counter structure

ihsandemir:
Implemented the CRDT PN Counter. The tests are added.


hazelcast/include/hazelcast/client/cluster/impl/VectorClock.h
```diff
@@ -0,0 +1,92 @@
...
+            namespace impl {
+                /**
+                 * Vector clock consisting of distinct replica logical clocks.
+                 * <p>
+                 * See https://en.wikipedia.org/wiki/Vector_clock
+                 * The vector clock may be read from different thread but concurrent
+                 * updates must be synchronized externally. There is no guarantee for
+                 * concurrent updates.
+                 */
+                class VectorClock : public serialization::IdentifiedDataSerializable {
```
sancar: We don't send this to remote so This does not need to be Serilazable, right ?

ihsandemir: Removed it.

hazelcast/src/hazelcast/client/cluster/impl/VectorClock.cpp
```diff
@@ -0,0 +1,83 @@
...
+                        boost::shared_ptr<int64_t> localReplicaTimestamp = getTimestampForReplica(replicaId);
+
+                        if (localReplicaTimestamp.get() == NULL || localReplicaTimestamp < otherReplicaTimestamp) {
+                            return false;
+                        } else if (*localReplicaTimestamp > *otherReplicaTimestamp) {
+                            anyTimestampGreater = true;
+                        }
+                    }
+                    // there is at least one local timestamp greater or local vector clock has additional timestamps
+                    return anyTimestampGreater || !containsAll(other.replicaTimestamps.keys(), replicaTimestamps.keys());
```
mustafaiman: if you reach here it is guaranteed that `replicaTimstamps` has all the keys that `other.replicaTimestamps` has. You can simply check if number of keys in `replicaTimestamps` is greater than the number of keys in `other.replicaTimestamps` instead of `containsAll`

ihsandemir: Yes, that can be done. But should we do this for Java client first?: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/cluster/impl/VectorClock.java#L104

We can also break the loop at line 56 where the anyTimestampGreater is true.

Should we do these two first at Java and then at C++ client?

hazelcast/include/hazelcast/client/DistributedObject.h
```diff
@@ -64,7 +65,7 @@ namespace hazelcast {
         * @see TransactionalSet
         * @see TransactionalList
         */
-        class HAZELCAST_API DistributedObject : public IDistributedObject {
+        class HAZELCAST_API DistributedObject : public virtual IDistributedObject {
```
sancar: Asking because I am curious. Can you describe why we need this ? What did it solve ?

hazelcast/include/hazelcast/client/cluster/impl/ClusterDataSerializerHook.h
```diff
@@ -31,7 +31,9 @@ namespace hazelcast {
                 enum ClusterDataSerializerHook {
                     F_ID = 0,
 
-                    ADDRESS = 1
+                    ADDRESS = 1,
+
+                    VECTOR_CLOCK = 43
```
sancar: Is this leftover from Vector being IdentifiedDataSerializable ?

hazelcast/src/hazelcast/client/proxy/ClientPNCounterProxy.cpp
```diff
@@ -0,0 +1,351 @@
...
+namespace hazelcast {
+    namespace client {
+        namespace proxy {
+            const std::string ClientPNCounterProxy::SERVICE_NAME = "hz:impl:PNCounterService";
+            const boost::shared_ptr<std::set<Address> > ClientPNCounterProxy::EMPTY_ADDRESS_LIST(
+                    new std::set<Address>());
+
+            ClientPNCounterProxy::ClientPNCounterProxy(const std::string &serviceName, const std::string &objectName,
+                                                       spi::ClientContext *context) : ProxyImpl(serviceName, objectName,
+                                                                                                context),
```
sancar: Can't you make this more human readable ?

hazelcast/src/hazelcast/client/cluster/impl/VectorClock.cpp
sancar: @mustafaiman make sense. I will do this change on hava. 

@ihsandemir I think we can't break the loop. For example following test fails when we break the loop at 56
assertFalse(vectorClock("A", 2).isAfter(vectorClock("A", 1, "B", 1)));

sancar: https://github.com/hazelcast/hazelcast/pull/13477

mmedenjak: Yes, we can't break the loop as those two vector clocks are concurrent, neither is greater than the other.

hazelcast/include/hazelcast/client/DistributedObject.h
ihsandemir: Reason is this: ClientPNCounterProxy inherits two classes (crdt::pncounter::PNCounter and ProxyImpl) . crdt::pncounter::PNCounter is only pure virtual interface. PNCounter extends the interface IDistributedObject and ProxyImpl also implements IDistributedObject. When the derived class inherits the base class through more than one class then we need to make the base class as virtual.

hazelcast/include/hazelcast/client/cluster/impl/ClusterDataSerializerHook.h
ihsandemir: Good catch.

hazelcast/src/hazelcast/client/proxy/ClientPNCounterProxy.cpp
ihsandemir: Ok, changed.

PR #443, opened 2018-08-07
Add example and documentation for Flake Id Generator

ihsandemir:
Flake Id Generator example and documentation is added.

### DIFF
--- a/examples/distributed-primitives/idgenerator/FlakeIdGenerator.cpp
+++ b/examples/distributed-primitives/idgenerator/FlakeIdGenerator.cpp
@@ -14,24 +14,19 @@
  * limitations under the License.
  */
 
- /**
-  * Pleasee NOTE that this id generator usage is deprecated in favor of using Flake Id Generator.
-  * See FlakeIdGenerator.cpp example.
-  */
-
 #include <hazelcast/client/HazelcastClient.h>
-#include <hazelcast/client/IdGenerator.h>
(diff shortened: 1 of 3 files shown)


examples/distributed-primitives/idgenerator/FlakeIdGenerator.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <hazelcast/client/HazelcastClient.h>
+#include <hazelcast/client/FlakeIdGenerator.h>
+
+int main() {
+    hazelcast::client::ClientConfig config;
+    hazelcast::client::HazelcastClient hz(config);
+
+    hazelcast::client::FlakeIdGenerator generator = hz.getFlakeIdGenerator("flakeIdGenerator");
+    for (int i = 0; i < 10000; ++i) {
+        hazelcast::util::sleep(1);
```
SaitTalhaNisanci: Is this in milliseconds ?

ihsandemir: that is seconds. sleepmillis for milliseconds.

SaitTalhaNisanci:
Why does this PR has `Estimation: L` label? Do we consider this PR with its implementation and is that why it has this label?

ihsandemir:
@SaitTalhaNisanci Corrected estimation to S.

Issue #331, opened 2017-09-27
specifying a log file for C++ client


aparnam-asg:
This is more of a question actually. It doesn't look like we can specify a log file to write to in the native client. Console is the only option. However we have a console application which uses this client. And we really don't want to show hazelcast specific output to the user but just send it elsewhere even when it is error level logging.  I wanted to make sure I am not missing something basic, if that is already possible somehow. 
Thanks for your support and time in answering our questions while we are adding these new hazelcast based feature to our product.

ihsandemir:
Hello @aparnam-asg 

Yes, the C++ client logs to the stdout only. Do you have any filtering support in your console application stdout, if yes, may be you can try filtering based on our log format which is standard format for all logs (except member list printing).

I will check if such a support may be added with the product management.

Thanks.

KalebNeguisse:
Hello,

I too have hit this issue, where native code error messages are going directly to the console. Do we have an ETA on when this will be resolved?

Thank you

Kaleb

KalebNeguisse:
IS there any change on this ticket? Please let me know.

PR #457, opened 2018-10-08
Integrate easylogging++ library and introduce logger configuration to allow logging to a file

ihsandemir:
Implemented the LoggerConfig and integrated easylogging++ library to the project.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/331


ihsandemir:
verify-linux

hazelcast/include/hazelcast/client/internal/config/ConfigUtils.h
```diff
@@ -44,10 +44,6 @@ namespace hazelcast {
                         if (configPatternKey.get() != NULL) {
                             return configPatterns.get(*configPatternKey);
                         }
-                        if ("default" != itemName && itemName.find("hz:") != 0) {
```
sancar: Why did we remove this part?

ihsandemir: Since it was a finest level message and otherwise we need to provide logger to this static method, i thought it would be easier t just not print the message. Also, it is being called from ClientConfig which has no associated logger with it, since the logger is per HazelcastClient instance but ClientConfig is created before the instance is created. The method returns null pointer which actually may imply the same thing except that this log is only for when item name starts with "hz:". Do you think that this log should exist and worth the effort?

sancar: On java we use a different logger here via static `Logger.getLogger()`. Can't we do something similar ? 
I mean the logs does not have to contain instance name. It seems we can't also apply users log option also. Not sure if we have to include  this part, I was just comparing to java. Decision is up to you.

ihsandemir: Exactly, we do not have a static logger. Yes, a static logger can be added but what will be its configuration? Write to file? Write to stdout? What will be the log level for this logger and it won't be related to any instance? I believe for Java it uses for example default logger level for log42j. I think it is better that we do not use static loggers like this.

PR #464, opened 2018-10-22
Implement ReplicatedMap structure

ihsandemir:
Implemented ReplicatedMap.


hazelcast/src/hazelcast/client/map/impl/DataAwareEntryEvent.cpp
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "hazelcast/client/map/impl/DataAwareEntryEvent.h"
+
+namespace hazelcast {
```
sancar: Is this file leftover ?

ihsandemir: Yes, it was left over and I just deleted this left-over file.

PR #469, opened 2018-11-09
The ASSERT_EQ_EVENTUALLY macro expression may not compile with complex expressions

ihsandemir:
The ASSERT_EQ_EVENTUALLY macro expression evaluation works better if expression is evaluated using parenthesis since it is checked for equality with the == operand and without the extra parenthesis the equality may not work correctly depending on the expression.

Centos 7 Builds failed due to the following warning:

`ClientReplicatedMapListenerTest.cpp:161:125:[m[K [01;31m[Kerror: [m[Ksuggest parentheses around comparison in operand of ‘[01m[K==[m[K’ [-Werror=parentheses]
02:20:04                  ASSERT_TRUE_EVENTUALLY(listener->keys.size() == 1U && listener->keys.pop() == 1 && listener->addCount.get() == 1);`

Also removed the unused code ASSERT_NE_EVENTUALLY.

### DIFF
--- a/hazelcast/test/src/TestHelperFunctions.h
+++ b/hazelcast/test/src/TestHelperFunctions.h
@@ -23,7 +23,7 @@
 #define ASSERT_EQ_EVENTUALLY(expected, actual) do{              \
             bool result = false;                                \
             for(int i = 0 ; i < 120 * 5 && !result ; i++ ) {    \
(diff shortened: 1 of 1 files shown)


ihsandemir:
verify-windows

ihsandemir:
verify-windows

PR #473, opened 2018-11-20
SocketOptions configuration is missing - Need to set tcp read/write buffer sizes

### LINKED ISSUE #465: Very low performance for IMap.set
If using IMap.set with values larger than 1 MByte the most limiting factor for performance shall be the network bandwidth. On Ubuntu 18.04 in VirtualBox and i5-6300U I first checked the achievable TCP performance on localhost with "qperf localhost tcp_bw tcp_lat":
bw  =  4.26 GB/sec and 
latency  =  56 us.

A good middleware would be able to transfer big key-value pairs with 95% of the TCP bandwidth (e.g. a DDS implementation from RTI).

With the C++-client using the simple IMap<string, vector<uint8_t> >.set(key, value) method I got reproduceable 1.4 MBytes/Sec for values from 1 to 16 MBytes size (s. attachment).

First, I was curious, if Hazelcast as whole has so low performance and checked the same test for the java client and the same instance of Hazelcast-Server. The throughput was a little bit instable, but in the range of 200 MBytes/Sec. It is still only 5% of the available network bandwidth, but factor 140 higher than C++ implementation.

Interestingly the IMap.get in C++ has higher performance of around 90 MBytes/Sec.

In the tests around IMap in C++ I have not found any one checking performance especially for larger kv-pairs, perhaps it is a good idea to put one.

The Cpp-library was built with
cmake .. -DHZ_LIB_TYPE=STATIC -DHZ_BIT=64 -DCMAKE_BUILD_TYPE=Release

In the attachment:
- main.cpp with the c++ test
- Client.java with java test (from code-samples)
- cppIMapPerformance.txt with columns
	messageSize, setDurationInMicroSeconds, setThrouput, getDurationInMicroseconds, getThroughput
- javaIMapPerformance.txt the same for java, but with Milliseconds resolution
- JavaIMapErrOutput.txt output from error-stream during the test run.

[cppIMapPerformance.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/2533226/cppIMapPerformance.txt)
[javaIMapErrOutput.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/2533227/javaIMapErrOutput.txt)
[javaIMapPerformance.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/2533228/javaIMapPerformance.txt)
[Client_java.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/2533233/Client_java.txt)
[main_cpp.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/2533234/main_cpp.txt)

ihsandemir:
Added `SocketOptions` configuration and increased default tcp read/write buffer size from 32KB to 128 KB.

fixes #465


hazelcast/test/src/cluster/SocketOptionsTest.cpp
```diff
@@ -0,0 +1,42 @@
...
+    namespace client {
+        namespace test {
+            class SocketOptionsTest : public ClientTestSupport {
+            };
+
+            TEST_F(SocketOptionsTest, testConfiguration) {
+                HazelcastServer instance(*g_srvFactory);
+
+                ClientConfig clientConfig;
+                clientConfig.getNetworkConfig().getSocketOptions().setKeepAlive(false).setReuseAddress(true).setTcpNoDelay(false).setLingerSeconds(5).setBufferSize(2 * 1024);
```
SaitTalhaNisanci: what is tested here?

ihsandemir: This is not testing the functionality of each option but just a test to see that the client works when the options are set. This was mostly for coverage. More meaningful functionality tests would be better but I did not come up with a way how to test tcp level such settings yet.

hazelcast/include/hazelcast/client/config/SocketOptions.h
```diff
@@ -0,0 +1,141 @@
...
+namespace hazelcast {
+    namespace client {
+        namespace config {
+            /**
+             * TCP Socket options
+             */
+            class HAZELCAST_API SocketOptions {
+            public:
+                /**
+                 * constant for kilobyte
```
SaitTalhaNisanci: I think we dont need this comment, it doesnt say more than the variable name

hazelcast/include/hazelcast/client/config/SocketOptions.h
```diff
@@ -0,0 +1,141 @@
...
+                 */
+                static const int KILO_BYTE = 1024;
+
+                /**
+                 * default buffer size of Bytes
```
SaitTalhaNisanci: also this one

hazelcast/include/hazelcast/client/config/SocketOptions.h
ihsandemir: These are from Java client documentation. If needed we need to change both. I think it can stay.

hazelcast/include/hazelcast/client/config/SocketOptions.h
ihsandemir: from java client again. https://github.com/hazelcast/hazelcast/blob/master/hazelcast-client/src/main/java/com/hazelcast/client/config/SocketOptions.java#L30

hazelcast/include/hazelcast/client/config/SocketOptions.h
```diff
@@ -0,0 +1,141 @@
...
+                SocketOptions &setReuseAddress(bool reuseAddress);
+
+                /**
+                 * Gets SO_LINGER with the specified linger time in seconds
+                 * @return lingerSeconds value in seconds
+                 */
+                int getLingerSeconds() const;
+
+                /**
+                 * Enable/disable SO_LINGER with the specified linger time in seconds
```
SaitTalhaNisanci: What value disables `SO_LINGER`?

ihsandemir: Added doc: ```
                *
                 * if set to a value of 0 or less then it is disabled.
                 *
                 * Default value is 3.
```

File: hazelcast/include/hazelcast/client/config/SocketOptions.h [hunk 5]
```diff
@@ -0,0 +1,141 @@
...
+                 * @param lingerSeconds value in seconds
+                 * @return SocketOptions configured
+                 */
+                SocketOptions &setLingerSeconds(int lingerSeconds);
+
+                /**
+                 * Gets the SO_SNDBUF and SO_RCVBUF options to the specified value in KB
+                 * @return bufferSize KB value
+                 */
+                int getBufferSize() const;
```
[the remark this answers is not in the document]
ihsandemir: it is actually bytes. Corrected the documentation now.

hazelcast/src/hazelcast/client/internal/socket/TcpSocket.cpp
```diff
@@ -296,6 +283,51 @@ namespace hazelcast {
                         return std::auto_ptr<Address>();
                     }
                 }
+
+                SocketInterface &TcpSocket::setSocketOptions(const client::config::SocketOptions &socketOptions) {
+                    int optionValue = socketOptions.getBufferSize();
```
SaitTalhaNisanci: I think for better readability instead of using `optionValue` we can use directly the option's name, for example here `int bufferSizeInKB = socketOptions.getBufferSizeInKB()`

hazelcast/include/hazelcast/client/config/SocketOptions.h
```diff
@@ -0,0 +1,141 @@
...
+
+                /**
+                 * Sets the SO_SNDBUF and SO_RCVBUF options to the specified value in KB
+                 *
+                 * @param bufferSize KB value
+                 * @return SocketOptions configured
+                 */
+                SocketOptions &setBufferSize(int bufferSize);
```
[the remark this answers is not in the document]
ihsandemir: It is bytes. Changed doc.

hazelcast/src/hazelcast/client/internal/socket/TcpSocket.cpp
ihsandemir: I named it this way to reuse the same local variable for each option instead of stack allocating a new one for each.

hazelcast/include/hazelcast/client/config/SocketOptions.h
SaitTalhaNisanci: I think they are not even used in JAVA :) I sent a PR for that `https://github.com/hazelcast/hazelcast/pull/14147/files`

hazelcast/include/hazelcast/client/config/SocketOptions.h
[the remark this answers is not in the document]
SaitTalhaNisanci: if it is bytes I think method name should be `setBufferSizeInBytes`

hazelcast/include/hazelcast/client/config/SocketOptions.h
SaitTalhaNisanci: As they are part of API we cant change them I guess. If there isnt any check like checkstyle for CPP we shouldnt have them here. They are probably there in Java because of checkstyle(otherwise the build fails).

PR #488, opened 2018-12-05
Fixes how the executor service is being shutdown. Added the missing a…

ihsandemir:
Fixes how the executor service is being shutdown. Added the missing await termination logic to the executor. Failing to put an upper bound on thread finish termination caused the client shutdown get stuck rarely in some builds. Also added some debug logs.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/434


hazelcast/include/hazelcast/util/SynchronizedQueue.h
```diff
@@ -50,6 +51,21 @@ namespace hazelcast {
...
 
+            size_t size() {
+                util::LockGuard lg(m);
+                return internalQueue.size();
+            }
+
+            std::vector<boost::shared_ptr<T> > values() {
+                util::LockGuard lg(m);
+                std::vector<boost::shared_ptr<T> > values;
+                for (typename std::deque<boost::shared_ptr<T> >::const_iterator it = internalQueue.begin();
```
SaitTalhaNisanci: If we had something like:

```
for (auto &it: internalQueue){
values.push_back(it);
}
```

Would it have the behavior?

hazelcast/src/hazelcast/util/impl/Executor.cpp
```diff
@@ -84,18 +86,61 @@ namespace hazelcast {
...
+                    runner->wakeup();
                 }
             }
 
+            bool SimpleExecutorService::awaitTerminationSeconds(int timeoutSeconds) {
+                return awaitTerminationMilliseconds(timeoutSeconds * CountDownLatch::MILLISECONDS_IN_A_SECOND);
+            }
+
+            bool SimpleExecutorService::awaitTerminationMilliseconds(int64_t timeoutMilliseconds) {
+                int64_t endTime = currentTimeMillis() + timeoutMilliseconds;
```
SaitTalhaNisanci: endTimeMilliseconds

hazelcast/src/hazelcast/util/impl/Executor.cpp
```diff
@@ -84,18 +86,61 @@ namespace hazelcast {
                     return;
                 }
 
+                size_t numberOfWorkers = workers.size();
+                size_t numberOfDelayedRunners = delayedRunners.size();
+
+                logger.finest() << "ExecutorService " << threadNamePrefix << " has " << numberOfWorkers << " workers and "
```
sancar: Don't we have `isFinest` checks like we have in java?

hazelcast/include/hazelcast/util/SynchronizedQueue.h
ihsandemir: We use BOOST_FOREACH but I can not put it here at this header not to cause external dependency sşnce this header may be accessed via user included headers. But we use BOOST_FOREACH internally.

hazelcast/src/hazelcast/util/impl/Executor.cpp
ihsandemir: yes, we have it. Since this was during shutdown, I did not think optimization is important at this line. Should we use it?

hazelcast/src/hazelcast/util/impl/Executor.cpp
```diff
@@ -189,27 +232,32 @@ namespace hazelcast {
...
+
+            SimpleExecutorService::RepeatingRunner::RepeatingRunner(SimpleExecutorService &executorService,
                     const boost::shared_ptr<util::Runnable> &command, int64_t initialDelayInMillis,
-                    int64_t periodInMillis, util::ILogger &logger) : DelayedRunner(command, initialDelayInMillis, periodInMillis, logger) {
+                    int64_t periodInMillis, util::ILogger &logger) : DelayedRunner(executorService, command,
+                            initialDelayInMillis, periodInMillis, logger) {
             }
 
             const std::string SimpleExecutorService::RepeatingRunner::getName() const {
                 return command->getName();
```
sancar: It seems that we are passing the executorService for `threadNamePrefix` but it is forgotten to be used here. 
And if the only purpose is `threadNamePrefix`, we should only pass that on the constructor.

ihsandemir: Deleted RepeatingRunner class.

hazelcast/src/hazelcast/util/impl/Executor.cpp
sancar: If we add the check as a general rule that would be great. That way we don't need to think if it is necessary or not.

PR #491, opened 2018-12-07
Add the soak test

ihsandemir:
Added the soak test codes. It is a standalone cmake project and works with the following cmake variables:

-DHAZELCAST_INSTALL_DIR=<Hazelcast Installation Directory> (This is the cpp folder when the released library zip is opened)
-DHZ_VERSION=<Version being tested> (E.g. 3.11-SNAPSHOT)


ihsandemir:
verify-linux

PR #493, opened 2018-12-11
Handle null message field in protocol exception factory

ihsandemir:
Fix for handling exception creation when the message is null. Also, changed the remote controller version from 0.4 to 0.3.


scripts/start-rc.sh
```diff
@@ -19,7 +19,7 @@ HZ_VERSION="3.11"
 HAZELCAST_TEST_VERSION=${HZ_VERSION}
 HAZELCAST_VERSION=${HZ_VERSION}
 HAZELCAST_ENTERPRISE_VERSION=${HZ_VERSION}
-HAZELCAST_RC_VERSION="0.4-SNAPSHOT"
+HAZELCAST_RC_VERSION="0.3-SNAPSHOT"
```
SaitTalhaNisanci: what is the reason for this change?

ihsandemir: Because the compatibility tests are using this lower version and I wanted both to be the same version. I

PR #495, opened 2018-12-27
[DO NOT MERGE - WIP] Executor Service Implementation

ihsandemir:
Added partial interfaces for the executor service.

Implemented the IFuture and the same implementation of ClientInvocationFuture as in Java.


ihsandemir:
Implemented and replaced by PR https://github.com/hazelcast/hazelcast-cpp-client/pull/500

PR #496, opened 2018-12-28
Need to provide different Future implementations for submit methods of the new ClientExecutionService

ihsandemir:
Added the IFuture interface and added the same implementation of ClientInvocationFuture as in Java.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/497


hazelcast/include/hazelcast/client/IFuture.h
```diff
@@ -0,0 +1,136 @@
...
+         * the computation.  The result can only be retrieved using method
+         * {@code get} when the computation has completed, blocking if
+         * necessary until it is ready.  Cancellation is performed by the
+         * {@code cancel} method.  Additional methods are provided to
+         * determine if the task completed normally or was cancelled. Once a
+         * computation has completed, the computation cannot be cancelled.
+         *
+         *
+         * <p>Memory consistency effects: Actions taken by the asynchronous computation
+         * <a href="package-summary.html#MemoryVisibility"> <i>happen-before</i></a>
```
sancar: This link is specific to java

hazelcast/include/hazelcast/client/IFuture.h
```diff
@@ -0,0 +1,136 @@
...
+         * actions following the corresponding {@code Future::get()} in another thread.
+         *
+         * @param <V> The result type returned by this Future's {@code get} method
+         */
+        template<typename V>
+        class IFuture {
```
sancar: Missing virtual destructor

hazelcast/include/hazelcast/client/ICompletableFuture.h
```diff
@@ -0,0 +1,69 @@
...
+    namespace client {
+
+        /**
+         * A Future where one can asynchronously listen on completion. This functionality is needed for the
+         * reactive programming model.
+         *
+         * @param <V> The result type returned by this Future's {@code get} method
+         */
+        template<typename V>
+        class ICompletableFuture : public IFuture<V> {
```
sancar: Missing virtual destructor

hazelcast/include/hazelcast/client/spi/InternalCompletableFuture.h
```diff
@@ -0,0 +1,57 @@
...
+namespace hazelcast {
+    namespace client {
+        namespace spi {
+            /**
+             * A {@link ICompletableFuture} with more functionality
+             *
+             * @param <T>
+             */
+             template <typename T>
+            class InternalCompletableFuture : public ICompletableFuture<T> {
```
sancar: Missing virtual destructor

hazelcast/include/hazelcast/client/spi/impl/AbstractInvocationFuture.h
```diff
@@ -0,0 +1,534 @@
...
+                                new ExceptionState(exception));
+
+                        return innerComplete(newState);
+                    }
+
+                protected:
+                    typedef ExecutionCallback<T> CALLBACKTYPE;
+                    
+                    class BaseState {
+                    public:
```
sancar: Missing virtual destructor

hazelcast/src/hazelcast/client/spi/impl/ClientInvocation.cpp
```diff
@@ -29,96 +29,100 @@ namespace hazelcast {
...
                                                    const std::string &objectName,
                                                    int partitionId) :
-                        ClientInvocationFuture(clientContext.getLogger()),
                         logger(clientContext.getLogger()),
                         lifecycleService(clientContext.getLifecycleService()),
                         clientClusterService(clientContext.getClientClusterService()),
                         invocationService(clientContext.getInvocationService()),
-                        executionService(clientContext.getClientExecutionService()),
+                        executionService(clientContext.getClientExecutionService().shared_from_this()),
                         clientMessage(boost::shared_ptr<protocol::ClientMessage>(clientMessage)),
```
sancar: You already get a shared_ptr. This extra wrapping seems to be not necessary anymore. Or the change on the `ClientInvocation` constructor is not necessary.

hazelcast/src/hazelcast/client/spi/impl/ClientInvocation.cpp
```diff
@@ -360,55 +373,12 @@ namespace hazelcast {
                 }
 
                 void ClientInvocation::onComplete() {
```
sancar: This method can be removed. There is no need for it anymore.

hazelcast/src/hazelcast/util/concurrent/CancellationException.cpp
```diff
@@ -26,6 +26,10 @@ namespace hazelcast {
             void CancellationException::raise() const {
                 throw *this;
             }
+
+            int32_t CancellationException::getErrorCode() const {
+                return client::protocol::CANCELLATION;
```
sancar: We are providing the error code in the constructor as client::protocol::ILLEGAL_STATE. Then we are overriding it with this method. Can't we just pass `CANCELLATION` in the constructor to base class and remove `virtual` from IException?

hazelcast/include/hazelcast/client/IFuture.h
ihsandemir: Removed it.

hazelcast/include/hazelcast/client/ICompletableFuture.h
ihsandemir: The derived abstract class (`AbstractInvocationFuture`) has the virtual destructor, hence I thought it was not needed at this interface. Do we also need it at the interface?

hazelcast/include/hazelcast/client/spi/impl/AbstractInvocationFuture.h
ihsandemir: I has no virtual method and hence should not need the virtual destructor, does it?

hazelcast/include/hazelcast/client/impl/HazelcastClientInstanceImpl.h
```diff
@@ -420,15 +420,15 @@ namespace hazelcast {
                 std::auto_ptr<internal::nearcache::NearCacheManager> nearCacheManager;
                 spi::impl::ClientClusterServiceImpl clusterService;
                 boost::shared_ptr<spi::impl::ClientPartitionServiceImpl> partitionService;
-                std::auto_ptr<spi::impl::ClientExecutionServiceImpl> executionService;
+                boost::shared_ptr<spi::impl::ClientExecutionServiceImpl> executionService;
```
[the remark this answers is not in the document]
ihsandemir: Yes, it works since the shared_ptr constructor which accepts the auto_ptr a argument takes ownership of the auto_ptr. I will change the method signature to make this more visible.

hazelcast/src/hazelcast/client/spi/impl/ClientInvocation.cpp
ihsandemir: needs this since reference returned at here:
`impl::ClientExecutionServiceImpl &ClientContext::getClientExecutionService() const`

I will make another PR which will change the `ClientContext::getClientExecutionService()` method to return shared_ptr.

hazelcast/src/hazelcast/client/spi/impl/ClientInvocation.cpp
ihsandemir: Removed.

hazelcast/src/hazelcast/util/concurrent/CancellationException.cpp
ihsandemir: OK, changed as suggested.

hazelcast/include/hazelcast/client/ICompletableFuture.h
sancar: The rule I use is:
If a class will be used polymorphically it needs a virtual destructor. 

More specifically, If a derived class is deleted via the base class,  and the base class does not have a virtual destructor, the behavior is undefined.

hazelcast/include/hazelcast/client/IFuture.h
ihsandemir: OK, added.

hazelcast/include/hazelcast/client/IFuture.h
```diff
@@ -0,0 +1,135 @@
...
+             * task should be interrupted; otherwise, in-progress tasks are allowed
+             * to complete
+             * @return {@code false} if the task could not be cancelled,
+             * typically because it has already completed normally;
+             * {@code true} otherwise
+             */
+            virtual bool cancel(bool mayInterruptIfRunning) = 0;
+
+            /**
+             * Returns {@code true} if this task was cancelled before it completed
```
SaitTalhaNisanci: If the return parameter explanation is enough for the function why is there a duplication? Both return type and function have the same sentence.

hazelcast/include/hazelcast/client/proxy/RingbufferImpl.h
```diff
@@ -153,9 +179,24 @@ namespace hazelcast {
                      * second ago, then there are 29 seconds remaining for that item. Using this policy you are going to overwrite no matter
                      * what.
                      */
-                     OVERWRITE = 0
+                            OVERWRITE = 0
                 };
 
+                static void checkSequence(int64_t sequence) {
```
SaitTalhaNisanci: I think this method should be `checkNonNegativeSequence` or you could add this as a util method `checkNonNegative(T value, const string& message)` and use it in a lot of places since this will return IllegalArgumentException in most of the cases.

hazelcast/include/hazelcast/client/spi/impl/AbstractInvocationFuture.h
```diff
@@ -0,0 +1,550 @@
...
+                    virtual boost::shared_ptr<T> get() {
+                        boost::shared_ptr<ThreadState> thread = boost::shared_ptr<ThreadState>(
+                                new ThreadState(util::getCurrentThreadId()));
+                        boost::shared_ptr<BaseState> response = registerWaiter(thread, boost::shared_ptr<Executor>());
+                        if (response != VOIDOBJECT) {
+                            // no registration was done since a value is available.
+                            return resolveAndThrowIfException(response);
+                        }
+
+                        for (;;) {
```
SaitTalhaNisanci: what happens if the thread is interrupted here?

hazelcast/include/hazelcast/client/spi/impl/AbstractInvocationFuture.h
```diff
@@ -0,0 +1,550 @@
...
+                            thread->park();
+                            if (isDone()) {
+                                return resolveAndThrowIfException(state);
+                            }
+                        }
+                    }
+
+                    virtual boost::shared_ptr<T> get(int64_t timeout, const TimeUnit &unit) {
```
SaitTalhaNisanci: As far as I see this (time-timeunit) pair is used a lot. It is not the scope of this PR but would it be better to have a class for this so that it would be easier to pass it around and do operations on it.

SaitTalhaNisanci: why isnt timeout also a constant?

hazelcast/include/hazelcast/client/spi/impl/AbstractInvocationFuture.h
```diff
@@ -0,0 +1,550 @@
...
+                                new ThreadState(util::getCurrentThreadId()));
+                        boost::shared_ptr<BaseState> response = registerWaiter(thread, boost::shared_ptr<Executor>());
+                        if (response != VOIDOBJECT) {
+                            return resolveAndThrowIfException(response);
+                        }
+
+                        int64_t deadlineNanos = util::currentTimeNanos() + unit.toNanos(timeout);
+                        int64_t timeoutNanos = unit.toNanos(timeout);
+                        while (timeoutNanos > 0) {
+                            thread->parkNanos(timeoutNanos);
```
SaitTalhaNisanci: here also what will happen if the thread is interrupted?

hazelcast/src/hazelcast/client/mixedtype/Ringbuffer.cpp
```diff
@@ -85,6 +87,14 @@ namespace hazelcast {
 
                 return TypedData(itemData, getContext().getSerializationService());
             }
+
+            void Ringbuffer::checkSequence(int64_t sequence) {
```
SaitTalhaNisanci: why do we have this method twice?

hazelcast/include/hazelcast/client/IFuture.h
ihsandemir: It is from java.util.concurrent.Future doc. I believe the upper description is as a general description of the method while the return statement and parameters would be the details. Yes there seems to be duplication but since Java did it this way so I would go with it.

SaitTalhaNisanci:
There are a lot of changes in this PR, as far as I see the code is not tested that much, have you checked the coverage difference for this PR?

hazelcast/include/hazelcast/client/proxy/RingbufferImpl.h
ihsandemir: We already have the Preconditions::checkNotNegative method. But this method here optimizes so that the error message string is not composed if the value is acceptable and this is a hot path.

hazelcast/include/hazelcast/client/spi/impl/AbstractInvocationFuture.h
ihsandemir: There is no interruption of threads similar to Java. This is how it works: if client is being shutdown, all the awaiting futures are completed with an exception and hence it continues out of the loop.

hazelcast/include/hazelcast/client/proxy/RingbufferImpl.h
SaitTalhaNisanci: I see I am not sure how much of overhead that saves but it makes sense that it is faster. I would still say the method name should be `checkNonNegativeSequence ` because checkSequence does not say what it checks.

hazelcast/include/hazelcast/client/spi/impl/AbstractInvocationFuture.h
SaitTalhaNisanci: I thought making it constant can speed up things as the compiler can make some optimizations if it is constant. So even though it is copied it will be faster.

See `https://stackoverflow.com/questions/16033751/c-const-modifier-with-primitive-types`

ihsandemir:
verify-linux

PR #504, opened 2019-01-23
Add more tests for executor service `execute` methods for coverage

ihsandemir:
Added more tests for executor service `execute` methods. Also, minor fix for not using void which is not serializable for the dummy template parameter.

Tests are adopted from Java ClientExecutorServiceExecuteTest.


ihsandemir:
verify

ihsandemir:
verify-windows

PR #506, opened 2019-01-25
Update the id generator proxy implementation to sync with Java client

ihsandemir:
Updated the id generator proxy implementation with latest changes at the Java client code. Implemented the proxy creation for atomic long.

### DIFF
--- a/hazelcast/src/hazelcast/client/IAtomicLong.cpp
+++ b/hazelcast/src/hazelcast/client/IAtomicLong.cpp
@@ -36,7 +36,7 @@ namespace hazelcast {
         }
 
         int64_t IAtomicLong::getAndAdd(int64_t delta) {
-            return impl->addAndGet(delta);
+            return impl->getAndAdd(delta);
         }
 
         int64_t IAtomicLong::getAndSet(int64_t newValue) {


hazelcast/include/hazelcast/client/IAtomicLong.h
```diff
@@ -13,24 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifndef HAZELCAST_ATOMIC_NUMBER
-#define HAZELCAST_ATOMIC_NUMBER
+#ifndef HAZELCAST_CLIENT_IATOMICLONG_H_
+#define HAZELCAST_CLIENT_IATOMICLONG_H_
```
stalek71: Is it possible to use something like pragma once?

ihsandemir: Yes, that may be a refactoring to simplify headers in the project. I see some caveats here related `pragma once`: http://www.wikizero.biz/index.php?q=aHR0cHM6Ly9lbi53aWtpcGVkaWEub3JnL3dpa2kvUHJhZ21hX29uY2U but i believe we would not suffer from them.

Can you open an issue for it and we can discuss to do it, but it will be on another PR.

Issue #132, opened 2016-05-16
Async API is missing


ihsandemir:
Current C++ API is missing the asyc API such as IMap::getAsync. Needs to be added for all containers for asynchronous operation.

PR #510, opened 2019-01-31
Implement the async API for Hazelcast structures

ihsandemir:
Implemented the async APIs for IMap and IAtomıcLong and added tests for it.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/132

### DIFF
--- a/hazelcast/include/hazelcast/client/IAtomicLong.h
+++ b/hazelcast/include/hazelcast/client/IAtomicLong.h
@@ -170,7 +170,7 @@ namespace hazelcast {
              * or {@code false} if the actual value was not equal to the expected value
              * @since cluster version 3.7
              */
-            virtual boost::shared_ptr<ICompletableFuture<bool> > compareAndSetAsync(int64_t expect, int64_t update);
+            boost::shared_ptr<ICompletableFuture<bool> > compareAndSetAsync(int64_t expect, int64_t update);
 
             /**
              * Atomically decrements the current value by one.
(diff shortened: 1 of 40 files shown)


ihsandemir:
verify-windows

hazelcast/include/hazelcast/client/IAtomicLong.h
```diff
@@ -111,8 +126,127 @@ namespace hazelcast {
...
+             * This method will dispatch a request and return immediately an
+             * {@link ICompletableFuture}.
+             *
+             * @param expect the expected value
+             * @param update the new value
+             * @return an {@link ICompletableFuture} with value {@code true} if successful;
+             * or {@code false} if the actual value was not equal to the expected value
+             * @since cluster version 3.7
+             */
+            virtual boost::shared_ptr<ICompletableFuture<bool> > compareAndSetAsync(int64_t expect, int64_t update);
```
sancar: Why most of these new methods are virtual? As far as I understand IAtomicLong is not further inherited.

ihsandemir: They are virtual methods from the impl::AtomicLongInterface.

sancar: All methods are from `AtomicLongInterface` but only new ones are virtual. 
As far as I know, if we make these virtual when a user calls these methods over IAtomicLong then there will be unnecessary virtual table lookup.  And when I read the code, it tells me that IAtomicLong will be further inherited which is not the case. I am not really concerned about performance impact since I am not sure if it causes a visible penalty, but in terms of readability of the code, it gives the wrong impression.

ihsandemir: Removed the virtual keywords from the method signatures.

PR #513, opened 2019-02-06
Async API implementation for the NearCachedClientMapProxy

ihsandemir:
Implemented ProxyManager::destroy and ClientProxy::onDestroy and let it being called during client shutdown.

Also, minor enhncement for SynchronizedMap so that we do not pre allocate the vector of known size which otherwise would require default empty constructor for the provided key objects but we do not want to enforce it.


ihsandemir:
verify-linux

Issue #515, opened 2019-02-08
VectorClock::replicaTimestamps may not need to be atomic


ihsandemir:
VectorClock::replicaTimestamps does not need to be atomic.

PR #517, opened 2019-02-08
Remove the SyncronizedMap usage from VectorClock

ihsandemir:
Removed the SyncronizedMap usage from VectorClock.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/515

### DIFF
--- a/hazelcast/src/hazelcast/client/cluster/impl/VectorClock.cpp
+++ b/hazelcast/src/hazelcast/client/cluster/impl/VectorClock.cpp
@@ -13,12 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <utility>
 #include <boost/foreach.hpp>
-#include <set>
 
 #include "hazelcast/client/cluster/impl/VectorClock.h"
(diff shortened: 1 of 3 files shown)


hazelcast/src/hazelcast/client/cluster/impl/VectorClock.cpp
```diff
@@ -27,41 +25,45 @@ namespace hazelcast {
...
+                                }
+                }
 
-                std::vector<std::pair<std::string, int64_t> > VectorClock::entrySet() {
-                    TimestampVector entries = replicaTimestamps.entrySet();
-                    std::vector<std::pair<std::string, int64_t> > result;
-                    BOOST_FOREACH(const TimestampVector::value_type &entry , entries) {
-                        result.push_back(std::make_pair(entry.first, *entry.second));
+                VectorClock::TimestampVector VectorClock::entrySet() {
+                    VectorClock::TimestampVector result;
```
stalek71: Please consider reusing TimestampVector passed as a constructor argument. This way you avoid recreating it from map/set

ihsandemir: Changed as suggested.

PR #525, opened 2019-02-18
Implement the missing async API for Ringbuffer

ihsandemir:
Implemented async API for Ringbuffer.


hazelcast/include/hazelcast/util/Preconditions.h
```diff
@@ -71,7 +100,7 @@ namespace hazelcast {
              */
             template<typename T>
             static const boost::shared_ptr<T> &isNotNull(const boost::shared_ptr<T> &argument,
-                                                     const std::string argName) {
+                                                         const std::string argName) {
```
stalek71: Should it be `cont std::string &`? 
I think you don't need additional string copy operation here, because you don't modify string passed here (only read operation is here).

ihsandemir: good catch. fixed it.

hazelcast/test/src/TestHelperFunctions.h
```diff
@@ -51,6 +51,11 @@
 #define WAIT_EQ_EVENTUALLY(expected, expression) WAIT_TRUE_EVENTUALLY((expected) == (expression))
 #define WAIT_NE_EVENTUALLY(expected, expression) WAIT_NE_EVENTUALLY((expected) != (expression))
 
+#define ASSERT_EQ_PTR(expected, actual, type) do{ \
```
stalek71: why do you use `do{}while(0)` here? Isn't sufficient just have `{}` here to create visibility scope (if it was the goal of course :) ) ?

sancar: A common question. Here is the answer 
https://stackoverflow.com/a/257425

ihsandemir:
verify

PR #531, opened 2019-02-26
IAtomicLong async API usage example

ihsandemir:
Implemented IAtomicLong async API usage example.


examples/distributed-primitives/atomiclong/AtomicLongAsyncApi.cpp
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2008-2018, Hazelcast, Inc. All Rights Reserved.
```
sancar: License date needs to be updated.

ihsandemir: Updated.

PR #536, opened 2019-03-01
Add JSON serializer support

ihsandemir:
Implemented the JSON serializer support.

### DIFF
--- a/hazelcast/include/hazelcast/client/serialization/pimpl/SerializationConstants.h
+++ b/hazelcast/include/hazelcast/client/serialization/pimpl/SerializationConstants.h
@@ -62,6 +62,13 @@ namespace hazelcast {
                         CONSTANT_TYPE_FLOAT_ARRAY = -18,
                         CONSTANT_TYPE_DOUBLE_ARRAY = -19,
                         CONSTANT_TYPE_STRING_ARRAY = -20,
+
+                        // ------------------------------------------------------------
+                        // LANGUAGE SPECIFIC SERIALIZERS
+                        // USED BY CLIENTS (Not deserialized by server)
+                                JAVASCRIPT_JSON_SERIALIZATION_TYPE = -130,
+                        // ------------------------------------------------------------
+
                         CONSTANT_TYPE_VOID_POINTER = INT32_MIN
                     };
                     // ------------------------------------------------------------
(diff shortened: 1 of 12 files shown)


hazelcast/include/hazelcast/client/serialization/pimpl/SerializationConstants.h
```diff
@@ -62,6 +62,13 @@ namespace hazelcast {
                         CONSTANT_TYPE_FLOAT_ARRAY = -18,
                         CONSTANT_TYPE_DOUBLE_ARRAY = -19,
                         CONSTANT_TYPE_STRING_ARRAY = -20,
+
+                        // ------------------------------------------------------------
+                        // LANGUAGE SPECIFIC SERIALIZERS
+                        // USED BY CLIENTS (Not deserialized by server)
```
sancar: Why do we have this comment? It seems wrong. Json is deserialized by the server.

ihsandemir: I copied the comment from here Javadoc https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/SerializationConstants.java#L94
I believe may be they wanted to say that it is not deserialized if no query, etc. But I agree it is better to remove it. I will remove it for C++ but may be Java side needs to be corrected as well, right?

sancar: Yes, java side also needs to be updated.

hazelcast/test/src/map/ClientMapTest.cpp
```diff
@@ -3370,6 +3370,31 @@ namespace hazelcast {
...
+                IMap<string, serialization::json::HazelcastJsonValue> map = ClientMapTest<TypeParam>::client->template getMap<std::string, serialization::json::HazelcastJsonValue>(
+                        ClientMapTest<TypeParam>::getTestName());
+                serialization::json::HazelcastJsonValue young("{ \"age\": 4 }");
+                serialization::json::HazelcastJsonValue old("{ \"age\": 20 }");
+                map.put("item1", young);
+                map.put("item1", old);
+
+                // Get the objects whose age is less than 6
+                std::vector<serialization::json::HazelcastJsonValue> result = map.values(
+                        query::GreaterLessPredicate<int>("a", 6, false, true));
```
sancar: Does this test pass ? Json does not have a "a" field . If this test is passing we need to check why ?

ihsandemir: I think there is no bug but i mistakenly inserted the second value with the same key as item1 and it overwrote the other value. And it was due to this mistake the test passed :) Corrected it now. Also added a check to verify that map has 2 entries.

PR #538, opened 2019-03-05
Update README for json serialization feature

ihsandemir:
Added README for json serialization feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -896,9 +896,9 @@ From now on, Hazelcast will use `MusicianSerializer` to serialize `Musician` obj
 
 ## 4.4. JSON Serialization
 
-You can use JSON formatted strings as objects in Hazelcast cluster. Starting with Hazelcast server version 3.12, the JSON serialization is one of the formerly supported serialization methods. Creating JSON objects in the cluster does not require any server side coding and hence you can just send a JSOn formatted string object to the cluster and query these objects by fields.
+You can use the JSON formatted strings as objects in Hazelcast cluster. Starting with Hazelcast IMDG 3.12, the JSON serialization is one of the formerly supported serialization methods. Creating JSON objects in the cluster does not require any server side coding and hence you can just send a JSON formatted string object to the cluster and query these objects by fields.
 
-In order to use JSON serialization, you should use `HazelcastJsonValue` object for the key or value. Here is an example IMap usage:
+In order to use JSON serialization, you should use the `HazelcastJsonValue` object for the key or value. Here is an example IMap usage:
 
 ```C++
     hazelcast::client::ClientConfig config;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -891,7 +894,43 @@ Now the last required step is to register the `MusicianSerializer` to the config
 
 From now on, Hazelcast will use `MusicianSerializer` to serialize `Musician` objects.
 
-## 4.4. Global Serialization
+## 4.4. JSON Serialization
+
+You can use JSON formatted strings as objects in Hazelcast cluster. Starting with Hazelcast server version 3.12, the JSON serialization is one of the formerly supported serialization methods. Creating JSON objects in the cluster does not require any server side coding and hence you can just send a JSOn formatted string object to the cluster and query these objects by fields.
```
Serdaro: ...use the JSON formatted strings as objects in the Hazelcast cluster. ...

... Starting with Hazelcast IMDG 3.12, the JSON serialization is one of the supported serialization methods ...

... you can just send a JSON ...

README.md
```diff
@@ -891,7 +894,43 @@ Now the last required step is to register the `MusicianSerializer` to the config
...
+
+In order to use JSON serialization, you should use `HazelcastJsonValue` object for the key or value. Here is an example IMap usage:
```
Serdaro: ... should use the `HazelcastJson......

README.md
```diff
@@ -891,7 +894,43 @@ Now the last required step is to register the `MusicianSerializer` to the config
...
+
+```C++
+    hazelcast::client::ClientConfig config;
+    hazelcast::client::HazelcastClient hz(config);
+
+    hazelcast::client::IMap<std::string, serialization::json::HazelcastJsonValue> map = hz.getMap<std::string, serialization::json::HazelcastJsonValue>(
+            "map");
+```
+
+We constructed a map in the cluster which has the `string` as the key and `HazelcastJsonValue` as the value. `HazelcastJsonValue` is a simple wrapper and identifier for the JSON formatted strings. You can get the JSON string from the `HazelcastJsonValue` object using the `toJsonString()` nethod.
```
Serdaro: ... which has `string` as the key ...

... using the `toJsonString()` method.

README.md
```diff
@@ -891,7 +894,43 @@ Now the last required step is to register the `MusicianSerializer` to the config
...
+    hazelcast::client::ClientConfig config;
+    hazelcast::client::HazelcastClient hz(config);
+
+    hazelcast::client::IMap<std::string, serialization::json::HazelcastJsonValue> map = hz.getMap<std::string, serialization::json::HazelcastJsonValue>(
+            "map");
+```
+
+We constructed a map in the cluster which has the `string` as the key and `HazelcastJsonValue` as the value. `HazelcastJsonValue` is a simple wrapper and identifier for the JSON formatted strings. You can get the JSON string from the `HazelcastJsonValue` object using the `toJsonString()` nethod. 
+
+You can construct the `HazelcastJsonValue` using one of the constructors. All constructors accept the JSON formatted string as the parameter. No JSON parsing is performed but it is your responsibility to provide correctly formatted JSON strings. The client will not validate the string, and it will send it to the cluster as it is. If you submit incorrect formatted JSON strings and later if you query those objects it is highly possible that you will get formatting errors since the server will fail to deserialize or find the query fields.
```
Serdaro: ... If you submit incorrectly formatted JSON strings and, later, if you query those objects, it is highly ...

README.md
```diff
@@ -891,7 +894,43 @@ Now the last required step is to register the `MusicianSerializer` to the config
...
+
+Here is an example of how you can construct a HazelcastJsonValue and put to the map:
```
Serdaro: ... construct a `HazelcastJsonValue` and ...

README.md
```diff
@@ -891,7 +894,43 @@ Now the last required step is to register the `MusicianSerializer` to the config
...
+
+```C++
+    map.put("item1", serialization::json::HazelcastJsonValue("{ \"age\": 4 }"));
+    map.put("item2", serialization::json::HazelcastJsonValue("{ \"age\": 20 }"));
+```
+
+You can query JSON objects in the cluster using the `Predicate`s of your choice. An example JSON query for querying values whose age is less than 6 is shown below:
```
Serdaro: ... querying the values whose ...

README.md
```diff
@@ -2671,7 +2710,69 @@ You can use `query::QueryConstants::getValueAttributeName()` (`this`) attribute
 
 In this example, the code creates a list with the values greater than or equal to "27".
 
-#### 7.7.1.4. Filtering with Paging Predicates
+#### 7.7.1.4. Querying with JSON Strings
+
+You can query JSON strings stored inside your Hazelcast clusters. To query a JSON string,
```
Serdaro: ... the JSON string ...

README.md
```diff
@@ -2671,7 +2710,69 @@ You can use `query::QueryConstants::getValueAttributeName()` (`this`) attribute
...
+you first need to create a `HazelcastJsonValue` from the JSON string using the `HazelcastJsonValue(const std::string &)`
+constructor(or one of the other constructors). You can use ``HazelcastJsonValue``s both as keys and values in the distributed data structures. Then, it is
```
Serdaro: ... constructor (or one of the ....

README.md
```diff
@@ -2671,7 +2710,69 @@ You can use `query::QueryConstants::getValueAttributeName()` (`this`) attribute
...
+    std::vector<serialization::json::HazelcastJsonValue> peopleUnder21 = idPersonMap.values(query::GreaterLessPredicate<int>("age", 21, false, true));
+```
+
+When running the queries, Hazelcast treats values extracted from the JSON documents as Java types so they
+can be compared with the query attribute. JSON specification defines five primitive types to be used in the JSON
+documents: `number`,`string`, `true`, `false` and `null`. The `string`, `true/false` and `null` types are treated
+as `String`, `boolean` and `null`, respectively. We treat the extracted `number` values as ``long``s if they
+can be represented by a `long`. Otherwise, ``number``s are treated as ``double``s.
+
+It is possible to query nested attributes and arrays in JSON documents. The query syntax is the same
```
Serdaro: ... in the JSON documents.

README.md
```diff
@@ -2671,7 +2710,69 @@ You can use `query::QueryConstants::getValueAttributeName()` (`this`) attribute
...
+ *
+ *
+ * The following query finds all the departments that have a person named "Peter" working in them.
+ */
+std::vector<serialization::json::HazelcastJsonValue> departmentWithPeter = departments.values(query::EqualPredicate<std::string>("people[any].name", "Peter"));
+```
+
+`HazelcastJsonValue` is a lightweight wrapper around your JSON strings. It is used merely as a way to indicate
+that the contained string should be treated as a valid JSON value. Hazelcast does not check the validity of JSON
+strings put into to maps. Putting an invalid JSON string in a map is permissible. However, in that case
```
Serdaro: ... strings put into the maps.

... invalid JSON string into a map ...

Issue #539, opened 2019-03-05
Serialization exception for extended ASCII character


ihsandemir:
It was reported at https://groups.google.com/forum/#!msg/hazelcast/xIwOza8MqGM/0kx7PJL8CAAJ that the extended ASCII encoded string key caused the following exception:

```
"IOException {Not enough bytes in internal buffer. Available:6 bytes but needed 7 bytes Error code:24} at DataInput::checkBoundary"
```
Java expects the extended ASCII character as two bytes but we are writing it as provided by the std::string buffer assuming that it was encoded as expected by Java (for better performance we were just bulk copying the characters) but this was not correct, the extended ASCII should be written as two bytes instead of single byte (which is actually enough to represent the extended ASCII character). This difference caused the mentioned problem.

The reproducer test:
```
                std::string utfStr = "Num\351ro";

                SerializationConfig serializationConfig;
                serialization::pimpl::SerializationService serializationService(serializationConfig);

                serialization::pimpl::Data data = serializationService.toData<std::string>(&utfStr);
                std::auto_ptr<std::string> deserializedString = serializationService.toObject<std::string>(data);
                ASSERT_EQ_PTR(utfStr, deserializedString.get(), std::string);
```

ihsandemir:
I examined the case and found out that the problem was the string encoding. Hazelcast strings work with utf-8 encoding and hence, it is expected that the provided std::string is utf-8 encoded. In the issue example the encoding was not done correctly in utf-8 as expected. The UTF encoding of `\351` (233) is `\xc3\xa9`. When we use this utf-8 encoding the client works properly. 

It is the user's responsibility to provide correctly encoded as `utf-8` in `std::string`. The std::string does not carry any encoding information, it is just a byte buffer. 

If you use C++11, you can easily provide string literals in utf-8 using `u8` keyword (https://stackoverflow.com/questions/13444930/is-the-u8-string-literal-necessary-in-c11).

PR #541, opened 2019-03-12
Add UTF-8 format validation for serialized and deserialized strings

ihsandemir:
Added UTF-8 format validation for serialized and deserialized strings We now provide exception to the user if the user supplies invalid utf-8 string.

related to https://github.com/hazelcast/hazelcast-cpp-client/issues/539

### DIFF
--- a/hazelcast/src/hazelcast/client/serialization/pimpl/DataInput.cpp
+++ b/hazelcast/src/hazelcast/client/serialization/pimpl/DataInput.cpp
@@ -152,14 +152,12 @@ namespace hazelcast {
                     } else {
                         utfBuffer.reserve((size_t) MAX_UTF_CHAR_SIZE * charCount);
                         byte b;
-                        size_t index = 0;
                         for (int i = 0; i < charCount; ++i) {
                             b = readByte();
-                            util::UTFUtil::readUTF8Char(*this, b, utfBuffer, index);
+                            util::UTFUtil::readUTF8Char(*this, b, utfBuffer);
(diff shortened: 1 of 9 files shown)


ihsandemir:
verify

ihsandemir:
verify-windows

ihsandemir:
verify

hazelcast/src/hazelcast/client/serialization/pimpl/DataInput.cpp
```diff
@@ -143,22 +146,18 @@ namespace hazelcast {
...
-                        for (int i = 0; i < len ; ++i) {
-                            checkAvailable(1);
-                            int numBytesForChar = getNumBytesForUtf8Char(&buffer[pos] + numBytesToRead);
-                            numBytesToRead += numBytesForChar;
-                            checkAvailable(numBytesToRead);
+                        utfBuffer.reserve((size_t) MAX_UTF_CHAR_SIZE * charCount);
+                        byte b;
+                        for (int i = 0; i < charCount; ++i) {
+                            b = readByte();
+                            util::UTFUtil::readUTF8Char(*this, b, utfBuffer);
```
sancar: Is there a specific reason why we diverged from java implementation here? 
The `readUTF8Char` seems more simple there.
And why did we create `ByteReadable` interface? If it is just for testing I think we need to complicate tests not code itself if necessary.

ihsandemir:
verify

PR #543, opened 2019-03-18
Update Json according to latest design decisions: change `toJsonString` to `toJson` and change namespace of `HazelcastJsonValue` class to hazelcast::client

ihsandemir:
Json update for latest design decision. Changed method `toJsonString` to `toString`. Also changed namespace of `HazelcastJsonValue` class to hazelcast::client.


hazelcast/include/hazelcast/client/HazelcastJsonValue.h
```diff
@@ -0,0 +1,88 @@
...
+         * In terms of querying, numbers in Json strings are treated as either
+         * {@code Long} or {@code Double}. Strings, booleans and null are treated as
+         * their C++ counterparts.
+         *
+         * HazelcastJsonValue keeps given string as it is.
+         *
+         */
+        class HAZELCAST_API HazelcastJsonValue {
+        public:
+            HazelcastJsonValue();
```
sancar: I think we should not need an empty constructor. Right?

hazelcast/src/hazelcast/client/serialization/json/HazelcastJsonValue.cpp
```diff
@@ -13,50 +13,37 @@
...
-                        return true;
-                    }
+        const std::string &HazelcastJsonValue::toString() const {
+            return jsonString;
+        }
 
-                    if (!jsonString) {
-                        return false;
-                    }
+        bool HazelcastJsonValue::operator==(const HazelcastJsonValue &rhs) const {
```
stalek71: Do you plan to add additional comparison operators here also or it's based on implicit conversion to std::string only?
I mean if any friend two arg comparisons should be added here also?

ihsandemir: No, I don't. Can you give an example of what is in your mind? Are you asking for "<", ">" kind of operators?

stalek71: I was thinking rather about the usage of std::string or const char* as right/left operator argument, but most of the compilers should be able to deduct proper implicit conversion to HazelcastJsonValue I think. Usually the best option is to provide < (lessThen) operator only, because it works properly in every kind of the sorting for instance and other operators can be defined using this single operator only in other operator definitions. We have here direct access to std::string (json string) so user can just "expose" it in case of sorting (or similar) operations.

ihsandemir: yes, `const char *` is automatically converted to `const std::string` and we use it this way throughout our API. We prefer `std::string` to `const char *` in general.

Regarding the less operator, i did not put it for now. Even the equal operator is extra. I do not want to add too much. We can add later if needed.

Issue #563, opened 2019-09-09
Unresolved getMap getQueue


yigitmultibankfx:
Using "Visual Studio 2019"
When I try to use getMap or getQueue I get linking error although I show the correct .lib version. Do I have to reference the dll file too? HazelCastClient object referencing works well I get linking error only when I call getMap getQueue etc.

Here is the linking error I receive:

LNK2019	unresolved external symbol "__declspec(dllimport) public: __thiscall hazelcast::client::serialization::pimpl::Data::Data(class hazelcast::client::serialization::pimpl::Data &&)" (__imp_??0Data@pimpl@serialization@client@hazelcast@@QAE@$$QAV01234@@Z) referenced in function "public: static void __cdecl std::_Default_allocator_traits<class std::allocator<class hazelcast::client::serialization::pimpl::Data> >::construct<class hazelcast::client::serialization::pimpl::Data,class hazelcast::client::serialization::pimpl::Data>(class std::allocator<class hazelcast::client::serialization::pimpl::Data> &,class hazelcast::client::serialization::pimpl::Data * const,class hazelcast::client::serialization::pimpl::Data &&)" (??$construct@VData@pimpl@serialization@client@hazelcast@@V12345@@?$_Default_allocator_traits@V?$allocator@VData@pimpl@serialization@client@hazelcast@@@std@@@std@@SAXAAV?$allocator@VData@pimpl@serialization@client@hazelcast@@@1@QAVData@pimpl@serialization@client@hazelcast@@$$QAV34567@@Z)	ManagerToDistributedMap	D:\Projects\C++\ManagerToDistributedMap\ManagerToDistributedMap\ManagerToDistributedMap\ManagerToDistributedMap.obj	1

ihsandemir:
Hello @yigitmultibankfx There may be a few thing to double check:

1. Are you using 64-bit or 32-bit application and did you point to the correct type library?
2. Did you use the compile flag as describe here?: https://github.com/hazelcast/hazelcast-cpp-client#1313-windows-client

I suspect it is mostly due to item 2.


1.

yigitmultibankfx:
Hi,

1- I am using the 32-bit application

2- I defined the compile flag as HAZELCAST_USE_SHARED. If I do not define that flag there are much more linking errors so I know that I actually set that. 

Here is how I can simply generate the error:

1- I create a simple console app with Visual Studio 2019. The active configuration is default as "Debug x86"

[...]

PR #564, opened 2019-09-12
Missing Data class move constructor symbol problem for C++ language level C++11 and later problem

ihsandemir:
Added the destructor to the Data class to prevent the generation of the move constructor, when the C++11 and later compiler is used by the users. Adding the destructor avoid implicit move constructor generation (see https://en.cppreference.com/w/cpp/language/move_constructor).

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/563


ihsandemir:
Since we already started using C++11, this issue is no longer valid. Closing it.

Issue #569, opened 2019-12-02
Pipeline functionality causes an assertion failure in Debug build and a crash in Release build


asif-ma:
C++ compiler version: Visual Studio 2013 and Visual Studio 2017
Hazelcast Cpp client version: 3.12
Hazelcast server version: 3.12
Number of the clients: 1
Cluster size, i.e. the number of Hazelcast cluster members: 1
OS version (Windows/Linux/OSX): Windows

Please attach relevant logs and files for client and server side.

#### Expected behaviour
In Debug mode there should be no assertions on the client side.
In Release mode there should be no exceptions on the server side and the client should not crash.

#### Actual behaviour
============================================
Debug Mode: Stack Trace
==============================================

Assertion failed: checkReadAvailable(UINT8_SIZE), file C:\Build\workspace\TPARTY_win\tparty\Temp\hazelcast\hazelcast-cpp-client-3.12\hazelcast\include\hazelcast/util/LittleEndianBufferWrapper.h, line 74

>	HazelcastClient3.12_64d.dll!hazelcast::util::LittleEndianBufferWrapper::getUint8() Line 74	C++
 	HazelcastClient3.12_64d.dll!hazelcast::util::LittleEndianBufferWrapper::getBoolean() Line 84	C++
 	HazelcastClient3.12_64d.dll!hazelcast::client::protocol::ClientMessage::getNullable<hazelcast::client::serialization::pimpl::Data>() Line 254	C++
 	HazelcastClient3.12_64d.dll!hazelcast::client::protocol::codec::MapPutCodec::ResponseParameters::ResponseParameters(hazelcast::client::protocol::ClientMessage & clientMessage) Line 68	C++
 	HazelcastClient3.12_64d.dll!hazelcast::client::protocol::codec::MapPutCodec::ResponseParameters::decode(hazelcast::client::protocol::ClientMessage & clientMessage) Line 73	C++
 	HazelcastClientVC13.exe!hazelcast::client::impl::DataMessageDecoder<hazelcast::client::protocol::codec::MapPutCodec,Country>::decodeClientMessage(const boost::shared_ptr<hazelcast::client::protocol::ClientMessage> & clientMessage, hazelcast::client::serialization::pimpl::SerializationService & serializationService) Line 52	C++
[...]

PR #570, opened 2019-12-23
Fix the decoder usage in IMap::getAsync call

ihsandemir:
Corrected the decoder usage on IMap::getAsync call. There are also some minor refactoring.

related to https://github.com/hazelcast/hazelcast-cpp-client/issues/569


sancar:
Isn't getAsync and putAsync decoder doing the same thing ? Both returns a data. How is this fixing the issue ?

ihsandemir:
> Isn't getAsync and putAsync decoder doing the same thing ? Both returns a data. How is this fixing the issue ?

I changed "fixing" to "related" since I can not relate why this can fix the issue but it corrects the codec usage even though they both do the same thing.

ihsandemir:
verify-windows

PR #571, opened 2020-01-13
Fix concurrent access to message decoding

ihsandemir:
Fix for resolving the race when there is concurrent access to client message decoding. A copy message is constructed with a fresh read `index` on the same message buffer and decodes using this new message. This was an existing `TODO` fix which already existed for the Java client.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/569

### DIFF
--- a/hazelcast/include/hazelcast/client/internal/ClientDelegatingFuture.h
+++ b/hazelcast/include/hazelcast/client/internal/ClientDelegatingFuture.h
@@ -153,15 +153,16 @@ namespace hazelcast {
                     if (decodedResponse != boost::static_pointer_cast<V>(VOIDOBJ)) {
                         return decodedResponse;
                     }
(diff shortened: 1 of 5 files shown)


asif-ma:
@ihsandemir , I have tested the latest changes from "fixCandidate" branch and can confirm it works fine.

asif-ma:
@ihsandemir will this fix be part of version 4 release or is there going to be a 3.13 release?

ihsandemir:
@asif-ma I requested product development to see if we can do a release specific with this fix only prior to 4.0 client release which may not be very soon. I will update you on this. Thanks again for your contributions on spotting this issue.

asif-ma:
Hi @ihsandemir, did you receive any update from product development regarding a potential release prior to 4.0?

PR #574, opened 2020-02-21
4.0 architecture related changes

ihsandemir:
The following changes are done:

* Remove auto_ptr (replace by std::unique_ptr)
* Remove boost::shared_ptr (replaced by std::shared_ptr)
* Replace boost::chrono with std::chrono.
* Replace util::Mutex with std::mutex partially.
* Changed the asio and googletest dependencies to latest released versions.
* Internal dependency to boost::property_tree could not be removed due to the usage os xml and json parsers.
* Added docker file for fedora-latest builds.
* Changed ssl tests protocol from tlsv1 to tlsv1.2 since the server does not except tlsv1 by default.
* Changes per new version of easylogging library. The file configuration parsing ignores incorrect parameters in the config but the configuration is loaded anyway.
* Fixed shutdown task and `ClientInvocationFuture` weak_ptr usage.

Using vcpkg on Windows for thrift libraries but using openssl as from C:\OpenSSL-Win32(64) even for thrift openssl dependency (thrift vcpkg cmake config file incorrectly tries to use from under thrift vcpkg, had to modify the thrift find cmake script).


ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-linux

ihsandemir:
verify-linux

ihsandemir:
verify-windows

sancar:
What is the reason of changing Mutex partially ? Do we postpone the change to another pr , or is there a technical reason that we can not replace ?

sancar:
About util::Atomic, I see that it is renamed to util::Sync . Where the name Sync is coming from ? 
It is called std::atomic on cpp standard library also. 
I also see that atomic<shared_ptr> only available on C++20. I assume that is the reason why we don't replace it with standard atomic. 

I think we also need to remove AtomicBoolean and AtomicInt classes eventually.

ihsandemir:
> What is the reason of changing Mutex partially ? Do we postpone the change to another pr , or is there a technical reason that we can not replace ?

yes, they all will be replace in following PRs. I just wanted a baseline PR to be able to merge with minimum changes for the new standard before changing the rest.

ihsandemir:
> About util::Atomic, I see that it is renamed to util::Sync . Where the name Sync is coming from ?
> It is called std::atomic on cpp standard library also.
> I also see that atomic<shared_ptr> only available on C++20. I assume that is the reason why we don't replace it with standard atomic.
> 
> I think we also need to remove AtomicBoolean and AtomicInt classes eventually.

Yes, unfortunately atomic<shared_ptr> is only at C++20. But I am planning to use optional as much as possible which may eliminate such needs as possible. There may be a number of places we can not eliminate such as passing objects to the threads or executors. Will see when I change them.

ihsandemir:
verify

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

PR #575, opened 2020-03-05
Merge cpp files for shorter build times

ihsandemir:
Refactored the library cpp codes into a single file decreasing the compilation time significantly.
    
    Also, refactored the tests into a single file (except the remote controller) which also decreases the compilation time.

    Refactored the protocol codecs to reside in single compilation unit (.cpp file) for faster compilation times.

Build times after fix:

Platform| Time(seconds) For Only Lib| Time(seconds) For All (Lib+test exe + examples)
------------------|-----------------|------------------
Windows|38|286.6
Linux|45|318
MAC OS|31|88

Build times before the fix(master branch):

Platform| Time(seconds) For Only Lib| Time(seconds) For All (Lib+test exe + examples)
------------------|-----------------|------------------
Windows|741|1333
Linux||
MAC OS||

This PR decreases the compile time significantly but increased the file size. The IDE environment requires some heap memory arrangements for it to work better with IDE. We can also decrease the file size a lot more by doing a number of cleanups such as merging the namespaces, etc.
For my settings, I am using the following memory just to be in safe place and avoid too many garbage collection pauses of JVM used by IDE: `-Xms6G
-Xmx12G -XX:ReservedCodeCacheSize=1G`


ihsandemir:
verify-linux

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

PR #576, opened 2020-03-10
Remove all submodule dependencies

ihsandemir:
Removed the header-only asio and switched to boost::asio since the boost::asio no longer requires boost::system library compilation and linkage but it can now be used as header-only (See https://www.boost.org/doc/libs/1_66_0/libs/system/doc/reference.html https://stackoverflow.com/questions/31110409/boost-error-code-header-only-does-not-have-the-advertised-effect/31131980#31131980).

Also started to use find_package for boost (This way it will be more flexible and compatible if the user wants to compile and use with their own version of boost). 

Removed all submodules.
    
Boost is being searched first in user's cmake environment, if it is not found, it is being downloaded and unzipped and the unzipped version is being used.
    
Googletest is also being searched first and if it is not installed, then it is added as a cmake external project so that it is also being downloaded buring configuration of cmake.


ihsandemir:
verify

PR #578, opened 2020-03-18
Merge TcpSocket and SSLSocket to utilize async IO (boost::asio)

ihsandemir:
Removed custom TcpSocket implementation and replaced it by using boost::asio tcp socket. Also removed io threads and using a single io thread which runs the boost::asio::io_context::run until shutdown. The number of threads can be changed to see performance affects.

Fixes based on test results. A bug is corrected in Connection.h reAuthentication.
    
Transaction test is corrected.


hazelcast/include/hazelcast/client/Socket.h
```diff
@@ -16,67 +16,44 @@
...
-            /**
-             * @param buffer
-             * @param len length of the buffer
-             * @param flag bsd sockets options flag. Only MSG_WAITALL is supported when SSL is enabled.
-             * @return number of bytes send
-             * @throw IOException in failure.
-             */
-            virtual int send(const void *buffer, int len, int flag = 0) = 0;
+            virtual void
+            asyncStart(const std::shared_ptr<connection::Connection> &connection,
```
sancar: It seems like the levels of concerns are mixed. A socket should not now about connection or authentication. I think this is not a socket anymore. 

I only see one implementation of this which is named BaseSocket, which also seems confusing. It seems that there is no need for this class to be pure virtual anymore.

ihsandemir: Discussed the reasons for this. Will leave it this way for now. In Java client master code these also changed, hence will change this again in the future.

PR #584, opened 2020-04-13
Architectural changes to use boost thread_pool, boost::future, std::promise, std::nested_exception, connection local invocation and event handler maps

ihsandemir:
Changed the event processing executor usage to use asio::thread_pool.

Changed the ResponseThread to ResponseProcessor and it now processes the response message directly by default and will use a thread_pool if the newly introduced parameter `hazelcast.client.response.executor.thread.count` is set to a positive value.

Removed custom Thread, Future implementations and used std::thread, boost::asio::thread_pool and boost::future/boost::promise.

boost::future supports continuation (`.then` support).

Changed the invocation and event handler map design. Using an `unordered_map` which which only being accessed from the same `io` thread for the connection. This eliminated the need for using sync maps for invocations and events. If any other thread needs to access the invocations, it passes this request to io thread. This is only needed when closing the socket for notifying the outstanding invocations. If an invocation can not be queued into the connection (i.e. no connection exists for the invocation and it is being retried), then it is registered in a sync map at the invocation service only for the first retry and cleared from the service when a response is finally received. This ensures that the user thread is always being notified.

Exception architecture is compeletely changed so that we use std exception_ptr capabilities. Also integrated std::nested exceptions for exceptions which has a cause exception. The cause is obtained by using `std::rethrow_if_nested` method.

The project now depends on boost::thread for boost::future. Changed cmake so that the boost installation is user's responsibility. Also using the boost Thread library version latest which is version 5.


hazelcast/include/hazelcast/client/IAtomicLong.h
```diff
@@ -190,7 +190,7 @@ namespace hazelcast {
              * @return an {@link ICompletableFuture} with the current value
              * @since cluster version 3.7
              */
-            std::shared_ptr<ICompletableFuture<int64_t> > getAsync();
+            future<std::shared_ptr<int64_t>> getAsync();
```
sancar, line 193: Should not these be future<int64_t> ? Do we need shared ptr's for primitives ?

ihsandemir: No, we don't but I wıll change it when I change SerializationService to not use shared_ptr, this pointer is due to current serializationservice `toObject` method.

hazelcast/include/hazelcast/client/IMap.h
```diff
@@ -1034,7 +1033,7 @@ namespace hazelcast {
              * @see ICompletableFuture
              * @see #setAsync(Object, Object)
              */
-            std::shared_ptr<ICompletableFuture<V> > putAsync(const K &key, const V &value) {
+            future<std::shared_ptr<V>> putAsync(const K &key, const V &value) {
```
sancar, line 1036: Would it make sense to use shared_future here ?

ihsandemir: I try not to use shared_future where future is enough. the user can convert it to shared_future using future::share() method anyway.

PR #586, opened 2020-04-18
Add loaded event type

### LINKED ISSUE #551: Missing "LOADED" event
See https://github.com/hazelcast/hazelcast-go-client/issues/464

IMDG 3.11 added LOADED event - https://docs.hazelcast.org/docs/3.12.1/javadoc/com/hazelcast/core/EntryEventType.html#LOADED

This is not handled by this client

enozcan:
Closes #551 

- `Loaded` event is added to EntryListener.
- ~~Type Id's are [updated.](https://github.com/hazelcast/hazelcast/blob/9c7c62f0eec24219c60fd8f92a9e8c52695fba6e/hazelcast/src/main/java/com/hazelcast/core/EntryEventType.java#L124)~~


hazelcast/include/hazelcast/client/EntryEvent.h
```diff
@@ -44,12 +44,13 @@ namespace hazelcast {
                 ADDED =  1,
                 REMOVED = 1 << 1,
                 UPDATED = 1 << 2,
-                EVICTED = 1 << 3 ,
-                EVICT_ALL = 1 << 4 ,
-                CLEAR_ALL = 1 << 5 ,
-                MERGED = 1 << 6 ,
-                EXPIRED = 1 << 7,
+                EVICTED = 1 << 3,
+                EXPIRED = 1 << 4,
```
ihsandemir, line 48: These numbers seems to be changed when going from Java server 3.12.x to 4.0.x. Your change is based on 4.0.x but client did not switch to work with that server version yet (protocol changes will be done in later PRs), hence the tests (for e.g. for expired) normally would not pass (we may be missing the tests)

enozcan: Reverted back to 3.12.x compatible types.

hazelcast/test/src/HazelcastTests4.cpp
```diff
@@ -3229,6 +3253,17 @@ namespace hazelcast {
...
                 }
 
+                TEST_F(RawPointerMapTest, testLoadEvent) {
+                    hazelcast::util::CountDownLatch latch(1);
+                    LoadListener loadListener(latch);
+                    std::string listenerId = storeEnabledMap->addEntryListener(loadListener, false);
+                    storeEnabledMap->put("key1", "value1");
+                    storeEnabledMap->evictAll();
+                    storeEnabledMap->get("key1");
+                    ASSERT_TRUE(latch.await(120));
```
ihsandemir, line 3263: Use `ASSERT_OPEN_EVENTUALLY` here

enozcan:
I did not have enough time to take another look for this issue. It seems like the handler is missing somewhere else hence the event is not handled properly. Since the codebase is now way more different than it was when the PR is opened, I'm closing this.

PR #593, opened 2020-05-08
Changed closedTime to closedTimeDuration to support correct atomic initialization

### LINKED ISSUE #592: Problem with compiling current master branch with gcc 9.3
C++ compiler version: gcc 9.3
Hazelcast Cpp client version: master branch
Hazelcast server version: N/A
Number of the clients: N/A
Cluster size, i.e. the number of Hazelcast cluster members: N/A
OS version (Windows/Linux/OSX): Linux Ubuntu 20.04

#### Expected behaviour
Hazelcast-cpp-client successfully compiles with gcc 9.3

#### Actual behaviour
```
/home/rikevoltz/work/junk/hazelcast-cpp-client/hazelcast/src/hazelcast/client/network.cpp: In constructor ‘hazelcast::client::connection::Connection::Connection(const hazelcast::client::Address&, hazelcast::client::spi::ClientContext&, int, const std::shared_ptr<hazelcast::client::connection::AuthenticationFuture>&, hazelcast::client::internal::socket::SocketFactory&, bool, hazelcast::client::connection::ClientConnectionManagerImpl&, std::chrono::_V2::steady_clock::duration&, boost::asio::ip::tcp::resolver&)’:
/home/rikevoltz/work/junk/hazelcast-cpp-client/hazelcast/src/hazelcast/client/network.cpp:911:77: error: use of deleted function ‘constexpr std::atomic<_Tp>::atomic() [with _Tp = std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > >]’
  911 |                       connectionManager(clientConnectionManager), alive(true) {
      |                                                                             ^
In file included from /home/rikevoltz/work/junk/hazelcast-cpp-client/hazelcast/include/hazelcast/util/AtomicBoolean.h:24,
                 from /home/rikevoltz/work/junk/hazelcast-cpp-client/hazelcast/include/hazelcast/client/connection/DefaultClientConnectionStrategy.h:21,
                 from /home/rikevoltz/work/junk/hazelcast-cpp-client/hazelcast/src/hazelcast/client/network.cpp:35:
/usr/include/c++/9/atomic:198:7: note: ‘constexpr std::atomic<_Tp>::atomic() noexcept [with _Tp = std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long int, std::ratio<1, 1000000000> > >]’ is implicitly deleted because its exception-specification does not match the implicit exception-specification ‘’
  198 |       atomic() noexcept = default;
      |       ^~~~~~
```
`cmake . && make` with gcc 9.3 as compiler

#### Additional info
Problem occured [there](https://github.com/hazelcast/hazelcast-cpp-client/blob/5fc77d0c6ec863c8b9864c6537c9f5d352ed2f85/hazelcast/src/hazelcast/client/network.cpp#L911) - there is  `std::atomic<std::chrono::steady_clock::time_point> closedTime` variable in Connection class, that doesn't initialized explicitly in the constructor of this class. It implies using default empty constructor that entails compile error (because `atomic` default empty constructor is noexcept and it obligates internal variable to have noexcept default empty constructor too) in gcc<10 (in 10.0 everything is ok) and clang<9.0 (in 9 everything is ok) (probably in new versions of compilers they made time_point constructor noexcept).
Solution, proposed [there](https://github.com/hazelcast/hazelcast-cpp-client/pull/591#issuecomment-625709801) may [take place](https://godbolt.org/z/JtGdeE) to fix compilation via gcc compilers, but for clang [it doesn't help](https://godbolt.org/z/mu5fo2).
But I can miss something in my small research.

RikeVoltz:
This PR fix #592, the idea of fix is taken from [this answer](https://stackoverflow.com/a/36316726/12480542)

### DIFF
--- a/hazelcast/include/hazelcast/client/connection/Connection.h
+++ b/hazelcast/include/hazelcast/client/connection/Connection.h
@@ -142,7 +142,7 @@ namespace hazelcast {
                 void innerClose();
 
                 std::chrono::steady_clock::time_point startTime;
-                std::atomic<std::chrono::steady_clock::time_point> closedTime;
+                std::atomic<std::chrono::steady_clock::duration> closedTimeDuration;
                 spi::ClientContext &clientContext;
                 protocol::IMessageHandler &invocationService;
                 std::unique_ptr<Socket> socket;
--- a/hazelcast/include/hazelcast/client/connection/ReadHandler.h
+++ b/hazelcast/include/hazelcast/client/connection/ReadHandler.h
@@ -60,7 +60,7 @@ namespace hazelcast {
             private:
 
                 protocol::ClientMessageBuilder builder;
-                std::atomic<std::chrono::steady_clock::time_point> lastReadTime;
+                std::atomic<std::chrono::steady_clock::duration> lastReadTimeDuration;
             };
         }
     }
--- a/hazelcast/src/hazelcast/client/network.cpp
+++ b/hazelcast/src/hazelcast/client/network.cpp
@@ -902,6 +902,7 @@ namespace hazelcast {
                                    boost::asio::ip::tcp::resolver &resolver)
                     : readHandler(*this, 16 << 10),
                       startTime(std::chrono::steady_clock::now()),
+                      closedTimeDuration(),
                       clientContext(clientContext),
                       invocationService(clientContext.getInvocationService()),
                       authFuture(authFuture),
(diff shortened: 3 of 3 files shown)


RikeVoltz:
Additional casts are proposed in this PR, is it suitable?

ihsandemir:
verify-linux

ihsandemir:
@RikeVoltz aws windows machine related test failures, other tests passed. it will be ok when windows build is up and running.

ihsandemir:
verify

ihsandemir:
verify-windows

PR #598, opened 2020-06-01
API Changes for C++11

ihsandemir:
API Changes for C++11:

* Changes all APIs to works as async by default. Removed the `...Async` apis.

* Changed the serialization so that we do not require serialization configuration and no need change to the user class for serialization. The serializer lookup is now performed during compilation time.

* All user APIs now return `boost::future` which supports continuations. See (https://www.boost.org/doc/libs/1_73_0/doc/html/thread/synchronization.html#thread.synchronization.futures.then)

* Removed the `RawPointer` and `Mixed` APIs. The same capability can now be achieved with the new API where user provides the template parameters (or otherwise it is template deducted) for each method call. This lets a different type for each method of a structure and if you need to do late deserialization then just use `TypedData` as the key and value types.

* The listeners are now temporary objects at the API which forces the user to provide a temporary object so that we can take ownership of the provided object. This makes the lifecycle management of the listeners more accurate so that we never call a listeners which was already destroyed. I intentionally kept it a temporary to explicitely avoid implicit copying of the user provided listeners (may happen if the user passes a `const &` object).

* Near cache now only works with data storage. The object storage will be re-established when the near cache is reimplemented (possibly using the `boost::any`)


ihsandemir:
verify-linux

PR #602, opened 2020-07-07
Use override, =default and =delete 

yemreinci:
Ran clang-tidy's checks `modernize-use-override`, `modernize-use-equals-delete` and `modernize-use-equals-default` on all of the code files excluding generated sources and external libraries. Also removed some "//@Override" annotation lines which were probably copy-pasted from the Java client. 

We may also consider to apply other modernization fixes ([listed here](https://clang.llvm.org/extra/clang-tidy/checks/list.html)), and maybe add a clang-tidy configuration for the project.

**Edit**: Here are the script I wrote to apply these fixes and the configuration file used by clang-tidy:

```bash
#!/usr/bin/env bash

SRC_FILES=$(find examples/ hazelcast/ | grep -v generated-sources | grep -v cpp-controller | grep -v soak-test | grep -E "\.cpp$" );

echo "Clang-tidy is going to work on these source files:";
for file in $SRC_FILES
do
    echo "  | $file";
done
echo;

HEADER_FILTER="($(pwd)/hazelcast/include/.*)|($(pwd)/hazelcast/test/.*)|($(pwd)/examples/.*)";

echo "Header filter:";
echo "  | $HEADER_FILTER";
echo;


echo "Effective checks:"
for check in $(clang-tidy --list-checks | tail -n +2)
do
    echo "  | $check"
done
echo;


clang-tidy -p $(pwd)/build-debug --header-filter=$HEADER_FILTER $SRC_FILES -fix
```

.clang-tidy:
```yaml
Checks:          'modernize-use-override,modernize-use-equals-delete,modernize-use-equals-default,-clang-diagnostic-*,-clang-analyzer-*'
```

Also I used -DCMAKE_EXPORT_COMPILE_COMMANDS=1 when running cmake to create a compile-commands.json file under the build path which then read by clang-tidy.

### DIFF
--- a/examples/Org.Website.Samples/ExecutorSample.cpp
+++ b/examples/Org.Website.Samples/ExecutorSample.cpp
@@ -91,12 +91,12 @@ namespace hazelcast {
 
 class PrinterCallback : public ExecutionCallback<std::string> {
 public:
-    virtual void onResponse(const boost::optional<std::string> &response) {
+    void onResponse(const boost::optional<std::string> &response) override {
         std::cout << "The execution of the task is completed successfully and server returned:" << *response
                   << std::endl;
     }
 
-    virtual void onFailure(std::exception_ptr e) {
+    void onFailure(std::exception_ptr e) override {
         try {
             std::rethrow_exception(e);
         } catch (hazelcast::client::exception::IException &e) {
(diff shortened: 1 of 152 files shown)


ihsandemir:
verify

PR #604, opened 2020-07-16
Change ItemListener and EntryListener

yemreinci:
Modified `ItemListener` and `EntryListener`. Instead of being interfaces with virtual event handlers left to be implemented by the user, they now hold std::function event handlers that are passed by the user using cascading setters. 
See examples for the new API.

### Benefits
* User may pass any type of callable for handlers that can be represented with [std::function](https://en.cppreference.com/w/cpp/utility/functional/function). (lambda functions, classes overloading operator(), bind expressions ...)
* No inheritance.
* No redundant typing for unimplemented handlers, the user doesn't have to write functions for the events that are not to their interest.


yemreinci:
verify

yemreinci:
verify

PR #607, opened 2020-07-28
Change MessageListener

yemreinci:
Made the same changes I did on `EntryListener` and `ItemListener` to `MessageListener`. (see #604)

### DIFF
--- a/docsConfig
+++ b/docsConfig
@@ -784,7 +784,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       = */util/* */iTest/* */test/* */collection/* */connection/* */executor/* */impl/* */map/* */multimap/* */protocol/* */proxy/* */queue/* */serialization/pimpl/* */spi/* */txn/* */exception/pimpl/*
+EXCLUDE_PATTERNS       = */util/* */iTest/* */test/* */collection/* */connection/* */executor/* */impl/* */map/* */multimap/* */protocol/* */proxy/* */queue/* */serialization/pimpl/* */spi/* */topic/* */txn/* */exception/pimpl/*
 
 # The EXCLUDE_SYMBOLS tag can be used to specify one or more symbol names
 # (namespaces, classes, functions, etc.) that should be excluded from the
--- a/docsConfig
+++ b/docsConfig
@@ -784,7 +784,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       = */util/* */iTest/* */test/* */collection/* */connection/* */executor/* */impl/* */map/* */multimap/* */protocol/* */proxy/* */queue/* */serialization/pimpl/* */spi/* */topic/* */txn/* */exception/pimpl/*
+EXCLUDE_PATTERNS       = */util/* */iTest/* */test/* */collection/* */connection/* */executor/* */impl/* */map/* */multimap/* */protocol/* */proxy/* */queue/* */serialization/pimpl/* */spi/* */txn/* */exception/pimpl/*
 
 # The EXCLUDE_SYMBOLS tag can be used to specify one or more symbol names
 # (namespaces, classes, functions, etc.) that should be excluded from the
--- a/examples/Org.Website.Samples/TopicSample.cpp
+++ b/examples/Org.Website.Samples/TopicSample.cpp
@@ -25,7 +25,7 @@ int main() {
     // Add a Listener to the Topic
     topic->addMessageListener(
         topic::MessageListener().
-            on_received([](const topic::Message &message) {
+            on_received([](topic::Message &&message) {
                 std::cout << "Got message " << message.getMessageObject().get<std::string>().value_or("null") << std::endl;
             })
     ).get();
--- a/examples/Org.Website.Samples/TopicSample.cpp
+++ b/examples/Org.Website.Samples/TopicSample.cpp
@@ -17,20 +17,18 @@
 
 using namespace hazelcast::client;
 
-class TopicSample {
-public:
-    void operator()(topic::Message &&message) {
-        std::cout << "Got message " << message.getMessageObject().get<std::string>().value_or("null") << std::endl;
-    }
-};
-
 int main() {
     // Start the Hazelcast Client and connect to an already running Hazelcast Cluster on 127.0.0.1
     HazelcastClient hz;
     // Get a Topic called "my-distributed-topic"
     auto topic = hz.getTopic("my-distributed-topic");
     // Add a Listener to the Topic
-    topic->addMessageListener(TopicSample{}).get();
+    topic->addMessageListener(
+        topic::MessageListener().
+            on_received([](const topic::Message &message) {
+                std::cout << "Got message " << message.getMessageObject().get<std::string>().value_or("null") << std::endl;
+            })
+    ).get();
     // Publish a message to the Topic
     topic->publish("Hello to distributed world").get();
     // Shutdown this Hazelcast Client
--- a/examples/Org.Website.Samples/TopicSample.cpp
+++ b/examples/Org.Website.Samples/TopicSample.cpp
@@ -24,7 +24,7 @@ int main() {
     auto topic = hz.getTopic("my-distributed-topic");
     // Add a Listener to the Topic
     topic->addMessageListener(
-        topic::MessageListener().
+        topic::Listener().
             on_received([](topic::Message &&message) {
                 std::cout << "Got message " << message.getMessageObject().get<std::string>().value_or("null") << std::endl;
             })
--- a/hazelcast/include/hazelcast/client/ITopic.h
+++ b/hazelcast/include/hazelcast/client/ITopic.h
@@ -21,7 +21,7 @@
 namespace hazelcast {
     namespace client {
         namespace topic {
-            class MessageListener;
+            class Listener;
         }
 
         /**
@@ -55,8 +55,8 @@ namespace hazelcast {
             /**
             * Subscribe to this topic. 
             * When a message is published on this topic, 
-            * the callback function that was provided to the given MessageListener object is called.
-            * \see MessageListener
+            * the callback function that was provided to the given Listener object is called.
+            * \see Listener
             *
             * Warning 1: If listener should do a time consuming operation, off-load the operation to another thread.
             * otherwise it will slow down the system.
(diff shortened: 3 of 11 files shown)


yemreinci:
verify

yemreinci:
verify

docsConfig
```diff
@@ -784,7 +784,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       = */util/* */iTest/* */test/* */collection/* */connection/* */executor/* */impl/* */map/* */multimap/* */protocol/* */proxy/* */queue/* */serialization/pimpl/* */spi/* */topic/* */txn/* */exception/pimpl/*
+EXCLUDE_PATTERNS       = */util/* */iTest/* */test/* */collection/* */connection/* */executor/* */impl/* */map/* */multimap/* */protocol/* */proxy/* */queue/* */serialization/pimpl/* */spi/* */txn/* */exception/pimpl/*
```
ihsandemir, line 787: i think this whole pattern needs to be updated. it looks out of date. we can do it in another PR. Can you open an issue to update this file?

examples/Org.Website.Samples/TopicSample.cpp
```diff
@@ -17,20 +17,18 @@
...
 int main() {
     // Start the Hazelcast Client and connect to an already running Hazelcast Cluster on 127.0.0.1
     HazelcastClient hz;
     // Get a Topic called "my-distributed-topic"
     auto topic = hz.getTopic("my-distributed-topic");
     // Add a Listener to the Topic
-    topic->addMessageListener(TopicSample{}).get();
+    topic->addMessageListener(
+        topic::MessageListener().
+            on_received([](const topic::Message &message) {
```
ihsandemir, line 28: previous API is passing as rvalue and I think we can keep it that way, since we will discard the object, better yet just pass it to the user so that he/she can move around without copy.

hazelcast/include/hazelcast/client/ITopic.h
```diff
@@ -50,37 +53,26 @@ namespace hazelcast {
...
+            * Subscribe to this topic. 
+            * When a message is published on this topic, 
+            * the callback function that was provided to the given MessageListener object is called.
+            * \see MessageListener
             *
-            *      class MyMessageListener {
-            *      public:
-            *          //....
-            *
-            *          void onMessage(topic::Message<std::string> message) {
```
ihsandemir, line 62: needs to be updated

hazelcast/include/hazelcast/client/topic/MessageListener.h
```diff
@@ -28,27 +29,50 @@
...
-             * Invoked when a message is received for the added topic. Note that topic guarantees message ordering.
-             * Therefore there is only one thread invoking onMessage. The user should not keep the thread busy, but preferably
-             * should dispatch it via an Executor. This will increase the performance of the topic.
-             *
-             * @param message the message that is received for the added topic
+             * Listen to messages from an ITopic
+             * \see ITopic::addMessageListener
              */
-            class HAZELCAST_API MessageListener {
+            class HAZELCAST_API MessageListener final {
```
ihsandemir, line 41: better naming suggestion: MessageListener -> listener

hazelcast/include/hazelcast/client/topic/MessageListener.h
```diff
@@ -28,27 +29,50 @@
...
-                *
-                * @param message the message that is received for the added topic
-                */
-                virtual void onMessage(Message &&message) = 0;
-            };
+                 * \copydoc MessageListener::on_received
+                 */
+                template<typename Handler,
+                         typename = util::enable_if_rvalue_ref_t<Handler &&>>
+                MessageListener &&on_received(Handler &&h) && {
```
ihsandemir, line 64: is this left-over? we already have this method above?

why are you returning rvalue here? I think just lvalue ref would be good enough. returning rvalue reference will also have its own problems https://stackoverflow.com/questions/1116641/is-returning-by-rvalue-reference-more-efficient

hazelcast/include/hazelcast/client/topic/MessageListener.h
```diff
@@ -28,27 +29,50 @@
...
+                    on_received(std::move(h));
+                    return std::move(*this);
+                }
 
+            private:
+                using HandlerType = std::function<void(const Message &)>;
```
ihsandemir, line 70: signature can take method parameter Message as rvalue.

hazelcast/include/hazelcast/client/topic/ReliableMessageListener.h
```diff
@@ -60,8 +60,19 @@ namespace hazelcast {
...
+                virtual ~ReliableMessageListener() = default;
+
+                /**
+                * Invoked when a message is received for the added topic. Note that topic guarantees message ordering.
+                * Therefore there is only one thread invoking onMessage. The user should not keep the thread busy, but preferably
+                * should dispatch it via an Executor. This will increase the performance of the topic.
+                *
+                * @param message the message that is received for the added topic
+                */
+                virtual void onMessage(Message &&message) = 0;
```
ihsandemir, line 74: i think this needs to match the signature of MessageListener on?received. Furthermore, lets try to remove the virtual methods where possible.

docsConfig
yemreinci: Sure, I think we may also consider to review all documentation comments, doxygen gives lots of warnings. There are Javadoc things in the docs that are not supported or should be done differently in doxygen.

hazelcast/include/hazelcast/client/ITopic.h
yemreinci: will do

hazelcast/include/hazelcast/client/topic/MessageListener.h
yemreinci: No, this is not sth accidental, I did the same thing on EntryListener and ItemListener as well. This is because we only accept rvalue parameters in addMessageListener, (or addItemListener and addEntryListener). If you don't return lvalue instead of rvalue ref to *this, then this won't compile:
```c++
topic.addMessageListener(
     MessageListener().
            on_receive(my_func)
)
```
But we and the compiler know that MessageListener() is rvalue here, and when on_receive is called, it will call the second overload with the `) && {`  and that will move the *this back to the caller as an rvalue. If we didnt have this overload than the user would have to use std::move explicitly to explicitly cast the returned lvalue back to rvalue.

yemreinci: About the question on StackOverFlow, I don't think we are returning a dangling ref here, we are still returning *this not a copy of *this. The problem there would still exist if the question was asked with lvalues.

hazelcast/include/hazelcast/client/topic/ReliableMessageListener.h
yemreinci: I didn't wish to convert ReliableMessageListener on this PR, I just copy pasted the virtual methods from the old MessageListener and removed the inheritance so that it wouldn't break. I thought I would do it in a separate PR because ReliableMessageListener needs extra parameters other than handler functions (isLossTolerant, etc.) and I think we should discuss about how the API should look for those.

hazelcast/include/hazelcast/client/ITopic.h
yemreinci: I removed the example here and haven't written a new one, because from what I observed, this is not a common thing through the whole codebase (addEntryListener, addItemListener don't have examples as comments). And we already have an example under the `examples/` directory. 
As a suggestion for giving examples in the doc, doxygen has a `\example` command which we can use to give references to the examples in the `examples/` directory.

examples/Org.Website.Samples/TopicSample.cpp
yemreinci: But then it is not consistent with the other listeners, can we pass EntryEvent, ItemEvent etc. as rvalue?

ihsandemir: We could, we are currently composing a separate event for each event message for each listener. If we are to make the event shared between listeners in the future, then it may be a different story but sharing events between listeners currently can only be considered for local listeners such as membership listener and such but for data structure listeners, the current way is a separate message for each listener even if they registered to the same data structure with the same flags.

hazelcast/include/hazelcast/client/ITopic.h
ihsandemir: `example` tag is a good idea, will it include the examples in the doxygen generated pages?

yemreinci: Yes, I guess it does. We need to set EXAMPLE_PATH.

hazelcast/include/hazelcast/client/topic/MessageListener.h
ihsandemir: can you give an example for the first method usage? From your answer i understand that only the second one would have been enough.

docsConfig
yemreinci: Resetted this change, opened a new issue: https://github.com/hazelcast/hazelcast-cpp-client/issues/609.

Another question, is there a specific reason why the the config file is named docsConfig? Can we rename it to Doxyfile? It is the standard thing to do and one of the things that someone will look for when they first see the project.

hazelcast/include/hazelcast/client/topic/MessageListener.h
yemreinci: This wouldn't compile, if we don't overload &:
```
MessageListener l;
l.on_receive(my_func);
topic_addMessageListener(std::move(l));
```
Well, this may look unlikely for the MessageListener, but when there are more events (EntryListener), someone might prefer this.

yemreinci: This method of overloading both & and && might look like a code repeat, but it actually isn't. I am calling the & overloaded version from the &&. So, what you implemented in the & version, applies to the && version as well. && version will always be 2 lines: call the & version, and return rvalue ref to *this. Even if we needed to sth more complex in these setter methods (e.g. calculating listener flags in EntryListener), that thing would only be implemented once in the & overload, and the && would automatically use it.

hazelcast/include/hazelcast/client/topic/MessageListener.h
yemreinci: listener or Listener ?

examples/Org.Website.Samples/TopicSample.cpp
yemreinci: Ok, converted parameters back to `Message &&`. Created an issue for converting other events: https://github.com/hazelcast/hazelcast-cpp-client/issues/610

docsConfig
ihsandemir: no reason but will need to update the release scripts when it is changed.

yemreinci: Ok, I am noting this in #609

hazelcast/include/hazelcast/client/topic/ReliableMessageListener.h
ihsandemir: can you open an issue to correct this API?

hazelcast/include/hazelcast/client/topic/MessageListener.h
ihsandemir: You can do `Listener` for now but I want to change all the code base to use all lower case, similar to the stl lib and boost lib naming.

PR #608, opened 2020-07-29
Utilize listener flags for EntryListener

yemreinci:
Now with the new EntryListener API (#604), the user can explicitly express which event types they want to listen to. 
This PR sets the listenerFlags parameter on the codec accordingly and avoids redundant events to be fired.

### DIFF
--- a/hazelcast/include/hazelcast/client/IMap.h
+++ b/hazelcast/include/hazelcast/client/IMap.h
@@ -479,7 +479,7 @@ namespace hazelcast {
                                         getName(), getContext().getClientClusterService(),
                                         getContext().getSerializationService(),
                                         std::move(listener),
-                                        includeValue, getContext().getLogger())), includeValue);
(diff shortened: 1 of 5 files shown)


hazelcast/include/hazelcast/client/IMap.h
```diff
@@ -479,7 +479,7 @@ namespace hazelcast {
                                         getName(), getContext().getClientClusterService(),
                                         getContext().getSerializationService(),
                                         std::move(listener),
-                                        includeValue, getContext().getLogger())), includeValue);
+                                        includeValue, getContext().getLogger())), includeValue, listener.flags);
```
ihsandemir, line 482: parameter order may be a problem. you are accessing flags field of the moved object !!!

yemreinci: fixed

PR #613, opened 2020-08-05
Remove `const_cast`s

yemreinci:
There are `const_cast`s used in the codebase merely because either a member function wasn't properly const-qualified or a mutex wasn't declared as mutable. And there are some cases where the const_cast is unnecessary (I assume those were necessary for the old hand-crafted atomic variables). 
This PR removes all `const_cast`s excluding those that are in the tests (we still have `const_casts`s in the tests), and adds const qualifier to member functions where necessary. There is one breaking change in the public API:

old:
```c++
std::unordered_map<std::string, std::string> &ClientConfig::getProperties();
```
new:
```c++
const std::unordered_map<std::string, std::string> &ClientConfig::getProperties() const;
```

We still have ClientConfig::setProperty for write access.

### DIFF
--- a/hazelcast/include/hazelcast/util/Sync.h
+++ b/hazelcast/include/hazelcast/util/Sync.h
@@ -73,11 +73,11 @@ namespace hazelcast {
                 v = i;
             }
 
-            operator T() {
+            operator T() const {
                 return get();
             }
 
-            T get() {
+            T get() const {
                 std::lock_guard<std::mutex> lockGuard(mutex);
                 return v;
             }
@@ -87,17 +87,17 @@ namespace hazelcast {
                 return --v;
             }
 
-            bool operator<=(const T &i) {
+            bool operator<=(const T &i) const {
                 std::lock_guard<std::mutex> lockGuard(mutex);
                 return v <= i;
             }
 
-            bool operator==(const T &i) {
+            bool operator==(const T &i) const {
                 std::lock_guard<std::mutex> lockGuard(mutex);
                 return i == v;
             }
 
-            bool operator!=(const T &i) {
+            bool operator!=(const T &i) const {
                 std::lock_guard<std::mutex> lockGuard(mutex);
                 return i != v;
             }
@@ -112,8 +112,7 @@ namespace hazelcast {
             }
 
         protected:
-
-            std::mutex mutex;
+            mutable std::mutex mutex;
             T v;
         };
     }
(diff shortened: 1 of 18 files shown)


yemreinci:
verify

hazelcast/include/hazelcast/util/Sync.h
```diff
@@ -112,8 +112,7 @@ namespace hazelcast {
             }
 
         protected:
-
-            std::mutex mutex;
+            mutable std::mutex mutex;
```
ihsandemir, line 115: this change seems to conflict with what the PR says it wants to achieve. We did hide such non-const changes in the past but surely it was not a good practice since it actually really is not a const operation to acquire a lock and may be it is better to let it seen this way from outside as well, so that the method signature is more reliable.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -1407,7 +1406,7 @@ namespace hazelcast {
                     std::vector<CipherInfo> supportedCiphers;
                     for (int i = 0; i < sk_SSL_CIPHER_num(ciphers); ++i) {
                         struct SSLSocket::CipherInfo info;
-                        SSL_CIPHER *cipher = const_cast<SSL_CIPHER *>(sk_SSL_CIPHER_value(ciphers, i));
+                        const SSL_CIPHER *cipher = sk_SSL_CIPHER_value(ciphers, i);
```
yemreinci, line 1409: @ihsandemir Can you double-check here? I am not confident with this as I didn't understand why this cast was used in the first place (maybe it used to be necessary, but now it's not.). I just removed the cast and declared the pointer as pointer-to-const-variable and it compiled.

yemreinci:
verify

hazelcast/include/hazelcast/util/Sync.h
ihsandemir: ok, per https://herbsutter.com/2013/01/01/video-you-dont-know-const-and-mutable/

hazelcast/src/hazelcast/client/network.cpp
yemreinci: I looked at the declarations of all the functions that we pass the pointer `cipher` here (SSL_CIPHER_get_name, SSL_CIPHER_get_bits, ..). They all take the cipher as `const SSL_CIPHER *c`, so I am sure it's fine to for the cipher to be const here. Merging it this way.

PR #616, opened 2020-08-24
New client protocol, ownerless client, uuid changes, services refactorings

ihsandemir:
This PR includes the following changes:

- The new client protocol 2.0 implementation. The related client protocol codec generation PR is at https://github.com/hazelcast/hazelcast-client-protocol/pull/338 . The client now works with Hazelcast server 4.0 and above.

- The ownerless client implementation changes (see Java PR https://github.com/hazelcast/hazelcast-client-protocol/pull/254)

- Changed all registration ids to boost::uuids::uuid (it was string before).

- Merged smart and non smart listener and invocation service implementations similar to Java client.

There are also some minor fixes.


yemreinci:
verify-windows

ihsandemir:
verify-linux

hazelcast/include/hazelcast/client/Member.h
```diff
@@ -48,12 +49,11 @@ namespace hazelcast {
 
             Member();
 
-            Member(const Address &address, const std::string &uuid, bool lite,
-                   const std::unordered_map<std::string, std::string> &attr);
+            Member(Address address, boost::uuids::uuid uuid, bool lite, std::unordered_map<std::string, std::string> attr);
```
sancar, line 52: Are we not using the const& anymore in the codebase to avoid copying staff?

ihsandemir: this way allows both copy and move. If called by an rvalue it will be a move. if called by lvalue it will be copied. I try calling with rvalue where possible.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -700,163 +588,177 @@ namespace hazelcast {
...
+                activeConnections.put(response.member_uuid, connection);
+
+                if (initial_connection) {
+                    cluster_id_ = new_cluster_id;
+                    fire_life_cycle_event(LifecycleEvent::LifecycleState::CLIENT_CONNECTED);
+                }
+
+                logger.info("Authenticated with server ", response.address , ":", response.member_uuid
+                            , ", server version: " , response.server_version
+                            , ", local address: ", *connection->getLocalSocketAddress());
```
sancar, line 638: What happens if you dereference an empty optional?

ihsandemir: if empty, dereferencing is either an illegal memory access or null ptr access, the process will die. I assumed that it could not be null at this point but it may be an oversight if the client disconnects immediately after successful authentication for some network failure. Hence, I am changing it to 
```                auto local_address = connection->getLocalSocketAddress();
                if (local_address) {
                    logger.info("Authenticated with server ", response.address , ":", response.member_uuid
                            , ", server version: " , response.server_version
                            , ", local address: ", *local_address);
                } else {
                    logger.info("Authenticated with server ", response.address , ":", response.member_uuid
                            , ", server version: " , response.server_version
                            , ", no local address (connection disconnected ?)");
                }
```

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -183,416 +208,272 @@ namespace hazelcast {
...
+                    return connection;
+                }
 
-            std::shared_ptr<Connection> ClientConnectionManagerImpl::getOwnerConnection() {
-                std::shared_ptr<Address> address = ownerConnectionAddress;
-                if (address.get() == NULL) {
-                    return std::shared_ptr<Connection>();
+                auto future = std::make_shared<ConnectionFuture>(address, connection, connectionsInProgress);
+                auto f = connectionsInProgress.putIfAbsent(address, future);
+                bool outstanding_connection_exists = static_cast<bool>(f);
```
sancar, line 218: Are we casting a `shared_ptr<ConnectionFuture>` to boolean?  Should we get if f.get() is null or not ?

ihsandemir: casting to bool is the same as `f.get() == nullptr`.

sancar: It seems you have used the same trick also here.
```
connection = getConnection(address);
                if (connection) {
```
That means shared ptr overrides bool operator

ihsandemir: yes, exactly shared_ptr and boost::optional both has this bool operator, and it is less code.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -183,416 +208,272 @@ namespace hazelcast {
...
-                - send auth message
-                - receive auth reply -> reply processing is offloaded to an executor. Did not start to run yet.
-                - member closes the connection -> the connection is trying to removed from map
-                                                                     but it was not there to begin with
-                - the executor start processing the auth reply -> it put the connection to the connection map.
-                - we end up with a closed connection in activeConnections map */
-                if (!connection->isAlive()) {
-                    removeFromActiveConnections(connection);
                 }
+                return protocol::ClientMessage();
```
sancar, line 364: This path looks really wierd. Can we put an assert or throw exception to emphasize that this path is not possible.

ihsandemir: It was there to make the compiler happy but surely we can put an assert. putting `assert(0);`

ihsandemir:
verify-linux

ihsandemir:
verify-linux

ihsandemir:
verify-linux

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -611,79 +493,86 @@ namespace hazelcast {
...
-                     * In that case, trying other addresses first to upgrade make more sense.
-                     */
-                    addresses.erase(*previousAddress);
-                    addresses.insert(*previousAddress);
+                std::unordered_set<Address> result;
+                for (auto &&a : std::move(addresses)) {
+                    result.insert(a);
                 }
-                return addresses;
-            }
```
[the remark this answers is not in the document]
ihsandemir: We try to implement the same log'c here https://github.com/hazelcast/hazelcast/blob/v4.0/hazelcast/src/main/java/com/hazelcast/client/impl/connection/nio/ClientConnectionManagerImpl.java#L491 
The member list should go before the provided addresses.

I realized that I changed the behaviour thinking that a set was better but reverted it to return vector<Address> now in compliance with Java.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -611,79 +493,86 @@ namespace hazelcast {
...
-                    /*
-                     * Previous owner address is moved to last item in set so that client will not try to connect to same one immediately.
-                     * It could be the case that address is removed because it is healthy(it not responding to heartbeat/pings)
-                     * In that case, trying other addresses first to upgrade make more sense.
-                     */
-                    addresses.erase(*previousAddress);
-                    addresses.insert(*previousAddress);
+                std::unordered_set<Address> result;
+                for (auto &&a : std::move(addresses)) {
+                    result.insert(a);
```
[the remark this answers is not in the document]
ihsandemir: ok fixed.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -45,51 +45,62 @@
...
+                      heartbeat(client, *this), partition_count_(-1),
+                      async_start_(client.getClientConfig().getConnectionStrategyConfig().isAsyncStart()),
+                      reconnect_mode_(client.getClientConfig().getConnectionStrategyConfig().getReconnectMode()),
+                      smart_routing_enabled_(client.getClientConfig().getNetworkConfig().isSmartRouting()),
+                      connect_to_cluster_task_submitted_(false),
+                      client_uuid_(client.random_uuid()),
+                      authentication_timeout_(heartbeat.getHeartbeatTimeout().count()),
+                      load_balancer_(client.getClientConfig().getLoadBalancer()) {
+                boost::uuids::uuid u;
+                std::memset(&u, 0, sizeof(boost::uuids::uuid));
```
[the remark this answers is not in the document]
ihsandemir: changed all similar uses.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -888,43 +798,29 @@ namespace hazelcast {
...
                       clientContext(clientContext),
                       invocationService(clientContext.getInvocationService()),
-                      authFuture(authFuture),
                       connectionId(connectionId),
                       connectedServerVersion(impl::BuildInfo::UNKNOWN_HAZELCAST_VERSION),
-                      logger(clientContext.getLogger()), asOwner(asOwner),
-                      connectionManager(clientConnectionManager), alive(true) {
+                      logger(clientContext.getLogger()), alive(true), authFuture(authFuture) {
                 socket = socketFactory.create(address, connectTimeoutInMillis);
+                std::memset(&remote_uuid_, 0, sizeof(boost::uuids::uuid));
```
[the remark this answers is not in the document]
ihsandemir: changed all similar uses.

ihsandemir:
build-windows

ihsandemir:
build-windows

ihsandemir:
verify-winndows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -112,11 +116,10 @@ namespace hazelcast {
...
 
             bool ClientConnectionManagerImpl::start() {
-                std::lock_guard<std::mutex> guard(lock);
-                if (alive) {
-                    return true;
+                bool expected = false;
+                if (!alive.compare_exchange_strong(expected, true)) {
+                    return false;
                 }
-                alive.store(true);
```
yemreinci, line 123: Do you intent to change the behavior here, or just use atomics instead of locks?
Previously if the thing was already alive (alive == true), the function returned true; but now it returns false.

sancar (approved):

PR #619, opened 2020-08-26
Fix 32 bit linux and all release builds

yemreinci:
32-bit linux builds have been failing since Boost::thread and Boost::chrono was started be linked. We use `-m32` flag to compile for 32-bit on a 64-bit host but since the flag is set after the search for boost libraries, boost's cmake files skip the 32-bit libraries. 
This PR sets the -m32/64 flag in the command line arguments for cmake in `build-linux.sh` so that the flag is available throughout the entire configuration process. We already have a similar logic in `build-windows.bat` where we set cmake's -A option as either 'win32' or 'x64' depending on the given parameter: 
https://github.com/hazelcast/hazelcast-cpp-client/blob/b5fcd27d4b19b6956c13ad83d2ea7df234a07f9d/scripts/build-windows.bat#L18-L24

Also, with this change it seems to me that the `HZ_BIT` option in CMakeLists.txt is no longer needed. It is not used for Darwin, used to define _WIN32/64 on Windows (which MSVC already does: [predefined macros](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019)) and used to set -m32/64 on Linux which now becomes redundant as this PR sets it beforehand.

Another problem with the release builds was: assert statements with side-effects. E.g:
```c++
assert(666 == in.read<int32_t>());
```
assert statements are ignored in CMake Release builds, so the call to `read<int32_t>` gets eliminated.


yemreinci:
verify-linux

PR #627, opened 2020-09-11
Fenced lock implementation

ihsandemir:
Implemented CP fenced lock and the tests.

### DIFF
--- a/hazelcast/src/hazelcast/cp/cp.cpp
+++ b/hazelcast/src/hazelcast/cp/cp.cpp
@@ -263,6 +263,8 @@ namespace hazelcast {
                      const std::string &objectName) : cp_proxy(SERVICE_NAME, name, &context, groupId, objectName) {}
 
         boost::future<bool> latch::try_set_count(int32_t count) {
+            util::Preconditions::checkPositive(count, "count must be positive!");
+
             auto request = countdownlatch_trysetcount_encode(group_id_, object_name_, count);
             return invokeAndGetFuture<bool>(request);
         }
@@ -651,6 +653,17 @@ namespace hazelcast {
         bool fenced_lock::lock_ownership_state::is_locked() {
             return fence != INVALID_FENCE;
         }
+    }
+}
+
+namespace std {
+    std::size_t
+    hash<hazelcast::cp::raft_group_id>::operator()(const hazelcast::cp::raft_group_id &group_id) const noexcept {
+        std::size_t seed = 0;
+        boost::hash_combine(seed, group_id.name);
+        boost::hash_combine(seed, group_id.seed);
+        boost::hash_combine(seed, group_id.group_id);
+        return seed;
     }
 }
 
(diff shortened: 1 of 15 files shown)


ihsandemir:
verify-linux

sancar (dismissed):

hazelcast/src/hazelcast/cp/cp.cpp
```diff
@@ -292,6 +318,354 @@ namespace hazelcast {
...
+                auto session_id = session_manager_.acquire_session(group_id_);
+                verify_locked_session_id_if_present(thread_id, session_id, true);
+                return do_lock(session_id, thread_id, invocation_uid).then(boost::launch::deferred, [=] (boost::future<int64_t> f) {
+                    try {
+                        auto fence = f.get();
+                        if (fence != INVALID_FENCE) {
+                            locked_session_ids_.put(thread_id, std::make_shared<int64_t>(session_id));
+                            return fence;
+                        }
+                        BOOST_THROW_EXCEPTION(exception::LockAcquireLimitReachedException(
```
yemreinci, line 356: Is this line of code covered by tests?
Because it says on the protocol doc [here](https://github.com/hazelcast/hazelcast-client-protocol/blob/59b52597a41bb4882eae6bbb875ead0daf9fca4e/protocol-definitions/FencedLock.yaml#L56-L57) that FencedLock::lock method returns -1 if the lock wasn't acquired. But above in the if statement, your are comparing `fence` with `INVALID_FENCE` which is equal to 0.

yemreinci:
```C++
                TEST_F(basic_lock_test, test_reentrancy) {
                    auto mutex = client->get_cp_subsystem().get_lock("my_lock");
                    for (int i = 0; i < 15; i++) {
                        auto fence = mutex->lock_and_get_fence().get();
                        ASSERT_NE(-1, fence);
                    }
                }
```
I run this test with this config:
```
<cp-subsystem>
        <cp-member-count>3</cp-member-count>
        <locks>
            <fenced-lock>
                <name>my_lock</name>
                <lock-acquire-limit>10</lock-acquire-limit>
            </fenced-lock>
        </locks>
    </cp-subsystem>
```
The assertion failed at the 11th iteration.

ihsandemir: I think that the doc at protocol is incorrect. I checked at Java code, it is the same here: https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/cp/internal/datastructures/lock/proxy/AbstractFencedLockProxy.java#L124   The related message task LockOp operation sets to INVALID_FENCE(0) at this line as https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/cp/internal/datastructures/lock/AcquireResult.java#L79   As for coverage, are you asking for  line[356](https://github.com/hazelcast/hazelcast-cpp-client/pull/627/files/1ca7b508adbb91b825dd382251c658c6f805f0e5#diff-cea164ffa8e795ad99c4c8caafb522d4R356), the LockAcquireLimitReachedException exception? it may not be covered.

yemreinci: Yeah, the protocol doc is incorrect. The member actually returns 0, not -1. I asked about the coverage to make sure if LockAcquireLimitReachedException was tested or not. Anyways, there isn't a bug here.

yemreinci (approved):

PR #635, opened 2020-09-28
Add missing std::move's

yemreinci:
I realized some rvalue references to `Data` objects were being passed through without being moved in `proxy.cpp`.


ihsandemir (approved):

yemreinci:
verify-linux

PR #636, opened 2020-09-30
CP counting_semaphore Implementation

ihsandemir:
Implemented the CP subsystem counting_semaphore and added tests.

### DIFF
--- a/hazelcast/src/hazelcast/cp/cp.cpp
+++ b/hazelcast/src/hazelcast/cp/cp.cpp
@@ -687,9 +687,9 @@ namespace hazelcast {
             return try_acquire_for(std::chrono::milliseconds::zero(), permits);
         }
 
-        boost::future<void> counting_semaphore::do_release(int32_t permits, int64_t thread_id) {
+        boost::future<void> counting_semaphore::do_release(int32_t permits, int64_t thread_id, int64_t session_id) {
             auto invocation_uid = getContext().getHazelcastClientImplementation()->random_uuid();
-            auto request = codec::semaphore_release_encode(group_id_, object_name_, internal::session::proxy_session_manager::NO_SESSION_ID, thread_id, invocation_uid, permits);
+            auto request = codec::semaphore_release_encode(group_id_, object_name_, session_id, thread_id, invocation_uid, permits);
             return toVoidFuture(spi::impl::ClientInvocation::create(context_, request, object_name_)->invoke());
         }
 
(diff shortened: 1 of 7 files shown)


hazelcast/src/hazelcast/cp/cp.cpp
```diff
@@ -653,6 +665,257 @@ namespace hazelcast {
...
+                            }
+                        });
+            });
+
+            return do_try_acquire_once().then(boost::launch::deferred, [=] (boost::future<std::pair<bool, bool>> f) {
+                auto result = f.get();
+                if (result.second) {
+                    return result.first;
+                }
+                for (;result.second;result = do_try_acquire_once().get());
```
yemreinci, line 832: I guess the condition should be `!result.second`

hazelcast/src/hazelcast/cp/cp.cpp
```diff
@@ -653,6 +665,257 @@ namespace hazelcast {
...
+                    session_manager_.release_session(group_id_, session_id, permits);
+                } catch(exception::SessionExpiredException &) {
+                    session_manager_.invalidate_session(group_id_, session_id);
+                    session_manager_.release_session(group_id_, session_id, permits);
+                    BOOST_THROW_EXCEPTION(illegal_state_exception(std::current_exception()));
+                }
+            });
+        }
+
+        exception::IllegalStateException session_semaphore::illegal_state_exception(std::exception_ptr e) {
```
yemreinci, line 856: Why this function have a return type? It is only throwing, not returning.
It could be `[[noreturn]] void session_semapho...`

hazelcast/src/hazelcast/cp/cp.cpp
```diff
@@ -653,6 +665,257 @@ namespace hazelcast {
...
+                }
+                for (;result.second;result = do_try_acquire_once().get());
+                return result.first;
+            });
+        }
+
+        boost::future<void> session_semaphore::release(int32_t permits) {
+            auto session_id = session_manager_.get_session(group_id_);
+            if (session_id == internal::session::proxy_session_manager::NO_SESSION_ID) {
+                BOOST_THROW_EXCEPTION(illegal_state_exception(nullptr));
```
yemreinci, line 840: BOOST_THROW_EXCEPTION is expecting an exception to be returned from the illegal_state_exception, but it's not returning the exception, it's throwing the exception.

yemreinci (changes requested):

hazelcast/src/hazelcast/cp/cp.cpp
ihsandemir: good catch. corrected.

hazelcast/src/hazelcast/cp/cp.cpp
ihsandemir: changed it to `throw_illegal_state_exception` with void return.

hazelcast/src/hazelcast/cp/cp.cpp
ihsandemir: fixed.

yemreinci (approved):

sancar (approved):

ihsandemir:
verify-windows

PR #639, opened 2020-10-07
Implement boomerang client (backup aware client) feature

ihsandemir:
Implemented boomerang client (backup aware client) feature.

Using the first 4-bytes of correlation id to match the connection id and the rest to match the automatically incremented per connection call id counter.

All backup events are being processed in the same thread as the related connection socket io thread.


hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -590,6 +592,7 @@ namespace hazelcast {
                 }
 
                 activeConnections.put(response.member_uuid, connection);
+                active_connection_ids_.put(connection->getConnectionId(), connection);
```
sancar, line 595: We should put to `active_connection_ids_` before we put to `activeConnections`. 
Although the window is small there is an ordering problem. An invocation can be done on a connection(we get the connection from activeConnections map), where it is not available on `active_connection_ids_` map.

ihsandemir: Good catch. fixed the order.

ihsandemir:
verify

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1321,13 +1360,44 @@ namespace hazelcast {
...
+
+                        pending_response_received_time_ = std::chrono::steady_clock::now();
+
+                        backup_acks_expected_ = expected_backups;
+
+                        // it is very important that the response is set after the backupsAcksExpected is set, else the system
+                        // can assume the invocation is complete because there is a response and no backups need to respond
+                        pending_response_ = msg;
+
+                        if (backup_acks_received_ != expected_backups) {
```
yemreinci, line 1380: Is this if statement necessary? We went into this block because `expected_backups > backup_acks_received_` was true, so `backup_acks_received_ != expected_backups` should also be true. 
I assume there isn't any other thread modifying these variables as we move from line 1368 to here, since it seems all things related to this invocation run within the same strand. 
I also checked the Java impl but these variables was declared volatile there, meaning that there is some concurrency at play, but its not true for our case.

ihsandemir:
verify

hazelcast/src/hazelcast/client/spi.cpp
ihsandemir: Yes, that seems an unneeded check. it can also be deleted at Java client. I will put a return statement there without the check.

As for non-atomic variables, yes, I do this on purpose, since I always access these variables from the same io thread (socket strand), hence i do not need atomic variables. If any access is needed I always do posting through the same io thread.

yemreinci (approved):

ihsandemir:
Performance test run: I start 4 hazelcast server (4.1-SNAPSHOT) on my local macbook pro computer. I run the client code below. The test is single thread test.

1 backup (default):

- Boomerang disabled: Total time:169940 msecs. 5884.43 ops/sec

- Boomerang enabled: Total time: 136315 msecs. 7335.95 ops/sec

Around a 25% increase in throughput is observed for single thread.


```
#include <iostream>
#include <hazelcast/client/HazelcastClient.h>

using namespace hazelcast;
using namespace hazelcast::client;

int main() {
    ClientConfig config;
    config.backup_acks_enabled(false);
    HazelcastClient client(config);
    auto map = client.getMap("throughput_test");
    auto start = std::chrono::steady_clock::now();
    constexpr static int NUM_OPERATIONS = 1000 * 1000;
    for (int i = 0; i < NUM_OPERATIONS; ++i) {
        map->put(i, i).get();
    }
    auto end = std::chrono::steady_clock::now();
    auto duration_msecs = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
    std::cout << "Total time:" << duration_msecs << " msecs. " << (double(NUM_OPERATIONS) * 1000) / duration_msecs << " ops/sec\n";
    return 0;
}
```

ihsandemir:
verify-linux

sancar (approved):

ihsandemir:
verify-linux

ihsandemir:
verify-linux

PR #644, opened 2020-10-13
Removed GroupConfig which is not needed anymore

ihsandemir:
Removed GroupConfig which is not needed anymore. The user can set the username password security config using the `ClientConfig::setCredentials(std::shared_ptr<security::username_password_credentials>(...))`.  Similarly user can also set a token credentials.

Added tests for username password and token credentials.


sancar (approved):

ihsandemir:
opened issue for unrelated test failure at linux: https://github.com/hazelcast/hazelcast-cpp-client/issues/645

sancar (approved):

yemreinci (dismissed):

yemreinci (approved):

PR #655, opened 2020-10-15
Add support to reading fragmented messages

ihsandemir:
Added the support to read fragmented messages.

Also removed unused API ClientConfig::addListener(MembershipListener *listener).


sancar:
Can we add a test for it? It would be better to verify this. Since the member does not send any fragmented messages, you probably need to implement split logic also. 
https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/test/java/com/hazelcast/client/impl/protocol/util/ClientMessageEncoderDecoderTest.java#L299

hazelcast/src/hazelcast/client/protocol.cpp
```diff
@@ -513,26 +530,21 @@ namespace hazelcast {
...
+                            message->drop_fragmentation_frame();
+                            if (message->is_flag_set(ClientMessage::BEGIN_FRAGMENT_FLAG)) {
                                 // put the message into the partial messages list
                                 addToPartialMessages(message);
-                            } else if (message->isFlagSet(ClientMessage::END_FLAG)) {
+                            } else if (message->is_flag_set(ClientMessage::END_FRAGMENT_FLAG)) {
                                 // This is the intermediate frame. Append at the previous message buffer
                                 appendExistingPartialMessage(message);
-                                isCompleted = true;
                             }
```
yemreinci, line 546: What happens if a message has 3 or more fragments? We are handling messages either with BEGIN_FRAGMENT_FLAG or END_FRAGMENT_FRAGMENT set but what about the middle ones?

hazelcast/src/hazelcast/client/protocol.cpp
```diff
@@ -547,15 +559,15 @@ namespace hazelcast {
             bool ClientMessageBuilder::appendExistingPartialMessage(std::unique_ptr<ClientMessage> &msg) {
                 bool result = false;
 
-/* TODO
-                MessageMap::iterator foundItemIter = partialMessages.find(msg->getCorrelationId());
-                if (partialMessages.end() != foundItemIter) {
-                    foundItemIter->second->append(msg.get());
-                    if (msg->isFlagSet(ClientMessage::END_FLAG)) {
+                auto found = partialMessages.find(msg->getCorrelationId());
```
yemreinci, line 562: Do the intermediate and the end message fragments have the correlationId? I'm looking at the protocol doc, and as far as I understand, only the first frame of the message contain the correlationId.
Shouldn't we use the fragment id to match fragments of the same message?

yemreinci (changes requested):

sancar (approved):

yemreinci (approved):

Issue #661, opened 2020-10-26
Test failure - basic_lock_test.test_lock_auto_release_on_client_shutdown


ihsandemir:
test basic_lock_test.test_lock_auto_release_on_client_shutdown rarely fails on linux. failure is: 
```
/home/jenkins/jenkins_slave/workspace/cpp-linux-32-STATIC-Debug/hazelcast/test/src/cp_test.cpp:710
Expected equality of these values:
  "0"
  response.result
    Which is: "1"
```

see http://jenkins.hazelcast.com/view/C++/job/cpp-linux-32-STATIC-Debug/239/testReport/junit/(root)/basic_lock_test/test_lock_auto_release_on_client_shutdown/

ihsandemir:
still fails:

http://jenkins.hazelcast.com/job/cpp-pr-builder-windows/578/testReport/junit/(root)/basic_lock_test/test_lock_auto_release_on_client_shutdown/

Stacktrace
/home/jenkins/jenkins_slave/workspace/cpp-pr-builder_2/hazelcast/test/src/cp_test.cpp:711
Value of: __result__
  Actual: false
Expected: true

PR #662, opened 2020-10-26
CP test_lock_auto_release_on_client_shutdown should check lock released eventually

ihsandemir:
CP test_lock_auto_release_on_client_shutdown test now does the check on released lock repeatedly until the lock is released eventually.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/661


ihsandemir:
There is an alternative fix to this PR:

- Add terminate method to client in addition to shutdown. Make the client wait on cp session close invocation responses on shutdown.

pros:
- when client shutdown returns server side session cleanup is guaranteed to be completed.

cons:
- slows down client shutdown.

- if there are many sessions this will mean many invocations and late shutdown.

- client does just ignore the invocation response, hence what is the point of waiting for the response?

Based on the discussions, my personal take on the subject is not to add a new API to the client. Let the client shutdown fast. In client shutdown, we guarantee that the invocation message is transferred via the socket to the server before shutdown, hence the server will definitely run the operation to cleanup sessions. The client will just not get the response of the operation to be discarded anyway. Thus, there will be no leak at the server side.

yemreinci (approved):

PR #671, opened 2020-10-30
Make ClientConfig::getReliableTopicConfig return reference

### LINKED ISSUE #649: ClientConfig::getReliableTopicConfig can return const boost::optional<config::ReliableTopicConfig> &
Do not return raw pointer but return const optional reference.

yemreinci:
Fixes #649.

There is no need for `optional` here because we construct and return a default config object if the key doesn't exist. The returned pointer was never `nullptr`.


yemreinci:
verify-windows

ihsandemir (approved):

yemreinci:
verify-windows

Issue #653, opened 2020-10-15
Change ClientNetworkConfig ConnectionAttemptPeriod and ConnectionTimeout to use std::chrono


ihsandemir:
ClientNetworkConfig ConnectionAttemptPeriod and ConnectionTimeout time ralted APIs can now use std::chrono.

PR #674, opened 2020-11-03
Use std::chrono::duration for timeout values of network configurations

ihsandemir:
Use std::chrono::duration for timeout values of network configurations.

Changed the durations to std::chrono::milliseconds to be more explicit in the API.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/653


sancar (approved):

ihsandemir:
verify-windows

Issue #663, opened 2020-10-27
Change API from camel case to snake case


ihsandemir:
In order to make the API more C user friendly, we can change the whole API to be snake case rather than the current camel case.

PR #680, opened 2020-11-06
camelCase to snake_case conversion

ihsandemir:
In this PR, I converted the API from camelCase to snake_case which is more used at the C/C++ community. The file names changes in the same way. 

The API is basically the same. The only name changes are:
    - Removed namespace cluster.
    - Moved `member_selector` to hazelcast::client namespace.
    - renamed `client` to `local_endpoint`.

All the rest API changes are just naming case change.

An example command used for changing method cases is:

`run-clang-tidy.py -fix -header-filter='.*' -checks='-*,readability-identifier-naming' -config="{CheckOptions: [ {key: readability-identifier-naming.PublicMemberCase, value: lower_case} ]}"`

Added `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON` cmake flag on cmake build locally on my machine to produce the compilation database `compile_commands.json` for the script to find the files properly.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/663


ihsandemir:
verify

ihsandemir:
verify-linux

sancar:
How about we just remove `cluster` namespace ? I just take a look and it seems there is not much usage of it anyway. For example, cluster service is not under `cluster` namespace.

sancar:
Hazelcast.h still starts with capital letters. Is that intentional? Seems odd between others.

sancar:
We have `hz_client` and `hazelcast_client` seems difficult to separate. 
Should we rename `hz_client` to something like `client_endpoint`

ihsandemir:
> How about we just remove `cluster` namespace ? I just take a look and it seems there is not much usage of it anyway. For example, cluster service is not under `cluster` namespace.

It was left over, changed it.

sancar (approved):

ihsandemir:
verify-windows

ihsandemir:
verify-linux

sancar:
👍

PR #685, opened 2020-11-12
Make client_config move-only, Use unique_ptr for load_balancer

### LINKED ISSUE #650: ClientConfig::set/getLoadBalancer API can be changed to use movable non-pointer implementation as done for Listeners
Change ClientConfig::set/getLoadBalancer API to use movable non-pointer implementation as done for Listeners.

yemreinci:
Copying the configuration object and using it for multiple clients is problematic. For example, if you set a load_balancer to a config object and then use it for two clients, the init method of the same load_balancer will be called twice:

```c++
client_config cfg;
load_balancer* lb = new some_load_balancer{};
cfg.set_load_balancer(lb);
hazelcast_client cli1{ cfg };
hazelcast_client cli2{ cfg };

// init method of the load balancer (lb->init) will be called twice.
```

So, a config object shouldn't be used more than once, that's why this PR makes the `client_config` a move-only type.

Also, `unique_ptr` is used to set and hold the load_balancer. 

Fixes #650.
Depends on #673.


yemreinci:
verify-windows

hazelcast/include/hazelcast/client/client_config.h
```diff
@@ -209,7 +217,7 @@ namespace hazelcast {
             *
             * \return loadBalancer
             */
-            load_balancer *const get_load_balancer();
+            std::unique_ptr<load_balancer> get_load_balancer();
```
ihsandemir, line 220: should we rather return reference here? With this version it looks more like  method like `move_load_balancer`. Consider that the user usually uses getters as read-only which is different from this usage.

hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
```diff
@@ -126,6 +126,8 @@ namespace hazelcast {
 
                 void notify_backup(int64_t call_id);
 
+                std::unique_ptr<load_balancer> get_load_balancer();
```
ihsandemir, line 129: return reference?

hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -686,14 +690,12 @@ namespace hazelcast {
             return redo_operation_;
         }
 
-        load_balancer *const client_config::get_load_balancer() {
-            if (!load_balancer_)
-                return &default_load_balancer_;
-            return load_balancer_;
+        std::unique_ptr<load_balancer> client_config::get_load_balancer() {
```
ihsandemir, line 693: just in case that the user can also make this call, i think that this can only return a read-only reference and we can have another internal api for moving it to lifecycle_service or just use reference also at lifecycle_service, since the config will be kept in client instance impl until it is destroyed.

yemreinci: OK, I am keeping the load balancer in the config, and holding a reference to it in ConnectionManagerImpl.

hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
yemreinci: This one was actually left over, it has no definition. Removed it.

ihsandemir (approved):

PR #688, opened 2020-11-13
Logger enhancements

### LINKED ISSUE #682: Logging comment
Just noticed this while browsing the latest code in master

```
./hazelcast/src/hazelcast/logger.cpp:          << HAZELCAST_VERSION << "] [" // TODO once we have an API for the library version, use that instead
```

There is an API for the library version now - [#657](https://github.com/hazelcast/hazelcast-cpp-client/pull/657)

yemreinci:
This PR includes some enhancements for the logging implementation:

* Use the new `client::version()` API instead of the `HAZELCAST_VERSION` macro in the default formatter. Fixes #682.
* Add tests for logger_config.
* Suppress MSVC C4151 warning in `logger.h`
* Don't use `localtime`. It is not guaranteed to be thread-safe. Instead, use `localtime_s` or `localtime_r` depending on the platform.


hazelcast/src/hazelcast/logger.cpp
```diff
@@ -53,6 +54,22 @@ void logger::log(const char *file_name, int line,
     handler_(instance_name_, cluster_name_, file_name, line, lvl, msg);
 }
 
+namespace {
+
```
ihsandemir, line 58: can we just utilize utilities from boost and not have an extra method? did you check boost?  e.g. https://stackoverflow.com/questions/2492775/get-local-time-with-boost  The `boost::posix_time::second_clock::local_time()` uses `localtime_r`

hazelcast/include/hazelcast/logger.h
```diff
@@ -67,5 +72,8 @@ enum class logger::level : int {
 
 HAZELCAST_API std::ostream& operator<<(std::ostream&, logger::level level);
 
+} // namespace hazelcast
 
-} // namespace hazelcast
\ No newline at end of file
+#if  defined(WIN32) || defined(_WIN32) || defined(WIN64) || defined(_WIN64)
+#pragma warning(pop)
+#endif
```
ihsandemir, line 79: no newline at the end?

hazelcast/src/hazelcast/logger.cpp
ihsandemir: We decided to do this later before 4.0 release. An issue will be opened for this.

ihsandemir (approved):
Once the issue is opened for `localtime` all is OK.

ihsandemir:
related to https://github.com/hazelcast/hazelcast-cpp-client/issues/692

PR #701, opened 2020-11-20
Eliminate the extra include and examples folders under the platform specific folders in release zip

ihsandemir:
Change the release folder structure similar to:

- `hazelcast/Windows_64`
    - `lib`: Shared and static library directory.
        - `tls`: Contains the library with TLS (SSL) support enabled.
    - `include`: Directory you need to include when compiling your project.
    - `examples`: Contains various examples for each C++ client feature. Each example produces an executable which you can run in a cluster. You may need to set the server IP addresses for the examples to run.

Also update the related section at README.


README.md
```diff
@@ -323,12 +328,13 @@ The simplest way to build the project from source code is to run the script `scr
...
 The exact same options work for the windows batch script `scripts/build-windows.bat`.
 
-The project depends on an up-to-date [Boost](http://boost.org/) library version.
-It also depends on OpenSSL only if you compile the project enabled with the TLS feature (i.e. if `COMPILE_WITHOUT_SSL` is NOT provided) 
+The project depends on OpenSSL only if you compile the project enabled with the TLS feature (i.e. if `COMPILE_WITHOUT_SSL` is NOT provided) 
 
 ### 1.3.3 Building Hazelcast Tests Source Code
 The Hazelcast project uses the [CMake](https://cmake.org/) build system. Hence, you should have cmake installed and visible to your environment Path.
 
+The test sources depends on the Boost development libraries (thread and chrono), OpenSSL (only needed if compiling with SSL support) and [apache thrift](https://github.com/apache/thrift) (needed for communications to remote controller).
```
yemreinci, line 336:
```suggestion
The test sources depend on the Boost development libraries (thread and chrono), OpenSSL (only needed if compiling with SSL support) and [apache thrift](https://github.com/apache/thrift) (needed for communications to remote controller).
```

README.md
```diff
@@ -438,19 +443,21 @@ pass this object to the client when starting it, as shown below.
...
 ```
 
 If you run the Hazelcast IMDG members in a different server than the client, you most probably have configured the members' ports and cluster
 names as explained in the previous section. If you did, then you need to make certain changes to the network settings of your client.
 
-### 1.4.2.1 Group Settings
+### 1.4.2.1 Cluster Name
 
-You need to provide the group name of the cluster, if it is defined on the server side, to which you want the client to connect.
+You only need to provide the name of the cluster if it is explcitely configured at the server side (otherwise the default value of `dev` is used).
```
yemreinci, line 456:
```suggestion
You need to provide the name of the cluster only if it is explicitly configured on the server-side (otherwise the default value of `dev` is used).
```

hazelcast/src/hazelcast/client/client_impl.cpp
```diff
@@ -377,8 +377,8 @@ namespace hazelcast {
 
         constexpr int address::ID;
 
-        const byte address::IPV4;
-        const byte address::IPV6;
+        constexpr byte address::IPV4;
+        constexpr byte address::IPV6;
```
yemreinci, line 381:
```suggestion
```

They are defined in `address.h`.

yemreinci (changes requested):

yemreinci (approved):

PR #705, opened 2020-11-24
Make the project compile at gcc 4.8.5 

ihsandemir:
Fixes to make gcc 4.8.5 compile the project in release mode. 

Due to missing C++11 feature implementations gcc 4.8.5 :

-  `std::for_each` does not compile. changed to for loop.

- Changed `std::atomic_int32_t` to `std::atomic<int32_t>`.

- These expressions are not supported:
```
static constexpr auto noop_entry_handler = util::noop<entry_event &&>;
```
hence removed constexpr.

- replaced std::regex with boost::replace_all and such methods inside the boost/algorithm package which is header-only dependency. (regex is not implemented in gcc4.8.5) 

Community request https://hazelcastcommunity.slack.com/archives/C015Q2TUBKL/p1605689527492800


hazelcast/src/hazelcast/client/stats.cpp
```diff
@@ -206,12 +206,12 @@ namespace hazelcast {
                 Statistics::PeriodicStatistics::PeriodicStatistics(Statistics &statistics) : statistics_(statistics) {}
 
                 std::string Statistics::escape_special_characters(const std::string &name) {
-                    std::regex reComma(",");
-                    std::string escapedName = std::regex_replace(name, reComma, std::string("\\,"));
-                    std::regex reEqual("=");
-                    escapedName = std::regex_replace(escapedName, reEqual, std::string("\\="));
-                    std::regex reBackslash("\\\\");
-                    escapedName = std::regex_replace(escapedName, reBackslash, std::string("\\\\"));
+                    boost::regex reComma(",");
```
sancar, line 209: If this is the only usage, we probably don't need regex dependency. This part should not be complicated to rewrite without regex.

ihsandemir:
verify-windows

ihsandemir:
verify-windows

hazelcast/src/hazelcast/client/stats.cpp
ihsandemir: Removed the regex dependency.

sancar (approved):

PR #706, opened 2020-11-25
`std::rethrow_if_nested(std::current_exception()) does not work in gcc 4.8

ihsandemir:
Fix to make gcc 4.8.5 compile the project in release mode.

- `std::rethrow_if_nested(std::current_exception());` does not work.

Hence, replaced it with the explicit exception.

Also, updated the README to make compiling from source section more visible.


sancar (approved):

ihsandemir:
verify-linux

PR #713, opened 2020-11-26
Fix documentation config

### LINKED ISSUE #609: Fix EXCLUDE_PATTERNS in docsConfig
EXCLUDE_PATTERNS includes directories that should not be excluded.  
Also rename docsConfig to Doxyfile, update doxygen calls in release scripts.

yemreinci:
Fixes #609.

This PR renames the file `docsConfig` to `Doxyfile` and updates the `EXCLUDE_PATTERNS` and `EXCLUDE_SYMBOLS` parameters. 

The standard name for a Doxygen configuration file is `Doxyfile`. Doxygen command line interface implicitly use this name if a configName parameter isn't passed.

Here are the changes in the configuration parameters with their explanations:

#### EXCLUDE_PATTERNS
* Remove `*/exception/pimpl/*`: no such directory exists.
* Remove `*/iTest/*`: no such directory
* Remove `*/executor/*`: no such directory other than `hazelcast/test/src/executor/` which is already ignored by pattern `*/test/*`
* Remove `*/multimap/*`: no such directory other than the one under `examples/distributed-map/`
* Remove `*/queue/*`: no such directory
* Remove `*/exception/pimpl/*`: no such directory
* Replace `*/topic/*` with `*/topic/impl/*`: There are `listener.h`, `reliable_listener.h` and `message.h` under `topic/`
* Add `*/internal/*`: to exclude classes under `hazelcast/client/internal/`
* Add `*/cp/cp_impl.h`

#### EXCLUDE_SYMBOLS
* Add `*::internal::*`, `*::impl::*`, `*::pimpl::*`: to exclude classes that are defined under excluded directories but have forward declarations somewhere else. For example, even though the class `serialization::pimpl::ClassDefinitionContext` is defined under `serialization/pimpl/`, it still ends up in the documentation because it is forward-declared in `serialization.h`. This config parameter solves the problem.


ihsandemir (commented):
Can you also regenerate the docs folder contents with this change so that the docs are up to date?

ihsandemir (dismissed):

ihsandemir (approved):

PR #721, opened 2020-12-03
Added serialization tests using the Java server side and fix to string serialization

ihsandemir:
Added serialization tests to test the serialization using the server side access. There are two test types: 1.) value is put from C++ client and read from Java server side. 2.)Value is put by server side Java code and read by the C++ client.

Fixed the utf serialization to put the bytes length instead of character count.

Also, removed the no longer needed util::Bits endian conversion methods.


hazelcast/test/src/HazelcastTests2.cpp
```diff
@@ -1743,6 +1657,282 @@ namespace hazelcast {
...
+                std::string value = u8"Iñtërnâtiônàlizætiøn";
+                map_->set("key", value).get();
+                ASSERT_EQ(value, (*map_->get<std::string, std::string>("key").get()));
+                auto response = get_value_from_server();
+                ASSERT_TRUE(response.success);
+                ASSERT_EQ(value, response.result);
+            }
+
+            TEST_F(SerializationWithServer, test_emoji) {
+                std::string value = u8"1⚐中\\uD83D\\uDCA62\\uD83D\\uDE2D\\u200D\\uD83D\\uDE46\\uD83D\\uDE145";
```
yemreinci, line 1790: Are you aware of the fact that you are not sending the encoding of the emojis but instead sending the Java escaped representations of them?

With the double backslashes (`\\`) escaped the string literally becomes: `\uD83D\uDCA62\uD83D\uDE2D\u200D\uD83D\uDE46\uD83D\uDE145`, so as far as the client and the server is concerned, this is just a string of ASCII characters, not emojis. 

For example, the correct escaping of the emoji 💦 in C++ should be like this: https://godbolt.org/z/Kd317v.

Both C++ and Java have `\uNNNN` for UTF character escaping but the meaning is different. See https://en.cppreference.com/w/cpp/language/escape.

hazelcast/test/src/HazelcastTests2.cpp
```diff
@@ -1743,6 +1657,282 @@ namespace hazelcast {
...
+                map_->set("key", value).get();
+                ASSERT_EQ(value, (*map_->get<std::string, std::string>("key").get()));
+                auto response = get_value_from_server();
+                ASSERT_TRUE(response.success);
+                ASSERT_EQ(value, response.result);
+            }
+
+            TEST_F(SerializationWithServer, test_utf_chars) {
+                std::string value = u8"\\u0040\\u0041\\u01DF\\u06A0\\u12E0\\u1D306";
```
yemreinci, line 1799: Same thing goes here.

yemreinci (changes requested):

hazelcast/test/src/HazelcastTests2.cpp
ihsandemir: Changed.

yemreinci (approved):

PR #723, opened 2020-12-04
Support CMake for users, Add install target, Update CI scripts

yemreinci:
This pull-request addresses mainly two issues:

1. The build system that our CMakeLists.txt files generate does not have an `install` target. This PR adds an install target that copies the binaries and header files into user's desired installation directory. 
2. The library does not support CMake for users. With this PR, users who are using CMake for their projects will be able to call `find_package(hazelcastcxx)` and and link against the library without specifying any compiler or linker flags.

Here is the summary of all changes:
1. Changes in the main `CMakeLists.txt` script:
    - Configuration options are provided using CMake's [option](https://cmake.org/cmake/help/latest/command/option.html) command, and they are documented. Options are renamed to: `WITH_OPENSSL`, `BUILD_STATIC_LIB`, `BUILD_SHARED_LIB`, `DISABLE_LOGGING`, `BUILD_TESTS`, `BUILD_EXAMPLES`. `HZ_BIT` option is removed.
    - Building the static and the shared library together is made possible. 
    - Build type (`CMAKE_BUILD_TYPE`) is set to `Release` for single-config generators (e.g. Makefiles) if it is empty.
    - Compiler flags and options that aren't required to build the project (e.g. `-Wall`, coverage flags, etc.) are moved to CI scripts.
    - Use of discouraged and legacy CMake commands are replaced with modern alternatives. E.g. `link_libraries` is replaced with `target_link_libraries`.
    - Necessary compiler options and definitions are set using `target_compile_options` and `target_compile_definitions` with proper `PUBLIC` or `PRIVATE` scopes. 
    - Library version is set using `set_target_properties` so that CMake can create version-related symlinks for the target system. 
    - Library name is set to one of `hazelcastcxx`, `hazelcastcxx_static`, `hazelcastcxx_ssl`, `hazelcastcxx_ssl_static` depending on the library type and whether it depends on OpenSSL.
    - `install` target is added. 
    - The project now supports CMake for users. Necessary config and target scripts are now placed in the installation directory and a CMake user can use the library with ease, without dealing with compiler and linker flags for the library and its dependencies. Example:
    ```
    find_package(hazelcastcxx)
    target_link_libraries(myexe PRIVATE hazelcastcxx)
    ```

2. Changes in CI scripts:
    - All CI scripts are now under the `scripts/` directory. No scripts at the topmost directory level.
    - All scripts take their parameters using environment variables, and forward command line arguments directly to CMake. For example:
    ```
    BUILD_DIR=build-debug ./build-unix.sh -DBUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug
    ```
    - `releaseWindows.bat`, `releaseWindowsDev.bat`, `releaseLinux.sh`, `releaseOSX.sh` are deleted since we'll no longer release binaries.
    - `build-linux.sh` and `test-linux.sh` are renamed to `build-unix.sh` and `test-unix.sh` as they work on MacOS (and possibly other unix-like systems) too. 
    - Parameters to scripts are documented.
    - `test-unix.sh` and `test-windows.sh` don't implicitly build the library. They only run tests and do test-related stuff (starting RC, core dump handling etc.).
    - Scripts in the form `verifyRelease<OS>.sh` are removed, `verify-installation-unix.sh` and `verify-installation-windows.bat` are implemented. They try to build examples using an installed instance of the library. 
    - `do-all-unix.sh` and `do-all-windows.bat` are implemented. They do all CI steps: build, test, install and verify.

3. CMake's [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) is used to automatically generate the export header. `hazelcast/util/hazelcast_dll.h` is renamed to `hazelcast/util/export.h`. `hazelcast/util/byte.h` header is created to put the definition for `hazelcast::byte`

4.  Uses of functions that are named `min` or `max` are put in parentheses to remove `-DNOMINMAX` dependency. Relevant [SO answer](https://stackoverflow.com/a/22023122).

5. The name of test executable is renamed from `clientTest` to `client_test`.

**Note:** I will send a separate PR to update the README sections regarding installation and usage.


examples/distributed-map/CMakeLists.txt
```diff
@@ -14,15 +14,15 @@
...
-add_subdirectory(entry-listener)
-add_subdirectory(entry-processor)
-add_subdirectory(eviction)
-add_subdirectory(index)
-add_subdirectory(locking)
-add_subdirectory(map-interceptor)
-add_subdirectory(multimap)
-add_subdirectory(partitionaware)
-add_subdirectory(removeAll)
+#add_subdirectory(near-cache)
```
ihsandemir, line 17: will you revert back?

hazelcast/test/src/CMakeLists.txt
```diff
@@ -22,26 +18,26 @@ IF (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     message(STATUS "Using THRIFT_INCLUDE_DIR: ${THRIFT_INCLUDE_DIR}")
     message(STATUS "Using THRIFT_LIB_DIR: ${THRIFT_LIB_DIR}")
     include_directories(${THRIFT_INCLUDE_DIR})
-    target_link_directories(${HZ_EXE_NAME} PRIVATE ${THRIFT_LIB_DIR})
-    target_link_libraries(${HZ_EXE_NAME} PRIVATE  thrift)
+    target_link_directories(clientTest PRIVATE ${THRIFT_LIB_DIR})
```
ihsandemir, line 21:
```suggestion
    target_link_directories(client_test PRIVATE ${THRIFT_LIB_DIR})
```

hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -1607,7 +1607,7 @@ namespace hazelcast {
         namespace test {
             class VersionTest: public ClientTestSupport {};
             TEST_F(VersionTest, test_client_version) {
-                ASSERT_EQ(HAZELCAST_VERSION, version());
+                ASSERT_EQ("4.0.0", version());
```
ihsandemir, line 1610: hard coded? will it change for every release?

CMakeLists.txt
```diff
@@ -13,236 +13,280 @@
...
+     "./hazelcast/src/*cpp"
+     "./hazelcast/generated-sources/src/*cpp")
+FILE(GLOB_RECURSE HEADER_FILES
+     "./hazelcast/include/*h"
+     "./hazelcast/generated-sources/src/*h")
+
+# options
+option(WITH_OPENSSL
+       "Use OpenSSL, setting this option to OFF causes SSL-related features to be disabled."
+       ON)
```
ihsandemir, line 58: should we have the default version SSL OFF not to cause OpenSSL dependency by default?

CMakeLists.txt
```diff
@@ -13,236 +13,280 @@
...
+    if (WITH_OPENSSL)
+        target_compile_definitions(
+            ${name}
+            PUBLIC HZ_BUILD_WITH_SSL
+        )
+        target_link_libraries(${name} PUBLIC OpenSSL::SSL OpenSSL::Crypto)
+    endif()
+
+    # windows-specific compile flags
+    if (WIN32)
```
ihsandemir, line 158: We used MSVC above, and the export flags may be only required when compiling with MSVC hence, this content may be added to the MSVC block above.

yemreinci: I was also thinking about Windows compilers other than MSVC while writing this. (There are Intel and MinGW, I suppose).
For example, MinGW doesn't have the /bigobj option, but has the dllimport/export logic. 

Anyways, I am planning to use CMake's [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html), so that CMake can handle this import/export stuff for us. In that case, we won't need this block at all.

CMakeLists.txt
yemreinci: Yes, you are right, we should. I will change the default to `OFF`.

hazelcast/test/src/CMakeLists.txt
```diff
@@ -5,13 +5,10 @@ FILE(GLOB_RECURSE HZ_TEST_HEADERS "./*h")
 
 set(GOOGLETEST_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/hazelcast/test/googletest/googletest/include)
 
-SET(HZ_EXE_NAME clientTest_${HZ_LIB_TYPE}_${HZ_BIT})
+add_executable(client_test ${HZ_TEST_SOURCES} ${HZ_TEST_HEADERS})
```
ihsandemir, line 8: should we change min re cmake version the same for all cmake files? was it forgotten here or was it needed to be different?

hazelcast/test/src/CMakeLists.txt
```diff
@@ -63,8 +60,16 @@ IF (${CMAKE_SYSTEM_NAME} MATCHES "Windows")
         add_definitions(-DHAZELCAST_USE_SHARED)
```
ihsandemir, line 60: `HAZELCAST_USE_SHARED` still used?

hazelcast/test/src/CMakeLists.txt
```diff
@@ -53,7 +50,7 @@ IF (${CMAKE_SYSTEM_NAME} MATCHES "Windows")
     message(STATUS "Using THRIFT_INCLUDE_DIR: ${THRIFT_INCLUDE_DIR}")
     message(STATUS "Using THRIFT_LIBRARIES: ${THRIFT_LIBRARIES}")
     include_directories(${THRIFT_INCLUDE_DIR})
-    target_link_libraries(${HZ_EXE_NAME} PRIVATE ${THRIFT_LIBRARIES})
+    target_link_libraries(client_test PRIVATE ${THRIFT_LIBRARIES})
 
     IF (${HZ_LIB_TYPE} MATCHES "STATIC")
         message(STATUS "Shall use the STATIC Hazelcast library for the tests.")
```
ihsandemir, line 56: `DHAZELCAST_USE_STATIC` was being defined from the main `CMakeLists.txt` file. do we define again?

ihsandemir:
Tried at windows 👍 
```
 cmake -S . -B build -DCMAKE_TOOLCHAIN_FILE=C:\\vcpkg\\scripts\\buildsystems\\vcpkg.cmake
 cmake --build build --verbose --parallel 8
$ cmake --install build --prefix my_install
-- Install configuration: "Release"
CMake Error at build/cmake_install.cmake:51 (file):
  file INSTALL cannot find
  "C:/Users/ihsan/Desktop/work/src/hazelcast-cpp-client/build/Release/hazelcastcxx.dll":
  No error.
```

By default, it generated Debug lib. But I expected Release lib to be produced.

```
$ ls build
ALL_BUILD.vcxproj          cmake_install.cmake  CMakeFiles  hazelcast-cpp-client.sln  hazelcastcxx.vcxproj          hazelcastcxx-config.cmake          include          INSTALL.vcxproj.filters  ZERO_CHECK.vcxproj
ALL_BUILD.vcxproj.filters  CMakeCache.txt       Debug       hazelcastcxx.dir          hazelcastcxx.vcxproj.filters  hazelcastcxx-config-version.cmake  INSTALL.vcxproj  x64                      ZERO_CHECK.vcxproj.filters
```

hazelcast/test/src/CMakeLists.txt
yemreinci: All version are set to 3.10.

yemreinci:
verify-windows

yemreinci:
verify-linux

scripts/do-all-unix.sh
```diff
@@ -0,0 +1,59 @@
...
+./scripts/build-unix.sh                      \
+	-DCMAKE_BUILD_TYPE=$BUILD_TYPE             \
+    -DCMAKE_INSTALL_PREFIX=$DESTINATION      \
+    -DBUILD_STATIC_LIB=$BUILD_STATIC_LIB     \
+    -DBUILD_SHARED_LIB=$BUILD_SHARED_LIB     \
+	  -DWITH_OPENSSL=$WITH_OPENSSL             \
+    -DBUILD_TESTS=ON                         \
+	  -DBUILD_EXAMPLES=OFF
+
+./scripts/test-unix.sh
```
ihsandemir, line 45: for speeding up the tests, the tests can be run at background in parallel to building the examples. examples build takes a while, hence it may decrease the build time significantly.

ihsandemir (approved):

PR #724, opened 2020-12-04
Add the endian byte order configuration to `serialization_config`

ihsandemir:
Added the endian byte order configuration to `serialization_config`.

### DIFF
--- a/hazelcast/src/hazelcast/client/serialization.cpp
+++ b/hazelcast/src/hazelcast/client/serialization.cpp
@@ -441,17 +441,13 @@ namespace hazelcast {
 
                 template<>
                 void data_output::write(char i) {
-                    if (is_no_write_) { return; }
                     // C++ `char` is one byte only, `char16_t` is two bytes
-                    write<byte>(0);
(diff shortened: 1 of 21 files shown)


hazelcast/src/hazelcast/client/serialization.cpp
```diff
@@ -453,21 +457,33 @@ namespace hazelcast {
...
+                        boost::endian::native_to_little_inplace(value);
+                    }
                     output_stream_.insert(output_stream_.end(), (byte *) &value, (byte *) &value + util::Bits::SHORT_SIZE_IN_BYTES);
                 }
 
                 template<>
                 void data_output::write(int32_t value) {
                     if (is_no_write_) { return; }
-                    boost::endian::native_to_big_inplace(value);
+                    if (byte_order_ == boost::endian::order::big) {
```
sancar, line 471: We should address little endianness on 
`write(char i) ` and `write(char16_t i) ` as well.

sancar (approved):

PR #727, opened 2020-12-09
Make load_balancer config API consistent with the other listener config APIs

### LINKED ISSUE #725: Change load_balancer to work with lambda functions similar to the event handlers
We previously implemented the load_balancer as an interface for the configuration but with C++11, it is sometimes easier to use lambda functions instead and share any state that needs to be shared between function calls.

ihsandemir:
Changed the load balancer to be a concrete class and work with `std::function`s. Removed the unneeded load balancer classes.

Also changed the return type of `membership_event::get_cluster` not to be `const`, because it has subscription methods which can not be used if it returns const cluster.

fixes #725


ihsandemir:
verify-windows

sancar (approved):

hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -703,12 +704,21 @@ namespace hazelcast {
...
             if (!load_balancer_) {
-                load_balancer_.reset(new impl::RoundRobinLB{});
+                auto index = std::make_shared<std::atomic<size_t>>(0);
+                load_balancer_ = load_balancer().next([=](cluster &c) {
+                    auto members = c.get_members();
+                    if (members.empty()) {
+                        return boost::optional<member>();
+                    }
+                    auto i = index->load();
+                    index->store(i + 1);
```
yemreinci, line 714:
```suggestion
                    auto i = index->fetch_add(1);
```

or

```suggestion
                    auto i = (*index)++;
```

Otherwise two separate threads can get the same index.

hazelcast/include/hazelcast/client/load_balancer.h
```diff
@@ -15,40 +15,71 @@
...
+                return std::move(*this);
+            }
 
             /**
-             * Returns the next member to route to
-             * @return Returns the next member or boost::none if no member is available
+             * Sets the function to be called when load balancer is initialized.
+             *
+             * @param cluster Cluster contains current membership information for initialization. And one can
+             * add membership through this class for future notifications.
```
yemreinci, line 62: Same doc with the init() method.

hazelcast/include/hazelcast/client/address.h
```diff
@@ -15,10 +15,8 @@
  */
 #pragma once
 
-#include "hazelcast/util/hazelcast_dll.h"
```
yemreinci, line 18: Isn't the `hazelcast_dll.h` header needed in this file? `HAZELCAST_API` is used down below.

yemreinci (changes requested):

hazelcast/include/hazelcast/client/address.h
ihsandemir: The next include `hazelcast/client/serialization/serialization.h` already includes it, hence less include lines.

hazelcast/include/hazelcast/client/load_balancer.h
ihsandemir: corrected.

hazelcast/include/hazelcast/client/address.h
yemreinci: https://softwareengineering.stackexchange.com/questions/262019/is-it-good-practice-to-rely-on-headers-being-included-transitively

ihsandemir: OK, put back the includes.

yemreinci (approved):

PR #734, opened 2020-12-14
Added a DNF fix for solving the docker build problem in earlier docker versions

ihsandemir:
Added a DNF fix for solving the docker build problem in earlier docker versions. Without this fix current nightly builds fail to build the docker image for 32-bit linux builds.


ihsandemir:
The problem came out to be a problem in docker not pulling the latest fedora (`docker pull fedora`) image at jenkins. When the image was pulled, the builds seem to start working fine. Hence, closing this PR.

PR #735, opened 2020-12-15
Make project compile with boost 1.74 or later

### LINKED ISSUE #732: Doesn't compile with Boost 1.74
The master branch does not compile with the latest Boost release: 1.74.
A similar compile error is emitted both on macos with clang and on linux with gcc.

With 1.74 the error is:
In file included from /Users/ihsan/Desktop/boost_74/include/boost/asio/query.hpp:22:
/Users/ihsan/Desktop/boost_74/include/boost/asio/traits/query_free.hpp:60:14: error: unexpected namespace name 'query': expected expression
    decltype(query(declval<T>(), declval<Property>()))

ihsandemir:
I had to make 2 changes to make newer boost versions compile:

1. The client::query namespace in asio  conflicts with query method in boost. Removed the `using namespace` statements from library headers to prevent name clashes.

2.  Changed the way that we use the socket strands. This is more portable and works with both old and new boost versions.

fixes #732

### DIFF
--- a/examples/Org.Website.Samples/QuerySample.cpp
+++ b/examples/Org.Website.Samples/QuerySample.cpp
@@ -64,7 +64,7 @@ void generate_users(std::shared_ptr<imap> users) {
 }
 
 int main() {
-    hazelcast_client hz;
+    hazelcast::client::hazelcast_client hz;
     // Get a Distributed Map called "users"
     auto users = hz.get_map("users").get();
     // Add some users to the Distributed Map
--- a/examples/Org.Website.Samples/QuerySample.cpp
+++ b/examples/Org.Website.Samples/QuerySample.cpp
@@ -64,7 +64,7 @@ void generate_users(std::shared_ptr<imap> users) {
 }
 
 int main() {
-    hazelcast::client::hazelcast_client hz;
+    hazelcast_client hz;
     // Get a Distributed Map called "users"
     auto users = hz.get_map("users").get();
     // Add some users to the Distributed Map
--- a/examples/replicated-map/entry-listener/ListeningReplicatedMap.cpp
+++ b/examples/replicated-map/entry-listener/ListeningReplicatedMap.cpp
@@ -32,22 +32,22 @@ int main() {
                     .on_updated([](hazelcast::client::entry_event &&event) {
                         std::cout << "[added] " << event << std::endl;
                     })
-            .on_evicted([](hazelcast::client::entry_event &&event) {
-                std::cout << "[updated] " << event << std::endl;
-            })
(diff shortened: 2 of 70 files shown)


ihsandemir:
verify-linux

yemreinci:
The reason why the query from Boost conflicts with our `hazelcast::client::query` is exactly this line: https://github.com/hazelcast/hazelcast-cpp-client/blob/036cae2363c56ce9357cecc883b7e4ba1149a09b/hazelcast/generated-sources/src/hazelcast/client/protocol/codec/codecs.h#L25

By using a `using namespace` we are exposing all of our symbols to the global namespace. We should never write a `using namespace` statement globally in an header file. The reason is explained here: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using-directive

Putting a prefix (like `hz_`) in front of symbols to avoid conflicts is the C way. C++ solves this problem with namespaces. As long as we keep all of our symbols under `hazelcast::`, we'll never have to change names to avoid conflicts.

sancar (approved):

examples/Org.Website.Samples/QuerySample.cpp
```diff
@@ -64,7 +64,7 @@ void generate_users(std::shared_ptr<imap> users) {
 }
 
 int main() {
-    hazelcast_client hz;
+    hazelcast::client::hazelcast_client hz;
```
yemreinci, line 67: There is `using namespace hazelcast::client` written at the top of this example.
I think we should either delete the `using namespace` statement or leave this line as it was.

There are other examples with the same thing.

yemreinci (changes requested):

examples/replicated-map/entry-listener/ListeningReplicatedMap.cpp
```diff
@@ -21,17 +21,17 @@ int main() {
...
+            hazelcast::client::entry_listener()
+                    .on_added([](hazelcast::client::entry_event &&event) {
+                        std::cout << "[added] " << event << std::endl;
+                    })
+                    .on_removed([](hazelcast::client::entry_event &&event) {
+                        std::cout << "[removed] " << event << std::endl;
+                    })
+                    .on_updated([](hazelcast::client::entry_event &&event) {
+                        std::cout << "[added] " << event << std::endl;
+                    })
```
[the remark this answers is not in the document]
ihsandemir: fixed

examples/Org.Website.Samples/QuerySample.cpp
ihsandemir: Reverted back the changes in examples.

yemreinci (approved):

PR #740, opened 2020-12-22
Fix for non-local use of connecting addresses and do NOT try connecting to old cluster members when client is in DISCONNECTED state

### LINKED ISSUE #738: Client reconnection takes very long in Kubernetes
Hazelcast Cpp client version: 4.0-BETA
Hazelcast server version: 4.1
Number of the clients: 1
Cluster size, i.e. the number of Hazelcast cluster members: 3

Client logs: [client-logs.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/5714935/client-logs.txt)

#### Expected behaviour
Client reconnects to the cluster shortly after the cluster is up again

#### Actual behaviour
It took a long time (30 min) for the client to reconnect to the cluster.

1. Cluster of 3 members is deployed in Kubernetes (resolved with DNS `hz-hazelcast`)

```
helm repo add hazelcast https://hazelcast-charts.s3.amazonaws.com/
helm repo update
helm install hz-hazelcast hazelcast/hazelcast
```

2. The client code is available [here](https://github.com/leszko/kubernetes/blob/cpp/cpp/).

```
#define BOOST_THREAD_VERSION 5

#include <iostream>
#include <hazelcast/client/hazelcast_client.h>
#include <thread>
#include <chrono>

int main() {
    hazelcast::client::client_config config;
    config.get_network_config().add_address(address("hz-hazelcast", 5701));
    config.get_network_config().set_connection_attempt_limit(INT32_MAX);
    hazelcast::client::hazelcast_client hz(std::move(config));

    std::cout << "Successful connection!" << std::endl;
    std::cout << "Starting to fill the map with random entries." << std::endl;

    auto map = hz.get_map("map").get();
    while (true) {
        int random_key = rand() % 100000;
        try {
            map->put("key-" + std::to_string(random_key), "value-" + std::to_string(random_key));
            if (random_key % 100 == 0) {
                std::cout << "Current map size: " + std::to_string(map->size().get()) << std::endl;
                std::this_thread::sleep_for(std::chrono::seconds(1));
            }
        } catch (const std::exception& e) {
            std::cout << e.what() << std::endl;
        }
    }

    return 0;
}
```

3, Client connects successfully to the cluster and start inserting data.
4. Terminate and start again all Hazelcast members (they start with different IP addresses)

```
kubectl delete pod hz-hazelcast-0 hz-hazelcast-1 hz-hazelcast-2
```

5. The client reconnects to the cluster but after a very long time. Please find the logs attached.

ihsandemir:
Fixed the pending connection list usage.

Added a check so that if there is no cluster connection we do not try opening connections to members list. Failing to do this causes an accumulation of scheduled connect requests to all members in the old member list for every second by the `connect to all members` executor threads (frequency of the reconnect test). 

Changed the place where the connection log is written so that we can see all the connection attempts including the periodic member connection tries.

The fix is tried with kubernetes and it works fine with cluster reconnects.

fixes #738


leszko (approved):
LGTM 👍 

Thanks for the fix @ihsandemir

mdumandag (approved):

leszko (approved):

PR #744, opened 2020-12-30
Client Exponential Connection Retry

### LINKED ISSUE #741: ConnectionRetryConfig feature
Java is using `ConnectionRetryConfig` (https://github.com/hazelcast/hazelcast/commit/6d3c6c33342d2435a55850c031b833897fa8dc6f) for connection retries. We may switch to this config to be compatible with Java.

ihsandemir:
Removed the **cluster connect attempt limit** and *cluster connect attempt period* configurations and added the client strategy config `connection_retry_config`. This makes the client more compatible with Java client configuration and it also allows exponential backoff algorithm. (back-ported https://github.com/hazelcast/hazelcast/pull/13248 and https://github.com/hazelcast/hazelcast/pull/15923)

Also back-ported, https://github.com/hazelcast/hazelcast/pull/17844

fixes #741

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,15 +37,14 @@
   * [5.1. Providing Member Addresses](#51-providing-member-addresses)
   * [5.2. Setting Smart Routing](#52-setting-smart-routing)
   * [5.3. Enabling Redo Operation](#53-enabling-redo-operation)
-  * [5.4. Setting Connection Timeout](#54-setting-connection-timeout)
-  * [5.5. Setting Connection Attempt Limit](#55-setting-connection-attempt-limit)
-  * [5.6. Setting Connection Attempt Period](#56-setting-connection-attempt-period)
-  * [5.7. Enabling Client TLS/SSL](#57-enabling-client-tlsssl)
-  * [5.8. Enabling Hazelcast AWS Cloud Discovery](#58-enabling-hazelcast-aws-cloud-discovery)
-  * [5.9. Authentication](#59-authentication)
-    * [5.9.1. Username Password Authentication](#591-authentication)
-    * [5.9.2. Token Authentication](#591-authentication)
-  * [5.10. Configuring Backup Acknowledgment](#59-configuring-backup-acknowledgment)    
+  * [5.4. Setting Cluster Connection Timeout](#54-setting-cluster-connection-timeout)
+  * [5.5. Advanced Cluster Connection Retry Configuration](#55-advanced-cluster-connection-retry-configuration)
+  * [5.6. Enabling Client TLS/SSL](#56-enabling-client-tlsssl)
+  * [5.7. Enabling Hazelcast AWS Cloud Discovery](#57-enabling-hazelcast-aws-cloud-discovery)
+  * [5.8. Authentication](#58-authentication)
+    * [5.8.1. Username Password Authentication](#581-authentication)
+    * [5.8.2. Token Authentication](#581-authentication)
+  * [5.9. Configuring Backup Acknowledgment](#59-configuring-backup-acknowledgment)    
 * [6. Securing Client Connection](#6-securing-client-connection)
   * [6.1. TLS/SSL](#61-tlsssl)
     * [6.1.1. TLS/SSL for Hazelcast Members](#611-tlsssl-for-hazelcast-members)
@@ -976,9 +975,8 @@ Here is an example of configuring the network for C++ Client programmatically.
     clientConfig.get_network_config().add_addresses({{"10.1.1.21", 5701}, {"10.1.1.22", 5703}});
     clientConfig.get_network_config().set_smart_routing(true);
     clientConfig.set_redo_operation(true);
-    clientConfig.get_network_config().set_connection_timeout(6000);
-    clientConfig.get_network_config().set_connection_attempt_period(5000);
-    clientConfig.get_network_config().set_connection_attempt_limit(5);
+    clientConfig.get_connection_strategy_config().get_retry_config().set_cluster_connect_timeout(
+            std::chrono::seconds(30));
 ```
 
 ## 5.1. Providing Member Addresses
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -1019,54 +1017,44 @@ It enables/disables redo-able operations. While sending the requests to the rela
...
 
-## 5.4. Setting Connection Timeout
-
-Connection timeout is the timeout value in milliseconds for the members to accept the client connection requests.
-If the member does not respond within the timeout, the client will retry to connect as many as `client_network_config::getConnectionAttemptLimit()` times.
+## 5.4. Setting Cluster Connection Timeout
+Cluster connection timeout is the timeout value for which the client tries to connect to the cluster. If the client can not connect to cluster during this timeout duration, the client shuts down itself and it can not be re-used (you need to obtain a new client).
  
-The following is an example configuration.
+The following example shows how yopu can set the cluster connection timeout to 30 seconds.
```
yemreinci, line 1023:
```suggestion
The following example shows how you can set the cluster connection timeout to 30 seconds.
```

sancar (approved):

hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
```diff
@@ -204,22 +216,27 @@ namespace hazelcast {
...
                 boost::uuids::uuid client_uuid_;
                 boost::chrono::milliseconds authentication_timeout_;
+                std::vector<std::string> labels_;
+                load_balancer &load_balancer_;
+                wait_strategy wait_strategy_;
+
+                // following fields are updated inside synchronized(clientStateMutex)
+                std::mutex client_state_mutex_;
+                util::SynchronizedMap<boost::uuids::uuid, Connection, boost::hash<boost::uuids::uuid>> active_connections_;
+                util::SynchronizedMap<int32_t, Connection> active_connection_ids_;
```
yemreinci, line 232: Now that the accesses to `active_connections_` and `active_connection_ids_` are protected with the mutex `client_state_mutex_`, can they be regular `unordered_map`s instead of `util::SynchronizedMap`?

hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -160,16 +160,75 @@ namespace hazelcast {
...
+                                                        "prefetchValidityMs must be non negative");
                 prefetch_validity_duration_ = duration;
                 return *this;
             }
 
-            int32_t client_network_config::CONNECTION_ATTEMPT_PERIOD = 3000;
+            constexpr std::chrono::milliseconds connection_retry_config::INITIAL_BACKOFF;
+            constexpr std::chrono::milliseconds connection_retry_config::MAX_BACKOFF;
+            constexpr std::chrono::milliseconds connection_retry_config::CLUSTER_CONNECT_TIMEOUT;
+            constexpr double connection_retry_config::JITTER;
```
yemreinci, line 171: These aren't necessary for static constexpr variables.
```suggestion
```

hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -278,11 +311,22 @@ namespace hazelcast {
                 return *this;
             }
 
+            connection_retry_config &client_connection_strategy_config::get_retry_config() {
+                return retry_config_;
+            }
+
+            client_connection_strategy_config &
+            client_connection_strategy_config::set_retry_config(const connection_retry_config &retry_config) {
+                retry_config_ = retry_config;
```
yemreinci, line 320: You can take by value, and then move:
```suggestion
            client_connection_strategy_config::set_retry_config(connection_retry_config retry_config) {
                retry_config_ = std::move(retry_config);
```

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -440,51 +455,51 @@ namespace hazelcast {
...
+                        }
 
-                    // If the address providers load no addresses (which seems to be possible), then the above loop is not entered
+                        auto connection = connect<address>(server_address, [=](const address &target) {
+                            return get_or_connect_to_address(target);
+                        });
+                        if (connection) {
+                            return true;
+                        }
+                    }
```
yemreinci, line 493: The method `ClientConectionManagerImpl::connect` doesn't actually connect anywhere. It just wraps the given function call with a try-catch and prints a log statement if the function throws an exception. I believe it's just there to avoid duplicating the same logic for these two for-loops. But still, these two for loops share the same piece of code.

I think I have a better solution. First, make `get_or_connect_to_address` and `get_or_connect_to_member` specialization of the same function template, like `get_or_connect<address>` and `get_or_connect<member>`. Then, write another method for the logic in the for loop:

```C++
template<typename T>
shared_ptr<Connection> try_connect(T target) {
   check_client_active()
   if(!tried_addresses_per_attempt.insert(server_address).second)
     return false;
   // same as in connect<>:
   try {
      return get_or_connect<T>(target)
   } except (...) {
      HZ_LOG(...)
   }
}

```

Now we can convert the for loops into:

```C++
for (auto m: member_list) {
   auto conn = try_connect(m);
   if (conn) return true;
}

for (auto a: addresses) {
   auto conn = try_connect(a);
   if (conn) return true;
}
```

yemreinci (changes requested):

hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
ihsandemir: There are still methods such as `get_connection(boost::uuids::uuid uuid)` and `get_connection(const address &address)` which do not use the `client_state_mutex_`, hence we can not completely use regular maps.

hazelcast/src/hazelcast/client/config.cpp
ihsandemir: Unfortunately it does not compile that way, the errors are like:
```
Undefined symbols for architecture x86_64:
  "hazelcast::client::config::connection_retry_config::MAX_BACKOFF", referenced from:
      hazelcast::client::config::connection_retry_config::connection_retry_config() in config.cpp.o
```

hazelcast/include/hazelcast/client/config/client_connection_strategy_config.h
```diff
@@ -84,10 +85,27 @@ namespace hazelcast {
...
+                connection_retry_config &get_retry_config();
+
+                /**
+                 * connection_retry_config controls the period among the retries and when a client should give up
+                 * retrying. Exponential behaviour can be chosen or jitter can be added to wait periods.
+                 *
+                 * @param connection_retry_config the connection retry config to read or modify.
+                 * @return the updated connection_retry_config
+                 */
+                client_connection_strategy_config &set_retry_config(const connection_retry_config &retry_config);
```
ihsandemir, line 103:
```suggestion
                client_connection_strategy_config &set_retry_config(connection_retry_config retry_config);
```

hazelcast/src/hazelcast/client/network.cpp
ihsandemir: Changed.

hazelcast/src/hazelcast/client/config.cpp
yemreinci: Could it be because the header `connection_retry_config.h` was not included in `config.cpp`?

yemreinci: OK, they are necessary. They are not for C++17 though, that is probably the reason I thought so.
Relevant: https://stackoverflow.com/a/45020525

yemreinci (changes requested):

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -380,19 +393,21 @@ namespace hazelcast {
             }
 
             void ClientConnectionManagerImpl::submit_connect_to_cluster_task() {
-                bool expected = false;
-                if (!connect_to_cluster_task_submitted_.compare_exchange_strong(expected, true)) {
+                // called in std::lock_guard<std::mutex>(client_state_mutex_)
+                if (connect_to_cluster_task_submitted_) {
```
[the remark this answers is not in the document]
ihsandemir: good catch. reverted back the place where we set it to true.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -204,17 +192,34 @@ namespace hazelcast {
...
+                connection = std::make_shared<Connection>(target, client_, ++connection_id_gen_,
+                                                          *socket_factory_, *this, connection_timeout_millis_);
+                connection->connect();
+
+                // call the interceptor from user thread
+                socket_interceptor_.connect_(connection->get_socket());
+
+                auto result = authenticate_on_cluster(connection);
+
+                return on_authenticated(connection, result);
```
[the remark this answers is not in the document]
ihsandemir: refactored.

yemreinci (approved):

ihsandemir:
verify-windows

PR #746, opened 2021-01-05
Remove file name and line number in log line and the logger function API

### LINKED ISSUE #737: Absolute file path in log statements may mislead users
Current log line looks like this: 

```
17/12/2020 12:03:52.408 INFO: [140346571843328] hz.client_1[dev] [4.0-BETA] [/hazelcast-cpp-client/hazelcast/src/hazelcast/client/network.cpp:589] Trying to connect to Address[10.216.7.4:5701]
```

The file path `/hazelcast-cpp-client/hazelcast/src/hazelcast/client/network.cpp` may not correspond to a valid path for the user of the binary library. If the user builds the library in the same environment as the application environment, then it may be a valid path but otherwise it only exists at the machine where the library was built. Therefore, we may find a better solution for this. It is very beneficial to go to the exact code line for a problem, that is very helpful in debugging.

An alternative solution: Use method name with the full namespace + line number. This can be done in a portable way with the help of boost:

```
BOOST_CURRENT_FUNCTION:__LINE__
```

ihsandemir:
Removed file name and log line number from the log lines.

fixes #737

### DIFF
--- a/hazelcast/test/src/HazelcastTests2.cpp
+++ b/hazelcast/test/src/HazelcastTests2.cpp
@@ -468,15 +468,13 @@ namespace hazelcast {
 
                 auto h = [&called](const std::string &,
                                    const std::string &,
-                                   const char*,
-                                   int,
                                    logger::level,
                                    const std::string &) {
                     called = true;
                 };
 
                 logger_config.handler(h);
-                logger_config.handler()("", "", "", 0, logger::level::fine, "");
+                logger_config.handler()("", 0, logger::level::fine, "");
 
                 ASSERT_TRUE(called);
             }
--- a/hazelcast/test/src/HazelcastTests2.cpp
+++ b/hazelcast/test/src/HazelcastTests2.cpp
@@ -460,7 +460,7 @@ namespace hazelcast {
                 ASSERT_THROW(config.get_logger_config().handler(nullptr), exception::illegal_argument);
             }
 
-            TEST_F(ClientConfigTest, testSetGetLoadHandler) {
+            TEST_F(ClientConfigTest, testSetGetLogHandler) {
                 client_config config;
                 auto logger_config = config.get_logger_config();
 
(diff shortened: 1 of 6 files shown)


hazelcast/test/src/logger_test.cpp
```diff
@@ -115,15 +107,14 @@ class default_log_handler_test : public ::testing::Test {
...
-    char lev[64], tid[64], msg[64], ins_grp[64], ver[64], file_line[256];
+    char lev[64], tid[64], msg[64], ins_grp[64], ver[64];
 
     int read = std::sscanf(sstrm_.str().c_str(), 
-        "%02d/%02d/%04d %02d:%02d:%02d.%03d %s %s %s %s %s %s\n", 
-         &day, &mon, &year, &hr, &mn, &sec, &ms, lev, tid, ins_grp, ver, file_line, msg);
+        "%02d/%02d/%04d %02d:%02d:%02d.%03d %s %s %s %s %s\n",
+         &day, &mon, &year, &hr, &mn, &sec, &ms, lev, tid, ins_grp, ver, msg);
 
     ASSERT_EQ(13, read);
```
yemreinci, line 119:
```suggestion
    ASSERT_EQ(12, read);
```

hazelcast/test/src/HazelcastTests2.cpp
```diff
@@ -468,15 +468,13 @@ namespace hazelcast {
...
-                                   const char*,
-                                   int,
                                    logger::level,
                                    const std::string &) {
                     called = true;
                 };
 
                 logger_config.handler(h);
-                logger_config.handler()("", "", "", 0, logger::level::fine, "");
+                logger_config.handler()("", 0, logger::level::fine, "");
```
yemreinci, line 477:
```suggestion
                logger_config.handler()("", "", logger::level::fine, "");
```


Can you also change the name of this test? I realized I made a typo there, it should have been `testSetGetLogHandler` but I somehow wrote `testSetGetLoadHandler`.

yemreinci (changes requested):

yemreinci (approved):

PR #754, opened 2021-01-08
Separate client construction and start - client start returns async future

ihsandemir:
Separated client construction and start. The client start returns async future.

Made the lifecycle_service start and stop which are user api, return future.
    
Renamed client `shutdown` to `stop`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1046,10 +1046,13 @@ The following example shows how you can set the cluster connection timeout to 30
 
 ## 5.5. Advanced Cluster Connection Retry Configuration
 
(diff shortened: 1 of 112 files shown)


ihsandemir:
verify-windows

ihsandemir:
verify-linux

sancar:
I would prefer to have a static newHazelcastClient method to return an async future if we have to. 
A separate start method really feels like the instance can be started after it is shutdown. 
Even if we can, it is a completely new client. They will not share any state. So it makes me feel better when we make it obvious on the API. 
And I think, the name SHUTDOWN should stay as it is the name of a Lifecycle Event. And it feels better to have this common across all clients..

ihsandemir:
i decided not to give this API for now but provide `future<void> new_client(client_config &&)` instead and make the  `hazelcast_client` constructor as private.

PR #755, opened 2021-01-09
Change `boost::launch::deferred` continuations to `boost::launch::async` 

### LINKED ISSUE #753: Program hangs when async launch policy is used with future continuations
This program hangs indefinitely:

```C++
#include <hazelcast/client/hazelcast_client.h>

int main(int argc, char* argv[]) {
    hazelcast::client::hazelcast_client hz;

    auto map = hz.get_map("my_map").get();
    
    map->get<int, int>(1)
        .then(boost::launch::async, [](boost::future<boost::optional<int>> res_future) {
            auto res = res_future.get();
            std::cout << (res ? res.get() : -1) << std::endl;
        })
        .get();

    return 0;
}
```

It is because we attach continuations with the policy `boost::launch::deferred` to the returned future object. For example, the [decode](https://github.com/hazelcast/hazelcast-cpp-client/blob/7caeca51e0ccc7ee3c5a9238bb2464d255e1f163/hazelcast/include/hazelcast/client/proxy/SerializingProxy.h#L242-L247) function template attaches a deferred continuation.

The program will hang if an async continuation is attached after a deferred one:
```C++
#include <iostream>

#define BOOST_THREAD_VERSION 5
#include <boost/thread.hpp>

int main() {

    boost::async([](){
        std::cout << "will be printed" << std::endl;
    }).then(boost::launch::deferred, [](boost::future<void> r){
        std::cout << "wont be printed" << std::endl;
    }).then( boost::launch::async, [](boost::future<void> r){
        std::cout << "wont be printed" << std::endl;
    }).get();

}
```

This problem probably exists for all the functions that return `boost::future` (because all of them call `decode` at some point). This means the user can only use continuations with a `deferred` launch policy.

ihsandemir:
Changed `boost::launch::deferred` continuations to `boost::launch::sync` except for the `ClientInvocation::invoke` and `ClientInvocation::invoke_urgent` methods, since these methods are being executed by the io thread which we do not want to block. This way, the io thread receives the message and composes the client message for delivery and the async thread does the message decoding and any continuations in a sync manner (a boost::async call for each client invocation result delivery to the user).

Current thread structure for an example imap::get call:

io_thread prepares the client_message -> a new thread is spawn to process the rest of the delivery of the message and other processing using boost::launch::async continuation. The rest of the processing occurs on this spawned thread, hence all the rest future continuations work in boost::launch::sync way.

Also added a new test `testIssue753 ` to test the problem.

fixes #753

### DIFF
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -124,7 +124,7 @@ namespace hazelcast {
                         client_proxy->get_service_name());
                 return spi::impl::ClientInvocation::create(client_, clientMessage,
                                                            client_proxy->get_service_name())->invoke().then(
-                        boost::launch::async, [=](boost::future<protocol::ClientMessage> f) {
+                        boost::launch::sync, [=](boost::future<protocol::ClientMessage> f) {
                             f.get();
                             client_proxy->on_initialize();
                         });
@@ -523,7 +523,7 @@ namespace hazelcast {
                         get_name(), get_service_name());
                 return spi::impl::ClientInvocation::create(get_context(), std::make_shared<protocol::ClientMessage>(
                         std::move(clientMessage)), get_name())->invoke().then(
-                        boost::launch::async, [](boost::future<protocol::ClientMessage> f) { f.get(); });
+                        boost::launch::sync, [](boost::future<protocol::ClientMessage> f) { f.get(); });
             }
 
             boost::future<boost::uuids::uuid>
@@ -1054,7 +1054,6 @@ namespace hazelcast {
                                                    boost::uuids::uuid uuid) :
                         logger_(client_context.get_logger()),
                         lifecycle_service_(client_context.get_lifecycle_service()),
-                        client_cluster_service_(client_context.get_client_cluster_service()),
                         invocation_service_(client_context.get_invocation_service()),
                         execution_service_(client_context.get_client_execution_service().shared_from_this()),
                         call_id_sequence_(client_context.get_call_id_sequence()),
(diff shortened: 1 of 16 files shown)


ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify-windows

ihsandemir:
verify

ihsandemir:
verify-linux

ihsandemir:
verify-linux

sancar (dismissed):

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1089,11 +1090,13 @@ namespace hazelcast {
                     // for back pressure
                     call_id_sequence_->force_next();
                     invoke_on_selection();
-                    return invocation_promise_.get_future().then(boost::launch::sync,
-                                                               [=](boost::future<protocol::ClientMessage> f) {
-                                                                   call_id_sequence_->complete();
-                                                                   return f.get();
-                                                               });
+                    hazelcast::util::hz_thread_pool p(1);
```
yemreinci, line 1093: Is this being used?

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1076,11 +1076,12 @@ namespace hazelcast {
                     // for back pressure
                     call_id_sequence_->next();
                     invoke_on_selection();
-                    return invocation_promise_.get_future().then(boost::launch::sync,
-                                                               [=](boost::future<protocol::ClientMessage> f) {
-                                                                   call_id_sequence_->complete();
-                                                                   return f.get();
-                                                               });
+                    auto id_seq = call_id_sequence_;
+                    return invocation_promise_.get_future().then(boost::launch::async,
```
yemreinci, line 1080: If I'm not mistaken, a new thread will be created for every invocation. Right?

ihsandemir: yes, that is true for most implementations.

hazelcast/src/hazelcast/client/spi.cpp
ihsandemir: this was forgotten to be deleted. good catch, deleted it.

sancar (approved):

yemreinci (approved):

PR #759, opened 2021-01-15
Fix boost version in cmake pkg config and fix multi config check

### LINKED ISSUE #752: Don't repeat required Boost version
We have recently updated the required Boost version from `1.72` to `1.71`. 
But the cmake config template still has `1.72` [here](https://github.com/hazelcast/hazelcast-cpp-client/blob/master/cmake/config.cmake.in).

Use a variable for the Boost version to avoid repetition.

yemreinci:
* The boost version in the cmake package config template is updated to 1.71.
* `GENERATOR_IS_MULTI_CONFIG` isn't a variable it's a global property, fixed the if condition to use it correctly.
* `Boost::headers` is replaced with `Boost::boost`. Because the `headers` target was added in a newer CMake version and it doesn't exists in 3.10.
* find_dependency macro in the package config file automatically infers the `REQUIRED` and `QUIET` keywords from the calling find_package command. So, the `REQUIRED` parameter is removed from the config.cmake.in file.

Fixes #752. Well the issue said not to repeat the version but on second thought, it's not something that would change that frequently and I guess an amount of repetition between the main CMakeLists.txt and the config.cmake.in is inevitable. We have to be careful when we update our package dependencies and keep the two files in sync.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,7 @@ endif ()
 find_package(Threads REQUIRED)
 
 # find Boost
-find_package(Boost 1.71 REQUIRED COMPONENTS headers thread chrono)
+find_package(Boost 1.71 REQUIRED COMPONENTS thread chrono)
 
 # find OpenSSL if building WITH_OPENSSL
 if (WITH_OPENSSL)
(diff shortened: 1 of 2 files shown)


yemreinci:
verify-windows

ihsandemir (approved):

PR #761, opened 2021-01-18
Update license year from 2020 to 2021

ihsandemir:
Updated license year from 2020 to 2021.


mdumandag (approved):
We should also update the template in the protocol repo

PR #807, opened 2021-02-03
Updates to testing instructions

yuce:


scripts/start-rc.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
```
ihsandemir, line 1: why?

yuce: `function ...` syntax only works in Bash, we have to remove `function` to make it compatible with the Bourne shell. I didn't check for other Bash-isms in the script, there could be more.

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
-* Maven
-* cmake
-* OpenSSL
-* Boost
-* thrift
+* Maven 3 or newer (for server)
+* cmake 3.10 or newer
+* OpenSSL 1.0.2 or newer
+* Boost >= 1.71
+* thrift 0.13
```
ihsandemir, line 3171: You may use a different version of thrift but this is the one tested with, hence, I am not sure about making such a bold claim.

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
+* Bash (on platforms other than Windows)
+* Hazelcast Enterprise license
```
ihsandemir, line 3173: This may mislead the user as if we are not open-source, we need to be clear about it. We better say this is only for starting the Java server.

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
+set max files limit ulimit -n
+
+You need to enable tests in the build with the cmake flag `-DBUILD_TESTS=ON`. When you build, the result will
 produce the executable `client_test` on Linux/macOS and `client_test.exe` on Windows. You need to start the remote
 controller first before running this executable by executing `scripts/start-rc.sh` on Linux/macOS
 or `scripts/start-rc.bat` on Windows. Then you can run the test executable.
 
+### 8.1.1 Testing on Linux/MacOS
+
+1. First, make sure you have all the requirements are met.
```
ihsandemir, line 3184:
```suggestion
1. First, make sure you have all the requirements satisfied.
```

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
+2. Make sure open files limit is set to a reasonably high value in all the terminals you run the test and Hazelcast Remote Control.
```
ihsandemir, line 3185: Are we sure about this? Did we test this throughly and understand the reason for it?

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
+   `10000` seems to be a good value. On Linux, you can use `ulimit -n 10000`.
+3. Start Hazelcast Remote Control.
```
ihsandemir, line 3187:
```suggestion
3. Start Hazelcast Remote Controller. You can run script `scripts/start-rc.sh` on Linux/MacOS and `scripts\start-rc.bat` on Windows.
```

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
+   Note that, currently we require Hazelcast Enterprise for the tests, so `HAZELCAST_ENTERPRISE_KEY` environment key should exist with an invalid key.
+   ```
+   HAZELCAST_ENTERPRISE_KEY=... ./scripts/start-rc.sh&
+   ```
+4. Create the build directory for testing and switch to that directory:
+   ```
+   mkdir build-test && cd $_
```
ihsandemir, line 3194:
```suggestion
   mkdir build && cd build
```

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
+   ```
+5. Run `cmake` with `-DBUILD_TESTS=ON` and any other options specific to your build:
+   ```
+   cmake -DBUILD_TESTS=ON ..  
+   ```
+6. Run `make` to create the test executable. Built test executable is at `${PROJECT_DIR}/built-test/hazelcast/test/src/client_test`.
```
ihsandemir, line 3200:
```suggestion
6. Run `make client_test` to create the test executable. Built test executable is at `${PROJECT_DIR}/build/hazelcast/test/src/client_test`.
```

ihsandemir, line 3200: Instruction only works for Linux/MacOS. What about windows?

README.md
yuce: v0.13 is the latest release of Thrift and it was released in Oct 2019. Pinning the version to 0.13 seems to be a pretty safe bet to me. I can include any other version you've tried.

README.md
yuce: ok

README.md
yuce: Thanks, I think I've mixed two different thoughts I had. I'll replace it with yours.

README.md
yuce: Since `build` is frequently used as the directory name for the library itself (not tests) I think `build-test` is more appropriate here. `$_` evaluates to the last argument so it saves a potential mistake.

README.md
yuce: The default target for the `Makefile` generated by cmake is `all`. Is that the wrong target?

yuce: I guess I thought this would be the first step in improving the testing section. I'll add the instructions for Windows.

README.md
yuce: Thanks, I missed the actual command.

README.md
yuce: I've encountered another error without raising the open file limit; epoll runs out of file descriptors. Raising the open files limit seems to solve both issues.

README.md
yuce: I was wrong, I didn't missed the command. Since this is the Linux/MacOS testing section, I didn't include the Windows command.

README.md
ihsandemir: can you reproduce the open file limit problem? I thought that this was not a problem with asio using the appropriate reactor. I know that this was a problem at 3.12 release but I am not expecting 4.0 release having this problem. If there is such a problem, I like to invest time on this.

README.md
```diff
@@ -3164,17 +3164,46 @@ tweak the implementation to your application's needs, you can follow the steps i
...
-* OpenSSL
-* Boost
-* thrift
+* Maven 3 or newer (for server)
+* cmake 3.10 or newer
+* OpenSSL 1.0.2 or newer
+* Boost >= 1.71
+* thrift 0.13
+* Bash (on platforms other than Windows)
+* Hazelcast Enterprise license (required only for running the tests)
```
ihsandemir, line 3173:
```suggestion
* For starting Enterprise Hazelcast IMDG Server: Hazelcast Enterprise license
```

README.md
ihsandemir: Will you send it in another PR or in another commit?

ihsandemir (commented):
Looks good overall.

README.md
yuce: I guess separate PR for Windows would work better.

yemreinci: I would suggest not mentioning `make` at all and using `cmake --build`.

PR #808, opened 2021-02-05
Cleanup CMake files for tests

yemreinci:
This PR addresses these issues:

1. `find_package(GTest)` didn't work because it defines the target `GTest::gtest` but we link against `gtest`. Installing googletest with `brew install googletest` breaks the build. Solved the problem by setting a variable named `GTEST_TARGET` depending on whether find_package is successful.

2. CMakeLists.txt files were using 2 features that were introduced after version 3.10: the TYPE option for `install(DIRECTORY)` and the HOMEPAGE_URL option for `project()`. They are removed. 
Alternatively, we can increase the min required CMake version. HOMEPAGE_URL was introduced in 3.12 and TYPE was introduced in 3.14. However, I prefer keeping the 3.10 as some major linux distros still come wıth fairly old CMake versions. (e.g. Ubuntu 18 comes with 3.10 and will be supported until 2023)

3. Linking to the Thrift library was not portable. Hard-coded paths didn't allow for a custom installation directory. A FindThrift.cmake module is added. Users can now explicitly set cache variables `Thrift_LIBRARY` and `Thrift_INCLUDE_DIR` or add the installation prefix in `CMAKE_PREFIX_PATH` if thrift was not found. 

Three additional small fixes:
* Order #include directives in test files using quotes ("") for local headers, and angle brackets (<>) for system headers, remove duplicates.
* Fix class and method name in a precondition check.
* I ran into [this](https://github.com/yemreinci/hazelcast-cpp-client/runs/1991737312?check_suite_focus=true) problem in a GA run. netstat outputs `udp4       0      0  *.59701                *.*`, making the test-unix.sh think the RC was started. I made the grep pattern stronger to always match a listening socket bound to port 9701.


yuce:
@yemreinci  Would it be possible to add an option to override Thrift include/lib? It's possible to override Boost home dir using `-DBOOST_ROOT` but something like that is not available for Thrift in the default cmake install.

yemreinci:
> @yemreinci Would it be possible to add an option to override Thrift include/lib? It's possible to override Boost home dir using `-DBOOST_ROOT` but something like that is not available for Thrift in the default cmake install.

Ideally, a Thrift installation should come with package config files for CMake, and we should be able to find it using `find_package`. And if a user wants to use a different home path for Thrift, they should be able to do so by setting the variable `Thrift_ROOT`. (`<PackageName>_ROOT` is a special variable that cmake looks when `find_package` is called. [See](https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html#variable:<PackageName>_ROOT)).

Current situation is that Thrift doesn't support CMake for users. Meaning, when you install it from source or download from a system package manager (like apt, yum, etc.) it doesn't come with package config files and therefore won't be found by `find_package`.  C++ package managers like Conan and Vcpkg remedy that by providing the pkg config file themselves. 

So, there is a `Thrift_ROOT` but it is only usable if you installed Thrift using Conan or vcpkg. 

For the general case, we can add and maintain a `FindThrift.cmake` module to the project (like it was done [here](https://github.com/apache/parquet-cpp/blob/master/cmake_modules/FindThrift.cmake)) but I am not sure how necessary that is. @ihsandemir what would you think?

@yuce BTW, you can still pass `-DCMAKE_PREFIX_PATH=/my/install/path/` if you installed from source into the prefix `/my/install/path` and it should work.

yuce:
> For the general case, we can add and maintain a `FindThrift.cmake` module to the project (like it was done [here](https://github.com/apache/parquet-cpp/blob/master/cmake_modules/FindThrift.cmake)) but I am not sure how necessary that is. @ihsandemir what would you think?
> 
> @yuce BTW, you can still pass `-DCMAKE_PREFIX_PATH=/my/install/path/` if you installed from source into the prefix `/my/install/path` and it should work.

Thanks for considering this change.

Ubuntu 18.04 official repo doesn't include `libthrift-dev` and it has a very long life time (2028! https://ubuntu.com/about/release-cycle) so it's not irrelevant. If some library is not provided in the official repos the first thing I try is to compile it, not get it from unofficial repos, etc. (probably I am not the only one).

Actually I tried to use a `FindThrift.cmake` module before, and it didn't work: It found the include dir but not the lib. So, what I currently do is manually replace include and lib dirs of Thrift at `hazelcast/test/CMakeLists.txt`. I think being able override Thrift include/lib dirs with something like `-DTHRIFT_INCLUDE`, etc. would be both beneficial and the implementation would be trivial. But updating `CMakeLists.txt` with this change is probably out of context of this PR. If so, I can send a separate PR.

I didn't try to install Thrift using cmake, just the `./configure` script. But I guess using cmake would probably help only if I installed it to `/usr`?

yemreinci:
@yuce The suggestion I made about setting `CMAKE_PREFIX_PATH` was wrong, I deleted that. But I believe you could still get it to work without updating the `CMakeLists.txt` files. Explicitly setting compiler flags is an option for example. (like `-DCMAKE_CXX_FLAGS="-I /path/to/thrift/include/"`). 

Anyways, as you expressed, the current user experience isn't smooth and can definitely be improved. The purpose of this PR is just to clean up the code for the current logic and fix some bugs. I believe sending your suggestion as a separate PR or opening an issue would be more appropriate.

hazelcast/test/CMakeLists.txt
```diff
@@ -1,38 +1,43 @@
...
-set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
-
 if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     #required to suppress the warning:
     # `gtest-death-test.cc:1301:24: error: 'dummy' may be used uninitialized [-Werror=maybe-uninitialized]`
     #for googletest source compilation
     set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wno-maybe-uninitialized")
 endif ()
 
-find_package(GTest QUIET)
```
ihsandemir, line 12: What about using the same method as you used for thrift, try find and use the found one or otherwise download. You may be able to solve the proper lib naming it with a variable.

CMakeLists.txt
```diff
@@ -19,7 +19,6 @@ cmake_minimum_required(VERSION 3.10)
 project(hazelcast-cpp-client
         VERSION 4.0.1
         DESCRIPTION "Hazelcast C++ Client"
-        HOMEPAGE_URL https://hazelcast.org/clients/cplusplus/
```
ihsandemir, line 22: why?

CMakeLists.txt
```diff
@@ -33,10 +32,12 @@ if ((NOT is_multi_config) AND (NOT CMAKE_BUILD_TYPE))
     set(CMAKE_BUILD_TYPE Release)
 endif ()
 
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules/")
```
ihsandemir, line 35: This can also be set from outside on cmake command. We should not overwrite and lose that.

CMakeLists.txt
```diff
@@ -243,7 +244,7 @@ install(
         hazelcast/generated-sources/src/
         ${CMAKE_CURRENT_BINARY_DIR}/include/
 
-    TYPE INCLUDE
```
ihsandemir, line 246: I could not understand what is this change for. They seem to be the same directory default but letting it be determined automatically sounds more portable.

hazelcast/test/CMakeLists.txt
```diff
@@ -1,38 +1,50 @@
...
-if (NOT GTest_FOUND)
-    message(STATUS "No existing googletest library installation is found. Will now add googletest as external project by downloading it from github.")
+find_package(GTest 1.10)
```
ihsandemir, line 10: did you test the min. supported version?

cmake/Modules/FindThrift.cmake
```diff
@@ -0,0 +1,100 @@
...
+
+#]=======================================================================]
+
+find_path(
+    Thrift_INCLUDE_DIR
+    NAMES thrift/Thrift.h
+    PATH_SUFFIXES include
+)
+
+find_library(
```
ihsandemir, line 61: should you put this after `find_package_handle_standard_args `?

cmake/Modules/FindThrift.cmake
```diff
@@ -0,0 +1,100 @@
...
+if(Thrift_FOUND AND NOT TARGET Thrift::thrift)
+    add_library(Thrift::thrift UNKNOWN IMPORTED)
+    set_target_properties(
+        Thrift::thrift PROPERTIES
+        IMPORTED_LOCATION "${Thrift_LIBRARY}"
+        INTERFACE_INCLUDE_DIRECTORIES "${Thrift_INCLUDE_DIR}"
+    )
+endif()
+
+mark_as_advanced(
```
ihsandemir, line 95: why?

cmake/Modules/FindThrift.cmake
```diff
@@ -0,0 +1,100 @@
...
+    Thrift_INCLUDE_DIR
+    Thrift_LIBRARY
+)
+
+set(Thrift_VERSION_STRING ${Thrift_VERSION})
```
ihsandemir, line 100: if find_package found thrift and this may be already set? Then are we overwriting it?

ihsandemir (changes requested):
Did you test the changes work as expected with vcpkg and conan?

hazelcast/test/CMakeLists.txt
yemreinci: Done.

CMakeLists.txt
yemreinci: I updated the description. See the 2nd item.

CMakeLists.txt
yemreinci: From CMake [docs for CMAKE_MODULE_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_MODULE_PATH.html):
```
... By default it is empty, it is intended to be set by the project. 
```

This is very common. See also: https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-To-Find-Libraries

CMakeLists.txt
yemreinci: Updated the description. See item 1.

cmake/Modules/FindThrift.cmake
yemreinci: I followed the tutorial at https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html.

`find_package_handle_standard_args` does the necessary checks at the end, and prints logs depending on whether the libraries and include directories were found successfully.

cmake/Modules/FindThrift.cmake
yemreinci: That is how it's done in the [tutorial](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html). It also explains the reason.

If the find_package(Thrift) fails, the error message will tell the names of these variables so the user can set them.

cmake/Modules/FindThrift.cmake
yemreinci: This was suggested by the tutorial for compatibility reasons. I deleted it anyways.

yemreinci:
@yuce, I added a FindThrift module and updated the description. You should now be able to do one of:

```
cmake -DBUILD_TESTS=1 -DThrift_INCLUDE_DIR=/path/to/thrift/include -DThrift_LIBRARY=/path/to/thrift/lib/libthrift.a
```
or
```
cmake -DBUILD_TESTS=1 -DCMAKE_PREFIX_PATH=/path/to/thrift/
```

I already tested these on my system but could you also try it on yours?

yemreinci:
> Did you test the changes work as expected with vcpkg and conan?

Yes. 

If you use vcpkg and pass `-DCMAKE_TOOLCHAIN_FILE=/to/vcpkg/scripts/buildsystems/vcpkg.cmake`, our FindThrift.cmake is called first, it runs find_package in config mode first (I've seen this approach in FindBoost.cmake btw), the ThriftConfig.cmake provided with vcpkg is found, and the config file takes it from there.

I also tested with Conan. But Conan doesn't provide package config files, it provides its own find modules. If you use the cmake_paths generator for example, and pass `-DCMAKE_TOOLCHAIN_FILE=conan_paths.cmake` then Conan's FindThrift.cmake is executed, ours is ignored (because our directory is at the very end of CMAKE_MODULE_PATH).

yuce:
@yemreinci Thanks, new Thrift flags work expected. But should we name them as all upper-case like `THRIFT_INCLUDE_DIR`?

IMO having the `THRIFT_LIB_DIR` and let cmake figure out the correct library path `THRIFT_LIBRARY` would be ideal. But if it takes too much time I think we can leave that as is, since it works OK.

yemreinci:
> @yemreinci Thanks, new Thrift flags work expected. But should we name them all upper-case like `THRIFT_INCLUDE_DIR` to keep cmake convention?

Current CMake convention IS Thrift_INCLUDE_DIR actually :). See [standard variable names](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#standard-variable-names). Old find modules like FindBoost still maintain the UPPERCASE variables for compatibility reasons. 


> IMO having the `THRIFT_LIB_DIR` and let cmake figure out the correct library path `THRIFT_LIBRARY` would be ideal. But if it takes too much time I think we can leave that as is, since it works OK.

Xxx_LIBRARY_DIR can be added as a HINT. I'll look into that.

Thanks for the review!

yuce (dismissed):

hazelcast/test/CMakeLists.txt
yemreinci: Just did. The version before 1.10 is 1.8.1, I installed gtest/1.8.1 from Conan and it gave compile errors.

ihsandemir:
> > @yemreinci Would it be possible to add an option to override Thrift include/lib? It's possible to override Boost home dir using `-DBOOST_ROOT` but something like that is not available for Thrift in the default cmake install.
> 
> Ideally, a Thrift installation should come with package config files for CMake, and we should be able to find it using `find_package`. And if a user wants to use a different home path for Thrift, they should be able to do so by setting the variable `Thrift_ROOT`. (`<PackageName>_ROOT` is a special variable that cmake looks when `find_package` is called. [See](https://cmake.org/cmake/help/latest/variable/PackageName_ROOT.html#variable:%3CPackageName%3E_ROOT)).
> 
> Current situation is that Thrift doesn't support CMake for users. Meaning, when you install it from source or download from a system package manager (like apt, yum, etc.) it doesn't come with package config files and therefore won't be found by `find_package`. C++ package managers like Conan and Vcpkg remedy that by providing the pkg config file themselves.
> 
> So, there is a `Thrift_ROOT` but it is only usable if you installed Thrift using Conan or vcpkg.
> 
> For the general case, we can add and maintain a `FindThrift.cmake` module to the project (like it was done [here](https://github.com/apache/parquet-cpp/blob/master/cmake_modules/FindThrift.cmake)) but I am not sure how necessary that is. @ihsandemir what would you think?

@yemreinci My opinion on this: Yes, lets please make it as trivial smooth experience as possible and since it is test code we can be more flexible. Hence, a fallback option would be to use user provided include and lib directory for Thrift as a last chance. Well, it can be done via an extra `FindThrift.cmake` or just adding those paths to the project, it would be your preference. 

Yes, it can be a separate PR.

ihsandemir (commented):
Looks good overall. Just a minor comment.

ihsandemir (approved):

PR #819, opened 2021-02-24
API-125 Add Github Actions workflows for nightly macOS and Linux builds

yemreinci:
This PR adds 3 GA workflows, each for a particular operating system and architecture:

1. `nightly-ubuntu-x86_64.yml`
2. `nightly-ubuntu-i386.yml`
3. `nightly-macos-x86_64.yml`

They run `./scripts/build-unix.sh`, `./scripts/test-unix.sh` and `./scripts/verify-unix.sh` for every combination of build type (debug or release), library type (shared or static), and SSL support (on, off). 

They are scheduled to run every night and can also be triggered manually.

See the runs on my fork:
* ubuntu-x86_64: https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/597264394
* ubuntu-i386: https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/597562963, https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/598727691
* macos-x86_64: https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/597852623

**Note:** I'll send workflow(s) for nightly Windows builds in a separate pull-request.


.github/workflows/nightly-ubuntu-i386.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Make sure the target architecture is 32 bit
+        run: |
+          echo 'int main() { return sizeof(void*) != 4; }' > test.c
+          gcc test.c -oa
+          ./a
+          rm a test.c
+
+      - name: Install CMake
+        run: |
+          curl -Lo cmake-3.19.5.tar.gz \
```
yuce, line 35: How about using cmake from the Ubuntu repo?

scripts/build-unix.sh
```diff
@@ -50,15 +50,18 @@ echo "CXXFLAGS    = $CXXFLAGS"
...
+mkdir $BUILD_DIR
+cd $BUILD_DIR
+
 echo "Configuring..."
-cmake -S . -B "$BUILD_DIR" "$@"
+cmake .. "$@"
 
 echo "Building..."
-cmake --build "$BUILD_DIR" --verbose --parallel 8
+cmake --build . --verbose --parallel 2
```
yuce, line 60: Do we need to specify the `--parallel` option?

scripts/install-thrift.sh
```diff
@@ -0,0 +1,27 @@
...
+rm thrift-$1.tar.gz
+cd thrift-$1
+./configure --disable-debug --disable-tests --disable-tutorial --disable-static \
+            --without-as3 --without-erlang --without-haskell --without-java --without-perl \
+            --without-php --without-php_extension --without-rs --without-python --without-ruby \
+            --without-swift --without-csharp --without-dart --without-qt5 --without-c_glib \
+            --without-nodejs --without-nodets --without-lua --without-py3 --without-ruby \
+            --without-go --without-cl --without-haxe --without-dotnetcore --without-d \
+            --without-zlib --without-libevent
+make -j
```
yuce, line 24: Value for `-j` seems to be missing. I think we should not specify `-j` since we don't know who's running this script (or their preferences) OR we can make it a parameter to the script with default `1`.

scripts/install-thrift.sh
```diff
@@ -0,0 +1,27 @@
...
+make install
+cd ..
+rm -rf thrift-$1
```
yuce, line 27: I think we shouldn't delete anything automatically.

scripts/install-boost.sh
```diff
@@ -0,0 +1,22 @@
...
+TARBALL_NAME=boost_$(echo "$1" | tr . _)
+
+curl --silent -Lo $TARBALL_NAME.tar.gz https://dl.bintray.com/boostorg/release/$1/source/$TARBALL_NAME.tar.gz
+tar xzf $TARBALL_NAME.tar.gz
+rm $TARBALL_NAME.tar.gz
+cd $TARBALL_NAME
+./bootstrap.sh
+./b2 --with-thread --with-chrono install
+cd ..
+rm -rf $TARBALL_NAME
```
yuce, line 22: I think we shouldn't delete anything automatically.

scripts/test-unix.sh
```diff
@@ -14,7 +14,7 @@ function cleanup {
         echo "Killing client with pid ${testPid}"
         kill -9 ${testPid}
         echo "Killing start-rc.sh with pid ${rcPid}"
-        kill ${rcPid}
+        kill -9 ${rcPid}
```
yuce, line 17: If the process with PID `rcPid` doesn't exist the script exits with error. We can append something like `|| true` to prevent that.

yuce (changes requested):
LGTM except minor issues. BUT I would recommend always quoting YAML strings to prevent any problems.

scripts/install-boost.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
ihsandemir, line 1: ubuntu package may be available for 1.71 boost version see https://pkgs.org/search/?q=boost

ihsandemir (changes requested):
Your fork shows a failed build (nightly-Ubuntu-i386 (Release, STATIC, ON))? https://github.com/yemreinci/hazelcast-cpp-client/runs/1974293859?check_suite_focus=true   

Can we write the variables explicitly at the CMake file start so that I can match easily the run output and the provided build config?

Can you separate build, test and verify into 3 steps?

scripts/install-boost.sh
yuce: Ubuntu 18.04 has 1.65, 20.04 has 1.71. I think Github runs workflows on 18.04 based containers at the moment (they will upgrade to 20.04 soon)

.github/workflows/nightly-ubuntu-i386.yml
yemreinci: I realized that our CMakeLists.txt didn't work with 3.10, I addressed the problem in #808. I'll use the one from the repo once it's merged. The one from the Ubuntu 18 repo is version 3.10 btw.

scripts/install-boost.sh
yemreinci: GA is currently Ubuntu 18, will become 20 soon. But I also use i386/ubuntu, which is 18 at the moment. So, I need to install boost by hand.

yemreinci: I also don't like the fact that these things are built from source in every build. A solution could be putting these steps in a Dockerfile and publishing container images on Docker Hub. Then, we can use the `container` option from GA and download binaries instead of building them every time.
Seems like we already have Docker images for CI jobs for Python and Go clients: https://hub.docker.com/r/hazelcast/jenkins-python-client-3.9, https://hub.docker.com/r/hazelcast/jenkins-golang-client.
How does this sound?

yemreinci:
> Your fork shows a failed build (nightly-Ubuntu-i386 (Release, STATIC, ON))? https://github.com/yemreinci/hazelcast-cpp-client/runs/1974293859?check_suite_focus=true
Well, they are flaky. Sometimes others fail, but this one passes.

> Can we write the variables explicitly at the CMake file start so that I can match easily the run output and the provided build config?
I can add a workflow step that prints out every matrix parameter. Will that work? Sth like:
```
OS: Ubuntu
architecture: i386
build_type: STATIC
...
```

> Can you separate build, test and verify into 3 steps?
OK.

.github/workflows/nightly-ubuntu-i386.yml
yemreinci: Done.

yemreinci:
Building part is divided into 3 steps: Build&install, test, and verify.

Links to most recent runs:
* ubuntu i386: https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/620243954
* ubuntu x86_64: https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/620244233
* macos x86_64: https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/620244491

scripts/install-boost.sh
ihsandemir: Dockerfile is more descriptive. But only if GA uses Dockerfile directly. If you put a docker image to docker hub, then how can I verify that binary is built by the specified Dockerfile?  We previously used docker images as you suggested but it is not clear from the binary image about the configuration. Does GA have some sort of caching capabilities? How much time does this installation add to the overall build+ test time?

ihsandemir:
> step that prints out every matrix parameter. Will that work? Sth like:

I want to see this: When I expand the Build&Install log, at the very start I like to see the hazelcast build related cmake variable values, namely, CMAKE_BUILD_TYPE, WITH_OPENSSL, BUILD_SHARED_LIBS. I want to verify easily that the build is building with the expected parameters (I can see matrix parameters in the job name but that is not what I am asking for). Is this more clear?

ihsandemir (commented):
LGTM except one comment that I answered your question.

scripts/install-boost.sh
yemreinci: Github doesn't have support for directly specifying a Dockerfile in the `container: ` field of the workflow. It has to be an image. We can do the `docker build` and `docker run` steps ourselves, in which case we'd also have to cache the image by hand, check whether the dockerfile was changed etc. 
The building and installing the dependencies part takes about 5% of the total workflow duration. I guess we can ignore it.

yemreinci:
CMake  now prints a summary of the parameters at the end. Example output:
```
-- Configuration summary:
--   BUILD_SHARED_LIBS =     ON
--   WITH_OPENSSL =          OFF
--   DISABLE_LOGGING =       OFF
--   CMAKE_INSTALL_PREFIX =  /usr/local
--   BUILD_TESTS =           OFF
--   BUILD_EXAMPLES =        OFF
```

Latest runs:
https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/635640035
https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/635639809
https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/635639536

ihsandemir (dismissed):

yuce (approved):
Thanks, LGTM

PR #845, opened 2021-04-12
Fire membership events on hot-restart

### LINKED ISSUE #816: [TRACKING ISSUE] Fire membership events when persistence enabled
The tracking issue for the Java side PR.

See https://github.com/hazelcast/hazelcast/pull/18245 for details.

---

When persistence enabled, client does not fire membership
events because members starts with same uuid's.

With this pr, we make sure that when cluster restarted,
we always update the local memberlist with empty member list
first.
This way, we will be able to fire member removed/added events
even if member uuid's does not change.

Note that if cluster uuid does not change, we will not fire
any event. This happens client disconnected and connected
back to the same cluster.
Note that cluster uuid is not preserved on hotrestart with
presistence.

There was already a test to verify this behaviour but it
was working wrong.
"ClientHotRestartTest extends ClientClusterRestartEventTest"
In this pr, we make sure that the test uses same address
for the restarted member to test this correctly.

fixes https://github.com/hazelcast/hazelcast/issues/18234

yemreinci:
Fixes #816


ihsandemir (dismissed):

yemreinci:
verify-linux

ihsandemir (approved):

PR #849, opened 2021-04-16
`client_config::reliable_topic_config_map_` should not be modified concurrently

### LINKED ISSUE #848: ReliableTopic listener stops with an unexpected exception
Community reported issue: See the [slack channel conversation](https://hazelcastcommunity.slack.com/archives/C01NC9N8CAE/p1618329061009600). 

User's problem statement: 
```
I got a deadlock recently when subscribing to a reliable topic with the C++ Client v4.0.1 and Hazelcast server v4.1.1.
Basically, the deadlock only occurs if the subscription to a reliable topic is made from a member method of a class. Yeah, pretty strange.
Subscribing to a "normal" topic in the same way does not raise a deadlock. And there is no deadlock if the subscription is made like in your examples.
The event listener simply prints the received string on the standard output. So not a time-consuming operation and no call made to Hazelcast inside the listener.
I manage to create a simple project to reproduce it. The project is attached for you to try to reproduce it.
By experience, the deadlock randomly occurs after less than 30 events are received (so max 30 seconds with the provided publisher).
```

Reproducer [hz_issue.zip](https://github.com/hazelcast/hazelcast-cpp-client/files/6319057/hz_issue.zip) is attached for reference.

ihsandemir:
`client_config::reliable_topic_config_map_` can be modified during client execution and hence can be modified through different user threads. Hence, we have to use a concurrent map. Normally, we should use the `client_config` as an immutable  config of parameters, hence it should not be modified once the client started. Hence, the solution is to use a different API in the internal implementation `config::reliable_topic_config *find_reliable_topic_config(const std::string &name)` which does not modify the map but uses the default batch size.

fixes #848


sancar:
This issue does not seem to be specific to reliable_topic. 
For example, flakeIdGenerator also uses the map from config on the proxy and modifies it on some cases.

What about introducing private methods to config which does not modify internal maps and return `optional`, which will be used by proxies. 
Not sure if this causes a cyclic dependency because related proxies should be a friend class of the config.

ihsandemir:
> private

Good catch. Yes, i see that the same problems for `flake_id_generator_config` and `near_cache_config`. The private method returning a raw pointer or nullptr may also work, let me try doing that way and see hot it works. I don't think it is cyclic dependency, we just forward declare a class to be a friend and hence the config will be used as the same from the other classes.

hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -852,11 +851,20 @@ namespace hazelcast {
...
         const config::reliable_topic_config &client_config::get_reliable_topic_config(const std::string &name) {
             auto it = reliable_topic_config_map_.find(name);
-            if (reliable_topic_config_map_.end() == it) {
-                reliable_topic_config_map_[name] = config::reliable_topic_config(name.c_str());
+            if (it != reliable_topic_config_map_.end()) {
+                return it->second;
             }
 
-            return reliable_topic_config_map_[name];
+            return reliable_topic_config_map_.emplace(name, config::reliable_topic_config(name)).first->second;
```
yemreinci, line 858:
```suggestion
            return reliable_topic_config_map_.emplace(name, name).first->second;
```
emplace will construct the config object in-place.

yemreinci (dismissed):
Looks OK, just 2 suggestions.

sancar (dismissed):

ihsandemir:
verify

ihsandemir:
verify

yemreinci (dismissed):

ihsandemir:
verify-windows

hazelcast/include/hazelcast/util/BlockingConcurrentQueue.h
```diff
@@ -37,7 +37,7 @@ namespace hazelcast {
             BlockingConcurrentQueue(size_t max_queue_capacity) : capacity_(max_queue_capacity), is_interrupted_(false) {
             }
 
-            void push(const T &e) {
+            void push(const T e) {
```
yemreinci, line 40:
```suggestion
            void push(T e) {
```
If this is left as const, call to std::move below won't cause the move ctor to be called, it will be a regular copy.

ihsandemir: good catch.

ihsandemir:
Solution is verified to be working by the community user who originally reported the issue #848 .

hazelcast/include/hazelcast/util/BlockingConcurrentQueue.h
```diff
@@ -37,7 +37,7 @@ namespace hazelcast {
             BlockingConcurrentQueue(size_t max_queue_capacity) : capacity_(max_queue_capacity), is_interrupted_(false) {
```
sancar, line 37: It looks like this class is not used anymore.

sancar (approved):

yemreinci (dismissed):

hazelcast/include/hazelcast/util/BlockingConcurrentQueue.h
ihsandemir: Removed it and related tests.

yemreinci (approved):

PR #850, opened 2021-04-18
Break cyclic shared_ptr usages between ClientInvocation, Connection and cluster_view_listener::event_listener

### LINKED ISSUE #847: Address sanizer leaks
When running the address sanitizer on Linux gcc 10 we observe memory leaks as in the attached file.
[test_out.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/6315510/test_out.txt)

I observed that most of these is due to cyclic shared_ptr dependency as described at this [stackoverflow question](https://stackoverflow.com/questions/57477703/only-indirect-leaks-and-no-direct-ones).

We identified two problems:

1.  Cyclic dependencies between Connection, ClientInvocation and cluster_view_listener::event_listener. The dependency relation is at the . attached diagram (thanks to @yemreinci for pointing it)

2.  [This line of code](https://github.com/hazelcast/hazelcast-cpp-client/blob/b1a984fae0beb46873eadb3415e94ca68dd5916d/hazelcast/src/hazelcast/client/spi.cpp#L1052-L1062) may cause  `heap usage after free` problem.

ihsandemir:
Changed the strong shared_ptr usages to `weak_ptr` usages to avoid the cyclic usage and memory leaks as reported in issue #847

fixes #847

### DIFF
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -1260,7 +1260,7 @@ namespace hazelcast {
                     if (invocation.is_bind_to_single_connection()) {
                         auto conn = invocation.connection_.lock();
                         if (conn) {
-                            target << "connection " << conn;
+                            target << "connection " << *conn;
                         }
                     } else if (invocation.partition_id_ != -1) {
                         target << "partition " << invocation.partition_id_;
@@ -2392,6 +2392,5 @@ namespace std {
         return std::hash<std::string>()(k.get_service_name() + k.get_object_name());
     }
 }
-
 
 
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -1346,9 +1346,15 @@ namespace hazelcast {
 
                 void ClientInvocation::wait_invoked() const {
                     // wait until the send_connection is set
-                    while (!send_connection_.load() && lifecycle_service_.is_running()) {
+                    do {
+                        auto conn_ptr = send_connection_.load();
+                        auto sent_conn = conn_ptr->lock();
+                        if (sent_conn) {
+                            break;
+                        }
                         std::this_thread::yield();
-                    }
+                    } while (lifecycle_service_.is_running());
+
                     if (!lifecycle_service_.is_running()) {
                         BOOST_THROW_EXCEPTION(exception::illegal_argument("Client is being shut down!"));
                     }
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -1070,10 +1070,20 @@ namespace hazelcast {
                         }
 
                         if (is_bind_to_single_connection()) {
-                            auto invoked = invocation_service_.invoke_on_connection(shared_from_this(), connection_);
+                            bool invoked = false;
+                            auto conn = connection_.lock();
+                            if (conn) {
+                                invoked = invocation_service_.invoke_on_connection(shared_from_this(), conn);
+                            }
                             if (!invoked) {
-                                notify_exception(std::make_exception_ptr(exception::io("", (boost::format(
-                                        "Could not invoke on connection %1%") % *connection_).str())));
+                                std::string message;
+                                if (conn) {
+                                    message = (boost::format("Could not invoke on connection %1%") % *conn).str();
+                                } else {
+                                    message = "Could not invoke. Bound to a connection that is deleted already.";
+                                }
+                                notify_exception(std::make_exception_ptr(
+                                        exception::io("ClientInvocation::invoke_on_selection", message)));
                             }
                             return;
                         }
@@ -1104,7 +1114,8 @@ namespace hazelcast {
                 }
 
                 bool ClientInvocation::is_bind_to_single_connection() const {
-                    return connection_ != nullptr;
+                    auto conn = connection_.lock();
+                    return conn != nullptr;
                 }
 
                 void ClientInvocation::run() {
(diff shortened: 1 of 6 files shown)


hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1104,7 +1114,8 @@ namespace hazelcast {
                 }
 
                 bool ClientInvocation::is_bind_to_single_connection() const {
-                    return connection_ != nullptr;
+                    auto conn = connection_.lock();
```
sancar, line 1117: This does not feel right. 
There will be a connection managed by probably connection manager. When the connection is gone, this will return `false` even tough, the invocation was bing to single connection when created. 
I think it is better to replace this logic with a boolean which is decided in the constructor.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -2207,24 +2231,33 @@ namespace hazelcast {
                     void
                     cluster_view_listener::event_handler::handle_partitionsview(int32_t version,
                                                                                 const std::vector<std::pair<boost::uuids::uuid, std::vector<int>>> &partitions) {
-                        view_listener.client_context_.get_partition_service().handle_event(connection, version, partitions);
+                        auto conn = connection.lock();
+                        view_listener.client_context_.get_partition_service().handle_event(conn, version, partitions);
```
yemreinci, line 2235: `conn` can be empty (nullptr) here. `handle_event` calls `should_be_applied`, and `should_be_applied` dereferences the connection object without a nullptr check at spi.cpp:1693.

yemreinci (changes requested):

hazelcast/src/hazelcast/client/spi.cpp
ihsandemir: Fixed. Added a bool flag `bound_to_single_connection_` to constructor.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1325,22 +1342,29 @@ namespace hazelcast {
...
+                    std::shared_ptr<connection::Connection> connection;
+                    do {
+                        auto loaded_conn = send_connection_.load();
+                        if (loaded_conn) {
+                            connection = loaded_conn->lock();
+                        }
+                        if (!connection) {
+                            std::this_thread::yield();
+                        }
+                    } while (lifecycle_service_.is_running() && !connection);
```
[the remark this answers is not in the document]
ihsandemir: Fixed it. changed the method name to `wait_invoked` and method signature to not return anything.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -2207,24 +2231,33 @@ namespace hazelcast {
...
-                            boost::str(boost::format("Registered ClusterViewListenerHandler to %1%") 
-                                                     % *connection)
-                        );
+                        auto conn = connection.lock();
+                        if (conn) {
+                            HZ_LOG(lg, finest,
+                                   boost::str(boost::format("Registered cluster_view_listener::event_handler to %1%")
+                                              % *conn));
+                        } else {
+                            HZ_LOG(lg, finest, "cluster_view_listener::event_handler but connection is deleted already!");
```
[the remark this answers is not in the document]
ihsandemir: fixed.

hazelcast/src/hazelcast/client/spi.cpp
ihsandemir: added null check.

hazelcast/src/hazelcast/client/spi.cpp
yemreinci: Resolving.

hazelcast/src/hazelcast/client/spi.cpp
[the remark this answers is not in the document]
yemreinci: No, it still not doing the same thing as before. See the constructor of ClientInvocation. It calls set_send_connection(nullptr) and that stores a **nonempty** shared_ptr that points to an empty weak_ptr. The condition `!send_connection_.load()` in the while loop will always evaluate to false and the loop will never run. Maybe you can modify the ctor so that the send_connection_ is initially empty.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1655,59 +1671,58 @@ namespace hazelcast {
...
-                            ([&current, &connection](){
-                                auto frmt = boost::format("Event coming from a new connection. Old connection: %1%, "
+                            ([&current, connection_id](){
+                                auto frmt = boost::format("Event coming from a new connection. Old connection id: %1%, "
                                                           "new connection %2%");
 
-                                if (current.connection) {
-                                    frmt = frmt % *current.connection;
+                                if (current.connection_id) {
+                                    frmt = frmt % current.connection_id;
```
yemreinci, line 1691: OK, we are only printing the connection id instead of the whole representation (`ClientConnection{alive=0 .. }`. But is this a good idea? It's just a monotonically increased (1, 2, 3,) number, right? Say I have seen this line in the log, I have no idea what this connection actually is. Unless I search in the log to find a full representation of the whole connection object, I don't know which member sent this event. And as far as I am aware, the whole connection representation is not printed on the log until the connection is closed. So there will be no way of understanding this log line before the connection is closed.

ihsandemir:
verify

hazelcast/src/hazelcast/client/spi.cpp
ihsandemir: Yes, it is monotonically increasing number. 

I understand the concern. Well, we can add the connection info onto the log where we print the auth result like at the end of the following line:
```
20/04/2021 19:12:45.223 INFO: [0x16f373000] hz.client_103[executor-test] [4.0.1] Authenticated with server  Address[localhost:5701]:55c2b3a8-95c9-497c-95d8-e9b863934bc4, server version: 4.1, local address: Address[127.0.0.1:62077]
```

But the user yes, still needs to look at the initial logs to match for the connection. Well, this was an idea to remove the connection pointer from the `cluster_view_listener::event_handler` and just use the integer connection id as the dependency but this is at the cost of not being as explicit log statement as before. It can still be resolved looking at the whole log. So the question is which is more preferable? The `weak_ptr` solution or no pointer but connection id solution.

hazelcast/src/hazelcast/client/spi.cpp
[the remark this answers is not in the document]
ihsandemir: Should be fixed now. Can you please double check?

hazelcast/src/hazelcast/client/spi.cpp
yemreinci: I think printing connection id is fine if we print the connection id immediately after a connection is made.

ihsandemir: Added the log.

yemreinci (approved):

ihsandemir:
verify-windows

PR #851, opened 2021-04-19
Fix memory leaks in test sources

yemreinci:
This PR fixes some trivial memory leaks in the tests:

* `ReliableTopicTest` was passing heap-allocated topic::message objects to a ConcurrentQueue, and ConcurrentQueue does not delete its elements. However, a concurrent queue is not necessary for any of the test cases because the access to the messages queue is synchronized by the latch. So, I replaced it with a std::vector. And with that, ConcurrentQueue is no longer used in the code, so I deleted it as well. It is very unsafe anyways(taking raw pointers and not deleting them).
* `get_shared_value` leaked a string object every time it's called. 
* `serialization_with_server` leaked a HazelcastServerFactory.

I also removed a number of redundant include directives from test sources.


ihsandemir (approved):

yemreinci:
verify

ihsandemir (approved):

PR #858, opened 2021-04-28
Fix dangling pointer in continuation function [API-379]

### LINKED ISSUE #854: Dangling pointer in continuation lambda
All of 32 bit builds are failing with a segmentation fault. The problem started occurring after #850:
* Before #850: https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/772177841
* After #850: https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/775777570

After reproducing and running with address sanitizer to spot the issue, I found that the thread executing the lambda function here tries to dereference an already-freed pointer to a `cluster_view_listener` object:
https://github.com/hazelcast/hazelcast-cpp-client/blob/0924788e2f12824673915a11244e12b4c24ab574/hazelcast/src/hazelcast/client/spi.cpp#L2189-L2196

The problem existed before #850, but seems like the change somehow made it more capable to cause errors.

The error can be reproduced on the `i386/ubuntu` docker image. I'm sharing a part of the ASAN's error report:

```
==26260==ERROR: AddressSanitizer: heap-use-after-free on address 0xf4240970 at pc 0xf724d744 bp 0xf07fed88 sp 0xf07fed78
WRITE of size 4 at 0xf4240970 thread T29
    #0 0xf724d743 in std::__atomic_base<hazelcast::client::connection::Connection*>::compare_exchange_strong(hazelcast::client::connection::Connection*&, hazelcast::client::connection::Connection*, std::memory_order, std::memory_order) /usr/include/c++/7/bits/atomic_base.h:752
    #1 0xf724d743 in std::atomic<hazelcast::client::connection::Connection*>::compare_exchange_strong(hazelcast::client::connection::Connection*&, hazelcast::client::connection::Connection*, std::memory_order) /usr/include/c++/7/atomic:511
    #2 0xf71e0833 in hazelcast::client::spi::impl::listener::cluster_view_listener::try_reregister_to_random_connection(std::shared_ptr<hazelcast::client::connection::Connection>) /hazelcast-cpp-client/hazelcast/src/hazelcast/client/spi.cpp:2203
    #3 0xf71dfaf8 in operator() /hazelcast-cpp-client/hazelcast/src/hazelcast/client/spi.cpp:2195
    #4 0xf7221df0 in call /usr/local/include/boost/thread/future.hpp:4563
    #5 0xf721b61b in run /usr/local/include/boost/thread/future.hpp:4596
    #6 0xf7230b0a in invoke<void (*)(boost::shared_ptr<boost::detail::shared_state_base>), boost::shared_ptr<boost::detail::future_async_continuation_shared_state<boost::future<hazelcast::client::protocol::ClientMessage>, void, hazelcast::client::spi::impl::listener::cluster_view_listener::try_register(std::shared_ptr<hazelcast::client::connection::Connection>)::<lambda(boost::future<hazelcast::client::protocol::ClientMessage>)> > > > /usr/local/include/boost/thread/detail/invoke.hpp:133
    #7 0xf722f321 in run2<1> /usr/local/include/boost/thread/detail/thread.hpp:79
    #8 0xf722efca in run /usr/local/include/boost/thread/detail/thread.hpp:85
    #9 0xf609108f in thread_proxy (/usr/local/lib/libboost_thread.so.1.75.0+0x708f)
    #10 0xf7a2d610  (/usr/lib/i386-linux-gnu/libasan.so.4+0x2f610)
    #11 0xf60663bc in start_thread (/lib/i386-linux-gnu/libpthread.so.0+0x63bc)
    #12 0xf5dd8c95 in clone (/lib/i386-linux-gnu/libc.so.6+0xf8c95)
```

yemreinci:
Fixes #854.

The continuation lambda object is constructed with these things: a raw pointer to `cluster_view_listener`, a shared_ptr\<Connection\>, and a shared_ptr<event_handler>. This is problematic for different reasons: (1) `cluster_view_listener` can be destroyed before the continuation starts to execute and become a dangling pointer, (2) the Connection object holds a reference to an io_context object, which can also be dangling if the client object is destroyed.

This PR passes a weak_ptr to the cluster_view_listener. The continuation function tries to lock the weak_ptr at start to make sure the `cluster_view_listener` is still alive before dereferencing. ~And the Connection is passed as a raw pointer because it is only used for synchronization purposes and not dereferenced.~ The id of the connection is used instead of the address to determine whether a registration is being / was made already. 

I ran the failing GA jobs on PR's branch and they are now green: https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/791389693


hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -2205,21 +2205,29 @@ namespace hazelcast {
...
-                        invocation->invoke_urgent().then([=] (boost::future<protocol::ClientMessage> f) {
-                            if (f.has_value()) {
-                                handler->on_listener_register();
-                                return;
-                            }
-                            //completes with exception, listener needs to be reregistered
-                            try_reregister_to_random_connection(connection);
-                        });
+                        std::weak_ptr<cluster_view_listener> weak_self = shared_from_this();
+                        connection::Connection *raw_conn = connection.get();
```
ihsandemir, line 2209: you can remove this line and use `self.get()` below inside the lambda.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -2205,21 +2205,29 @@ namespace hazelcast {
...
+                                    if (!self)
+                                        return;
+
+                                    if (f.has_value()) {
+                                        handler->on_listener_register();
+                                        return;
+                                    }
+
+                                    //completes with exception, listener needs to be reregistered
+                                    self->try_reregister_to_random_connection(raw_conn);
```
ihsandemir, line 2223:
```suggestion
                                    self->try_reregister_to_random_connection(self.get());
```

ihsandemir (dismissed):

hazelcast/src/hazelcast/client/spi.cpp
yemreinci: I don't understand. `self` is a `shared_ptr<cluster_view_listener>`, the method takes a `Connection *`.

ihsandemir: I meant this API: https://en.cppreference.com/w/cpp/memory/shared_ptr/get

ihsandemir: I see, then don't you need to pass `std::weak_ptr<Connection>` rather than raw pointer so that we do not use deleted connection in the lambda.

yemreinci: It won't be dereferenced, only the address's equality to the `listener_added_connection_ ` variable is checked for synchronization purposes.

yemreinci: OK, I changed the code so that it uses the connection id instead of the address. This should be safer. I've started nightly builds with the new version: [linux64](https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/793671415), [linux32](https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/793671823), [mac64](https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/793672380). 

We can merge after everything is green.

ihsandemir (approved):

PR #860, opened 2021-04-28
Fix java version check in start-rc.sh

yemreinci:
Minor issue with the Java version check I added to the start-rc.sh script in #845: turns out some JDKs print the version differently. 

E.g. The JDK on my computer prints `java version "1.8.xx"`, JDK on the GA macos runner prints `openjdk version "1.8.xx"`. This change should generalize to both.


scripts/start-rc.sh
```diff
@@ -82,7 +82,7 @@ hazelcast-enterprise-${HAZELCAST_ENTERPRISE_VERSION}-tests.jar:\
 hazelcast-${HAZELCAST_TEST_VERSION}-tests.jar"
 
 # necessary arguments for Java 9+
-if ! java -version 2>&1 | grep -E "java version \"1\." > /dev/null; then
+if ! java -version 2>&1 | grep -E "version \"1\." > /dev/null; then
```
ihsandemir, line 85: actually i think that you can get a number for java version like [this solution](https://stackoverflow.com/a/7596535). When I ran it on my open jdk 16, i see the output as `16` and when I ran it for `java version "1.8.0_112"` i got `1.8.0_112` as expected. Hence, this may be a better way of parsing the output.

ihsandemir (commented):
What about windows `start-rc.bat` script?

scripts/start-rc.sh
yemreinci: OK, I used awk instead of grep for parsing the output. But the previous one worked too though, module arguments should be used if the version string does not start with "1." (i.e. version is 9 or higher), and by grepping "version 1." we check exactly that. Anyways, both methods work just fine.

I will send a separate PR for the .bat file. It's not urgent as all windows CI environments we use have java 8.

yemreinci:
verify-linux

ihsandemir (approved):

yemreinci:
verify-linux

Issue #857, opened 2021-04-28
Address sanitizer heap usage after free problem (ClientInvocation::invoke_urgent)


ihsandemir:
This [line](https://github.com/hazelcast/hazelcast-cpp-client/blob/b1a984fae0beb46873eadb3415e94ca68dd5916d/hazelcast/src/hazelcast/client/spi.cpp#L1052-L1062) of code may cause heap usage after free problem.

related to https://github.com/hazelcast/hazelcast-cpp-client/issues/847

PR #865, opened 2021-05-05
io_context should process all posted works before shutting down

ihsandemir:
Removed the stop statement while shutting down io_context so that it can indeed process all posted works before shutting down. Failing to do this may cause some posted requests, e.g., notifying any outstanding invocation requests to be not completed before shutdown, but they are expected to be completed with `target_disconnected` exception with close reason "Hazelcast client is shutting down".

Added a new test `invocation_should_not_block_indefinitely_during_client_shutdown` which can be run indefinitely and observe that it is not being blocked in any run. An example way to reproduce the problem without this PR fix is :
    ```
    for ((;;)) { /Users/ihsan/Desktop/work/src/hazelcast-cpp-client/cmake-build-arm-debug/hazelcast/test/src/client_test --gtest_filter=IssueTest.invocation_should_not_block_indefinitely_during_client_shutdown ; }
    ```
    
This run blocks after several runs in master branch but runs indefinite time with this PR fix with no problem.

I am hoping that this may fix the issues #852 and #857 .

### DIFF
--- a/hazelcast/src/hazelcast/client/network.cpp
+++ b/hazelcast/src/hazelcast/client/network.cpp
@@ -165,12 +165,8 @@ namespace hazelcast {
 
                 spi::impl::ClientExecutionServiceImpl::shutdown_thread_pool(executor_.get());
 
+                // release the guard so that the io thread can stop gracefully
                 io_guard_.reset();
-/*
-                io_context_->stop();
-                boost::asio::use_service<boost::asio::detail::resolver_service<boost::asio::ip::tcp>>(
-                        *io_context_).shutdown();
-*/
                 io_thread_.join();
 
                 connection_listeners_.clear();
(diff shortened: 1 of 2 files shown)


hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -165,10 +165,8 @@ namespace hazelcast {
 
                 spi::impl::ClientExecutionServiceImpl::shutdown_thread_pool(executor_.get());
 
+                // release the guard so that the io thread can stop gracefully
                 io_guard_.reset();
-                io_context_->stop();
```
sancar, line 169: Already scheduled ones should be completed. And this fix for that
But it feels like we are not stopping new ones to be accepted as well. Will this cause a leak on io_context_ ? May be it will continue to run after connection_manager is deleted ?

sancar (approved):

yemreinci (approved):

yemreinci:
I pushed the same branch on my fork and ran nightly builds manually:

https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/813547700
https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/813548133
https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/813548957

PR #868, opened 2021-05-06
Print the word "INFINITE" for the cluster connect timeout instead of max milliseconds number

### LINKED ISSUE #867: [TRACKING ISSUE] [PLAT-305] Pretty print cluster connect timeout in logs
The tracking issue for the Java side PR.

See https://github.com/hazelcast/hazelcast/pull/18642 for details.

---

The default value for the cluster connect timeout is infinite,
so it is represented with `Long.MAX_VALUE`. When we try to log
this as it is, we output something like below.

```
... cluster connect timeout: 9223372036854775807 ms, ...
```

To make this more clear, we will replace this with the
output below.

```
... cluster connect timeout: INFINITE, ...
```

If the user has provided some other timeout value (such as 999 ms),
the output will still be as it is now.

```
... cluster connect timeout: 999 ms, ...
```

ihsandemir:
Print the word "INFINITE" for the cluster connect timeout instead of the max milliseconds number possible for better user experience.

fixes #867

### DIFF
--- a/hazelcast/include/hazelcast/client/connection/wait_strategy.h
+++ b/hazelcast/include/hazelcast/client/connection/wait_strategy.h
@@ -48,6 +48,7 @@ namespace hazelcast {
                 std::chrono::milliseconds cluster_connect_timeout_millis_{0};
                 std::chrono::steady_clock::time_point cluster_connect_attempt_begin_{
                         std::chrono::steady_clock::now()};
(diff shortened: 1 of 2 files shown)


mdumandag (approved):

sancar (approved):

yuce (approved):
LGTM

PR #869, opened 2021-05-06
Add google benchmark tests for the C++ client map

ihsandemir:
Added the google benchmark tests for the C++ client map usages, for put, get and remove operations. It needs to be built separate from the main project. Instructions are in `README.md` file.

options that can be used with the executable:
```
hazelcast_benchmark --help 
benchmark [--benchmark_list_tests={true|false}]
          [--benchmark_filter=<regex>]
          [--benchmark_min_time=<min_time>]
          [--benchmark_repetitions=<num_repetitions>]
          [--benchmark_report_aggregates_only={true|false}]
          [--benchmark_display_aggregates_only={true|false}]
          [--benchmark_format=<console|json|csv>]
          [--benchmark_out=<filename>]
          [--benchmark_out_format=<json|console|csv>]
          [--benchmark_color={auto|true|false}]
          [--benchmark_counters_tabular={true|false}]
          [--v=<verbosity>]
```

A sample run result on my computer:
```
Unable to determine clock rate from sysctl: hw.cpufrequency: No such file or directory
2021-05-06T16:30:51+03:00
Running /Users/ihsan/Desktop/work/src/hazelcast-cpp-client/hazelcast/test/benchmark/cmake-build-debug/hazelcast_benchmark
Run on (8 X 24.1254 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 128 KiB (x8)
  L2 Unified 4096 KiB (x8)
Load Average: 2.08, 2.03, 1.84
--------------------------------------------------------------------------------
Benchmark                      Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------
map_put/threads:32         22010 ns        15185 ns        45792 Put Rate=65.8536k/s
map_get/threads:32         20527 ns        13443 ns        52192 Get Rate=74.3891k/s
map_remove/threads:32      20296 ns        13332 ns        51680 Remove Rate=75.0093k/s
```

### DIFF
--- a/hazelcast/test/benchmark/README.md
+++ b/hazelcast/test/benchmark/README.md
@@ -1,0 +1,24 @@
+#Dependencies:
+    - [google benchmark](https://github.com/google/benchmark): Mac OS: brew install google-benchmark
+    - hazelcast-cpp-client 
+    - Boost
+
+#Build Instructions:
+mkdir build
+cd build
+cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/path/to/boost/install/dir;/path/to/hazelcast-cpp-client/install/dir
+make
+
+#Running Instructions
+Run inside build directory as:
+```shell
+./hazelcast_benchmark
+```
+
+Print the options:
+```shell
+./hazelcast_benchmark --help
+```
+You can play with the number of repetitions, output format, aggregates, minimum bencmark time, etc.
+
+The test is using 32 threads by default, you can change it in the source code.
--- a/hazelcast/test/benchmark/README.md
+++ b/hazelcast/test/benchmark/README.md
@@ -6,7 +6,7 @@
 #Build Instructions:
 mkdir build
 cd build
-cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/path/to/boost/install/dir;/path/to/hazelcast-cpp-client/install/dir
+cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=/path/to/boost/install/dir;/path/to/hazelcast-cpp-client/install/dir
 make
 
 #Running Instructions
--- a/hazelcast/test/benchmark/CMakeLists.txt
+++ b/hazelcast/test/benchmark/CMakeLists.txt
@@ -1,0 +1,31 @@
+#
+# Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+cmake_minimum_required(VERSION 3.10)
+
+project(hazelcast-cpp-client-benchmark
+        VERSION 4.1.0
+        DESCRIPTION "Hazelcast C++ Client Benchmark Project"
+        LANGUAGES CXX)
+
+# Set the C++ standard
+set(CMAKE_CXX_STANDARD 11)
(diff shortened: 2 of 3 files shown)


hazelcast/test/benchmark/README.md
```diff
@@ -0,0 +1,24 @@
+#Dependencies:
+    - [google benchmark](https://github.com/google/benchmark): Mac OS: brew install google-benchmark
+    - hazelcast-cpp-client 
+    - Boost
+
+#Build Instructions:
+mkdir build
+cd build
+cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/path/to/boost/install/dir;/path/to/hazelcast-cpp-client/install/dir
```
sancar, line 9: I have tried to follow the instructions 
And I run the following

cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=/usr/local/Cellar/boost/1.75.0_2;/Users/sancar/projects/hazelcast-cpp-client

And it could find Thrift.
```
  By not providing "FindThrift.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Thrift", but
  CMake did not find one.

  Could not find a package configuration file provided by "Thrift" (requested
  version 0.13.0) with any of the following names:

    ThriftConfig.cmake
    thrift-config.cmake

  Add the installation prefix of "Thrift" to CMAKE_PREFIX_PATH or set
  "Thrift_DIR" to a directory containing one of the above files.  If "Thrift"
  provides a separate development package or SDK, be sure it has been
  installed.
```

Note that I can build the client itself with tests, so thrift should be there.

sancar: Does this needs to build the tests of hazelcast-cpp-client ? I wonder why it is trying to find Thrift

yemreinci: It shouldn't need Thrift. 
Is the path `/Users/sancar/projects/hazelcast-cpp-client` contain the repository, or is it an install directory?

ihsandemir: did you change to the benchmark directory `cd hazelcast/test/benchmark/`?

sancar: Ok, I have managed to make it work by correctly installing `hazelcast-cpp-client`

sancar (dismissed):

hazelcast/test/benchmark/README.md
yemreinci, line 9: Isn't `Release` more appropriate for benchmarks? `Debug` mode doesn't use optimization flags (`-O3`). 

From my computer: 

compiled with `-DCMAKE_BUILD_TYPE=Debug`:
```
--------------------------------------------------------------------------------
Benchmark                      Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------
map_put/threads:32         50451 ns        33010 ns        22592 Put Rate=30.2934k/s
map_get/threads:32         51861 ns        31480 ns        22304 Get Rate=31.7657k/s
map_remove/threads:32      48031 ns        30989 ns        22304 Remove Rate=32.2697k/s
```

compiled with `-DCMAKE_BUILD_TYPE=Release`:
```
--------------------------------------------------------------------------------
Benchmark                      Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------
map_put/threads:32         45438 ns        21636 ns        35392 Put Rate=46.2183k/s
map_get/threads:32         44832 ns        21132 ns        33024 Get Rate=47.3207k/s
map_remove/threads:32      44187 ns        20806 ns        32832 Remove Rate=48.0634k/s
```

Release is about ~50% better.

**Note:** In both of the above, the client itself was compiled in `Release` mode, the difference is just the benchmark suite.

yemreinci (approved):

hazelcast/test/benchmark/README.md
ihsandemir: It is actually user specified and this was a recommendation for the default, but yes, I now changed it to Release as suggested in the README. By the way, even though they both use the Release version of hazelcast-cpp-client library it is also interesting to see this much difference for the performance test program build. This may be the affect of the templated library code optimization.

yemreinci: Someone who is inexperienced with CMake (a QA person for example) may read this, its better to suggest the optimizing build for better comparison.

We have considerable amount of code in the header files, the optimization of them are done when the benchmark suite is compiled. The performance difference can be because of that.

Issue #873, opened 2021-05-11
Test executable crashes when running ReliableTopicTest


yemreinci:
The test executable occasionally crashes on GA while executing the tests under the suite `ReliableTopicTest`. 

Here is a run that crashed: https://github.com/hazelcast/hazelcast-cpp-client/runs/2559526252?check_suite_focus=true

ihsandemir:
Can you also attached the stack trace and log here in case the GA build log is lost?

yemreinci:
> Can you also attached the stack trace and log here in case the GA build log is lost?

[2_Ubuntu-x86_64 (Debug, Static, SSL).txt](https://github.com/hazelcast/hazelcast-cpp-client/files/6464645/2_Ubuntu-x86_64.Debug.Static.SSL.txt)

hakanaktas0:
Related to [#852](https://github.com/hazelcast/hazelcast-cpp-client/issues/852)

akeles85:
Solved with PR #1071

PR #875, opened 2021-05-11
[API-351] Add Github Actions workflow for nightly Windows builds

yemreinci:
Adds a GA workflow that builds, installs, tests and verifies the client on Windows. 

The disk space of Github's runners (16GB) was not enough for some of the configurations ([example run](https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/805088972)). To save space, the verification step builds examples one by one and cleans the build directory regularly.  

Here is the passing build from my fork: https://github.com/yemreinci/hazelcast-cpp-client/runs/2560353013?check_suite_focus=true. The red ones are related to #873 and not in the context of this PR.


.github/workflows/nightly-windows.yml
```diff
@@ -0,0 +1,133 @@
...
+        run: |
+          choco install openssl ${{ matrix.arch.choco_options }}
+          Invoke-WebRequest `
+              "https://curl.se/ca/cacert.pem" `
+              -OutFile "C:\cacert.pem"
+
+      - name: Install Boost
+        run: |
+          Invoke-WebRequest `
+              "https://boostorg.jfrog.io/artifactory/main/release/1.75.0/source/boost_1_75_0.tar.gz" `
```
ihsandemir, line 53:
```suggestion
              "https://boostorg.jfrog.io/artifactory/main/release/1.76.0/source/boost_1_76_0.tar.gz" `
```

.github/workflows/nightly-windows.yml
```diff
@@ -0,0 +1,133 @@
...
+              -OutFile "boost_1_75_0.tar.gz"
```
ihsandemir, line 54:
```suggestion
              -OutFile "boost_1_76_0.tar.gz"
```

.github/workflows/nightly-windows.yml
```diff
@@ -0,0 +1,133 @@
...
+          tar xzf boost_1_75_0.tar.gz
```
ihsandemir, line 55:
```suggestion
          tar xzf boost_1_76_0.tar.gz
```

.github/workflows/nightly-windows.yml
```diff
@@ -0,0 +1,133 @@
...
+          rm boost_1_75_0.tar.gz
```
ihsandemir, line 56:
```suggestion
          rm boost_1_76_0.tar.gz
```

.github/workflows/nightly-windows.yml
```diff
@@ -0,0 +1,133 @@
...
+          cd boost_1_75_0
```
ihsandemir, line 57:
```suggestion
          cd boost_1_76_0
```

.github/workflows/nightly-windows.yml
```diff
@@ -0,0 +1,133 @@
...
+          .\bootstrap.bat
+          .\b2 address-model=${{ matrix.arch.address_model }} --with-thread --with-chrono install
+          cd ..
+          Remove-Item boost_1_75_0 -Recurse -Force
```
ihsandemir, line 61:
```suggestion
          Remove-Item boost_1_76_0 -Recurse -Force
```

.github/workflows/nightly-windows.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Install Thrift
```
ihsandemir, line 63: vcpkg has thrift, boost and openssl, why not just use vcpkg for dependency management (and use `-DCMAKE_TOOLCHAIN_FILE` option)? is choco more preferable compared to vcpkg? vcpkg repo already has thrift 0.13.0 in the ports.

hazelcast/test/src/HazelcastTests7.cpp
```diff
@@ -1322,12 +1322,8 @@ namespace hazelcast {
                     clientConfig.get_network_config().get_aws_config().set_enabled(true).
                             set_access_key(std::getenv("AWS_ACCESS_KEY_ID")).set_secret_key(std::getenv("AWS_SECRET_ACCESS_KEY")).
                             set_tag_key("aws-test-tag").set_tag_value("aws-tag-value-1");
-
-#if  defined(WIN32) || defined(_WIN32) || defined(WIN64) || defined(_WIN64)
```
ihsandemir, line 1326: if you delete tis line, how will we test `.set_inside_aws(true)` API? This code runs at the windows aws machine.

scripts/build-windows.bat
```diff
@@ -34,7 +34,7 @@ cmake --build %BUILD_DIR% --verbose --parallel --config %BUILD_CONFIGURATION% ||
 
 if "%INSTALL%" == "ON" (
     @echo Installing...
-    cmake --install %BUILD_DIR% || exit /b 1
+    cmake --install %BUILD_DIR% --config %BUILD_CONFIGURATION% || exit /b 1
```
ihsandemir, line 37: How was the previous version without this flag working?

ihsandemir (changes requested):
Is choco a widely used package manager? should we add our client into its community repository?

.github/workflows/nightly-windows.yml
yemreinci: Yes it could be done, but vcpkg is too slow. Installing Boost, for example, took nearly an hour on Github's runners. Chocolatey is faster because it downloads binaries instead of building packages from source. I would install Boost and Thrift from Chocolatey as well if it had the versions we need but it doesn't.

scripts/build-windows.bat
yemreinci: When no --config is given, it installs all existing configs. But since do-all-windows.bat sets `-DCMAKE_CONFIGURATION_TYPES=%BUILD_CONFIG%`, only the %BUILD_CONFIG% exists and it gets installed.

I think it would be better to put the -DCMAKE_CONFIGURATION_TYPE option in this file instead of setting it from the do-all script. I will do that. Also, I realized the CMAKE_TOOLCHAIN_FILE=%VCPKG_ROOT%... line above in line 27 which is not used by the GA workflow. Setting that from the do-all script is more appropriate, I think.

hazelcast/test/src/HazelcastTests7.cpp
yemreinci: I think there should be separate test cases (they can be parametrized) that use `.set_inside_aws(true)` and they should only be run if the corresponding runner is on AWS.

.github/workflows/nightly-windows.yml
ihsandemir: i am still confused how it is faster when you now compile boost manually anyway? hence, no help from choco on the boost side. why do you think that manually compiling boost is faster than having boost being compiled using vcpkg (i wonder if it has anything to do with git clone of vcpkg)?

is choco already installed on the GA container, or do we install it? what about conan versus choco? conan has the boost and thrift as up to date versions available and it downloads binaries? choco also may be completely ok but if conan is used you may need to have python on the windows machine but then you would not need the steps to compile thrift and boost.

hazelcast/test/src/HazelcastTests7.cpp
ihsandemir: yes, it can be done, lets put these in place before removing these usages so that we do not lose the coverage.

.github/workflows/nightly-windows.yml
yemreinci: >  i am still confused how it is faster when you now compile boost manually anyway? hence, no help from choco on the boost side. why do you think that manually compiling boost is faster than having boost being compiled using vcpkg (i wonder if it has anything to do with git clone of vcpkg)?

I don't know the exact reason but it was also mentioned here: https://github.com/Microsoft/vcpkg/issues/3724#issuecomment-605609605 . It's slow even if you specify only a few packages of Boost. 

> is choco already installed on the GA container, or do we install it?

It's already installed. GA virtual environments come contain many preinstalled packages. A detailed list for Windows is given [here](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md). The preinstalled package managers are under[ this section](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#package-management). There is vcpkg but no Conan.

>  conan has the boost and thrift as up to date versions available and it downloads binaries

Conan is not guaranteed to download binaries. Sometimes the binary for your OS and architecture isn't cached on the servers and you have to add the `--build` option. I considered Conan when I was implementing macOS and Linux builds but I decided against it as it required too much setup (creating a conan.yml/py file, passing the toolchain file, etc.).

What is the upside of using a package manager for CI jobs? We currently have a working workflow file, and, considering the build time, I think we are at a good state. Building/downloading OpenSSL + Boost + Thrift takes about 6 minutes, which is 10% of the total CI time. Plus, we have more control when we build from source. For example, I am using `-DBUILD_COMPILER=OFF` when building Thrift because we don't need the thrift compiler.

The package management stuff in this PR is actually very similar to what we did for  Linux. We install OpenSSL from package managers (`apt-get` for Ubuntu, `choco` for Windows), and build Boost and Thrift from source. I kind of think of Chocolatey as the equivalent of `apt` on Ubuntu (I may be wrong). It doesn't provide many packages or options but it's fast and reliable.

ihsandemir: i was just thinking that if it would be even faster than 6mins if you used conan binary packages. if it is also slow or now available, it is not an important problem and we can live with current solution since this is for preparing test environment anyway.

yemreinci: As I said, Conan is not guaranteed to have binaries. For example, [in this workflow](https://github.com/yemreinci/hazelcast-cpp-client/runs/2709457575?check_suite_focus=true) I tried installing dependencies and it couldn't find the binary for thrift/0.13.0. I believe problems like these will be likely to occur for other platforms and architectures as well. Although they can be worked around (adding --build parameter if failed etc.) I don't think they are worth the effort.

I have [also ran with](https://github.com/yemreinci/hazelcast-cpp-client/runs/2709506621?check_suite_focus=true) `--build thrift`, and it took 6.5 mins.

hazelcast/test/src/HazelcastTests7.cpp
yemreinci: I have parametrized the first 3 tests of the `AwsClientTest` suite. They will run both with `set_inside_aws(true)` and `set_inside_aws(false)`. The `true` one will be skipped if the environment variable `INSIDE_AWS` is not set. I will add `set INSIDE_AWS=1` for Windows jobs on Jenkins.

ihsandemir (approved):

PR #884, opened 2021-06-10
Take HZ version as a parameter in start-rc scripts

### LINKED ISSUE #879: Support running with different hazelcast versions in test-X.sh/bat
We should be able to specify the server version before running `test-X.sh/bat` and `start-rc.bat/sh` scripts. This is required for backward compatibility tests. Something like this could work:

```
HZ_VERSION=4.1
./test-unix.sh
```

yemreinci:
Fixes #879


scripts/start-rc.sh
```diff
@@ -15,7 +15,7 @@ set +x
 
 trap cleanup EXIT
 
-HZ_VERSION="4.1.4-SNAPSHOT"
+HZ_VERSION="${HZ_VERSION:-4.1.4-SNAPSHOT}"
```
ihsandemir, line 18: why don't you have the same `if` check as in `[start-rc.bat](https://github.com/hazelcast/hazelcast-cpp-client/pull/884/files#diff-75df704834e901ab6db5127a7e2b4929c058b10692782b95ef52ec39d45020f5R4)`?

yemreinci: This syntax does the same thing: if HZ_VERSION was not set, it uses the default 4.1.1-SNAPSHOT. Windows batch scripts doesn't have a similar syntax, so I used an if.

ihsandemir (approved):

PR #885, opened 2021-06-10
Add .clang-format file, format code

yemreinci:
I have added a `.clang-format` file and applied the suggested changes to every source file. I haven't made any manual editing, all changes were produced by clang-format. So, the code should still behave the same and a detailed review shouldn't be necessary.

The formatting configuration is open for discussion, I can change the parameters if we don't like something.


yemreinci:
Closing in favor of #946

PR #890, opened 2021-06-21
Add the constructor to `in_predicate` where we can also accept a vector [API-551]

### LINKED ISSUE #888: strange in_predicate behaviour
C++ compiler version: gcc-9
Hazelcast Cpp client version: 4.1.1
Hazelcast server version: 4.2
Number of the clients: 1

#### Expected behaviour
Having a map<string, string> I would expect the in_predicate for keys to contain a vector<string> similar to queryExample.cpp
e.g.
```
std::vector<std::string> myKeys ...;
map->remove_all(query::in_predicate(hz, query::query_constants::KEY_ATTRIBUTE_NAME, myKeys)).get();
```

#### Actual behaviour
```
It works for one string:
map->remove_all(query::in_predicate(hz, query::query_constants::KEY_ATTRIBUTE_NAME, myKeys[0])).get();
but it does not work for a vector<string>:
map->remove_all(query::in_predicate(hz, query::query_constants::KEY_ATTRIBUTE_NAME, myKeys)).get();

```

The main use-case for me is a fast removal of thousands of entries - without predicates I would have to start a lot of futures in imap->delete_entry with presumable a lot of overhead. If this is expected behaviour, it should be annoted at least in the source code, e.g. in predicates.h.

```
int main()
{
    auto hz = hazelcast::new_client().get();
    auto map = hz.get_map("testmap").get();
    map->clear().get();
    std::vector<uint8_t> test1(5, 112);
    map->put<std::string, std::string >("12", "hello").get();
    map->put<std::string, std::string >("23", "world").get();
    map->put<std::string, std::string >("34", "!!!!").get();


    std::vector<std::string> myKeys;
    myKeys.push_back("12");
    myKeys.push_back("34");

    auto v = map->entry_set<std::string, std::string>(
                query::in_predicate(hz, query::query_constants::KEY_ATTRIBUTE_NAME, myKeys)).get();

    assert(v.size() == 2);

    map->remove_all(query::in_predicate(hz, query::query_constants::KEY_ATTRIBUTE_NAME, myKeys)).get();

    return 0;
}
```

ihsandemir:
Added the constructor to `in_predicate` where we can also accept a vector of keys/values for the query. Also added a new test `IssueTest::issue_888` and updated the `ClientMapTest::testValuesWithPredicate` to use a vector of keys. The query example should also work with this fix.

fixes #888


hazelcast/include/hazelcast/client/query/predicates.h
```diff
@@ -64,6 +64,15 @@ namespace hazelcast {
                     out_stream.write<int32_t>(static_cast<int32_t>(sizeof...(values)));
                     out_stream.write_objects(values...);
                 }
+
+                template<typename T>
+                multi_predicate(const std::string attribute_name, hazelcast_client &client, const std::vector<T> &values) : base_predicate(client) {
```
yuce, line 69: Why is the `hazelcast_client` needed to create a predicate? IMO it would be great to get rid of it if possible.

ihsandemir: unfortunately, we need it to access the serialization service. I also do not like it but i had to do it due to the fact that i needed to prepare the binary during construction.

yuce: I see. I wish we could defer binary construction. Taking a client as a parameter has more problems, like would passing a different to `multi_predicate` than the one calling the actual query cause problems? But I guess this is not something that can be changed in this PR.

yuce (approved):

hazelcast/include/hazelcast/client/query/predicates.h
```diff
@@ -64,6 +64,15 @@ namespace hazelcast {
...
+                    out_stream.write(attribute_name);
+                    out_stream.write<int32_t>(static_cast<int32_t>(values.size()));
+                    for (const T &value : values) {
```
yemreinci, line 72: won't compile when T is `bool`

```suggestion
                    for (std::vector<T>::const_reference value : values) {
```

yemreinci (approved):

hazelcast/include/hazelcast/client/query/predicates.h
yemreinci: sorry, my mistake. it does compile. Doesn't compile when reference isn't const.

hazelcast/include/hazelcast/client/query/predicates.h
ihsandemir: Passing a different client, yes, that may cause a problem if serialization configs are different. Alternative was to use the client as the factory method to create the `in_predicate` but it again has the same problem if the user would use a client predicate with the other client. I am open to any better solution but could not find one as of yet. Yes, we can open another issue to discuss this topic.

hazelcast/include/hazelcast/client/query/predicates.h
ihsandemir: I think that the suggestion is the same as the existing one, do you really want the suggestion applied?

PR #897, opened 2021-07-05
Implement binary metrics format [API-247]

yemreinci:
Implements the new binary metrics format. For compatibility with MC 4.0, metrics are still sent via the clientAttributes parameter too.

I have also added the "runtime.availableProcessors" metric because it's very simple to implement with the `std::thread::hardware_concurrency()`, and more importantly, MC 4.0.x releases have [this line](https://github.com/hazelcast/management-center/blob/cff5c3bba169c5db6e6018970c7f2bd6ee313d1e/src/main/java/com/hazelcast/webmonitor/controller/api/ClientController.java#L238) which causes a NullPtrException to be thrown when runtime.availableProcessors isn't sent.

For now, the Zlib dependency is found via find_package during CMake config step (same with how we find boost). Alternatively, we can copy the zlib source code in our repo and build it ourselves during installation.


hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+std::vector<byte> zlib_compress(const std::vector<byte> &input) {
+    const auto max_output_len = compressBound(input.size());
+
+    auto output_len = max_output_len;
+    std::vector<byte> output(output_len);
+
+    int rc = compress2(output.data(), &output_len, input.data(), input.size(), Z_BEST_SPEED);
+
+    if (rc == Z_MEM_ERROR) {
+        throw std::bad_alloc{};
```
ihsandemir: why not use one of our library exceptions? we usually say that we will throw exception derived from `iexception`.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    assert(rc == Z_OK);
```
ihsandemir: in release build, continuing without this check would be a problem.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+std::vector<byte> metrics_compressor::get_blob() {
+    write_dictionary();
+    
+    std::vector<byte> compressed_dictionary = zlib_compress(dictionary_buffer_.content());
+    std::vector<byte> compressed_metrics = zlib_compress(metrics_buffer_.content());
+
+    output_buffer blob;
+
+    blob.write(static_cast<byte>((BINARY_FORMAT_VERSION >> BITS_IN_BYTE) & BYTE_MASK));
+    blob.write(static_cast<byte>(BINARY_FORMAT_VERSION & BYTE_MASK));
```
ihsandemir, line 261: if BINARY_FORMAT_VERSION is defined as byte, you would not need to`&` with `BYTE_MASK `, right?

hazelcast/src/hazelcast/client/stats.cpp
```diff
@@ -425,6 +536,8 @@ namespace hazelcast {
                 }
 
                 const double NearCacheStatsImpl::PERCENTAGE = 100.0;
+
```
ihsandemir, line 539: extra spaces

hazelcast/test/src/HazelcastTests1.cpp
```diff
@@ -443,6 +443,7 @@ namespace hazelcast {
     }
 }
 
+
```
ihsandemir, line 446: extra new line

hazelcast/include/hazelcast/client/impl/metrics/metric_descriptor.h
```diff
@@ -0,0 +1,73 @@
...
+    NS = 2,
+    PERCENT = 3,
+    COUNT = 4,
+    BOOLEAN = 5,
+    ENUM = 6,
+
+    NONE = -1
+};
+
+class HAZELCAST_API metric_descriptor
```
ihsandemir, line 43: why not make this a simple struct with all public fields?

hazelcast/include/hazelcast/client/impl/metrics/metrics_compressor.h
```diff
@@ -0,0 +1,73 @@
...
+class HAZELCAST_API metrics_compressor
+{
+public:
+    void add_long(const metric_descriptor &descriptor, int64_t value);
+    void add_double(const metric_descriptor &descriptor, double value);
+    std::vector<byte> get_blob();
+
+private:
+    int calculate_descriptor_mask(const metric_descriptor &descriptor);
+    int get_dictionary_id(const boost::optional<std::string> &word);
```
ihsandemir, line 59:
```suggestion
    int get_or_put_into_dictionary(const boost::optional<std::string> &word);
```

The method name is not intuitive in the sense that it not only gets but it can also modify the underlying dictionary. A better naming would be easier to understand.

hazelcast/include/hazelcast/client/impl/metrics/metrics_dictionary.h
```diff
@@ -0,0 +1,51 @@
...
+class HAZELCAST_API metrics_dictionary
+{
+private:
+    using map = std::map<std::string, int>;
+public:
+    using const_iterator = map::const_iterator;
+
+    int get_dictionary_id(const std::string &word);
+
+    const_iterator begin() const noexcept;
```
ihsandemir, line 40: these are for testing only, right? you can either clearly document this or make it private and use friend.

ihsandemir (changes requested):
test `metrics_compressor_test, test_blob` fails on my local. Output:

```
/Users/ihsan/Desktop/work/src/hazelcast-cpp-client/hazelcast/test/src/metrics_test.cpp:108: Failure
Expected equality of these values:
  expected_blob
    Which is: { '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '\n' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), '=' (61, 0x3D), ... }
  actual_blob
    Which is: { '\0', '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '\n' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), ... }
With diff:
@@ -1,2 +1,2 @@
-{ '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '
-' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), '=' (61, 0x3D), ... }
+{ '\0', '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '
+' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), ... }
```

hazelcast/include/hazelcast/client/impl/metrics/metrics_dictionary.h
yemreinci: No. They are for this: 

> test `metrics_compressor_test, test_blob` fails on my local. Output:
> 
> ```
> /Users/ihsan/Desktop/work/src/hazelcast-cpp-client/hazelcast/test/src/metrics_test.cpp:108: Failure
> Expected equality of these values:
>   expected_blob
>     Which is: { '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '\n' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), '=' (61, 0x3D), ... }
>   actual_blob
>     Which is: { '\0', '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '\n' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), ... }
> With diff:
> @@ -1,2 +1,2 @@
> -{ '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '
> -' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), '=' (61, 0x3D), ... }
> +{ '\0', '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '
> +' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), ... }
> ```

Probably because Apple has its own Zlib implementation and is producing a different binary. This won't matter if we use our implementation.

yemreinci:
> test `metrics_compressor_test, test_blob` fails on my local. Output:
> 
> ```
> /Users/ihsan/Desktop/work/src/hazelcast-cpp-client/hazelcast/test/src/metrics_test.cpp:108: Failure
> Expected equality of these values:
>   expected_blob
>     Which is: { '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '\n' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), '=' (61, 0x3D), ... }
>   actual_blob
>     Which is: { '\0', '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '\n' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), ... }
> With diff:
> @@ -1,2 +1,2 @@
> -{ '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '
> -' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), '=' (61, 0x3D), ... }
> +{ '\0', '\x1' (1), '\0', '\0', '\0', '\x98' (152), 'x' (120, 0x78), '\x1' (1), '\xED' (237), '\x94' (148), 'Y' (89, 0x59), '\xE' (14), '\xC2' (194), '0' (48, 0x30), '\f' (12, 0xC), '\x5' (5), '\x87' (135), 'P' (80, 0x50), '(' (40, 0x28), '\xFB' (251), 'z' (122, 0x7A), '\x12' (18), '\xB6' (182), '\x3' (3), 'A' (65, 0x41), '\x1' (1), '\xB5' (181), '\b' (8), '
> +' (10, 0xA), '\x2' (2), '\x84' (132), '\xB8' (184), ... }
> ```

Probably because Apple has its own ZLib implementation and it is producing a different binary. This won't be a problem if we use our Zlib implementation.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+    auto pos = buffer_.size();
+    buffer_.resize(pos + sizeof(int64_t));
+
+    boost::endian::endian_store<int64_t, sizeof(int64_t), boost::endian::order::big>(
+      buffer_.data() + pos, val);
+}
+
+void output_buffer::write(double val) {
+    auto pos = buffer_.size();
+    buffer_.resize(pos + sizeof(double));
```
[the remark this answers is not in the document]
yemreinci: I think we already assuming `double` is a IE-754 64bit floating point number when we do [this](https://github.com/hazelcast/hazelcast-cpp-client/blob/a97de79c20b071e744efb8d94e56dae224c3c8b2/hazelcast/include/hazelcast/client/serialization/pimpl/data_input.h#L153-L161) and [this](https://github.com/hazelcast/hazelcast-cpp-client/blob/a97de79c20b071e744efb8d94e56dae224c3c8b2/hazelcast/src/hazelcast/client/serialization.cpp#L501-L510). I can add a static assert to enforce `std::numeric_limits<double>::is_iec559 `.

I also realized it is actually not necessary. We don't send double metrics anyways. I can remove the double support completely.

hazelcast/src/hazelcast/client/stats.cpp
```diff
@@ -162,45 +170,148 @@ namespace hazelcast {
...
-                        add_stat(stats, prefix, "ownedEntryCount", nearCacheStats->get_owned_entry_count());
-                        add_stat(stats, prefix, "expirations", nearCacheStats->get_expirations());
-                        add_stat(stats, prefix, "invalidations", nearCacheStats->get_invalidations());
-                        add_stat(stats, prefix, "invalidationRequests",
-                                nearCacheStats->get_invalidation_requests());
-                        add_stat(stats, prefix, "ownedEntryMemoryCost",
-                                nearCacheStats->get_owned_entry_memory_cost());
-                        std::string persistenceFailure = nearCacheStats->get_last_persistence_failure();
-                        if (!persistenceFailure.empty()) {
-                            add_stat(stats, prefix, "lastPersistenceFailure", persistenceFailure);
```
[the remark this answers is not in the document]
yemreinci: That's because that field is never filled anywhere, it's always empty.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+constexpr int MASK_DISCRIMINATOR_VALUE = 1 << 3;
+constexpr int MASK_UNIT = 1 << 4;
+constexpr int MASK_EXCLUDED_TARGETS = 1 << 5;
+constexpr int MASK_TAG_COUNT = 1 << 6;
+
+constexpr int SIZE_VERSION = 2;
+constexpr int SIZE_DICTIONARY_BLOB = 4;
+constexpr int SIZE_COUNT_METRICS = 4;
+
+constexpr int VALUE_TYPE_LONG = 0;
```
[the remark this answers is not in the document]
yemreinci: They are just two constants used only once in the code. Enums are good when you pass these around and want to be type-safe.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+namespace client {
+namespace impl {
+namespace metrics {
+
+namespace {
+
+constexpr int MAX_WORD_LENGTH = 255;
+constexpr int NULL_DICTIONARY_ID = -1;
+
+constexpr int MASK_PREFIX = 1 << 0;
```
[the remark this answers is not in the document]
yemreinci: I think so. When you think in context, prefix is the 0th bit of the mask, metric is the 1st etc.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+            mask |= MASK_DISCRIMINATOR_VALUE;
+        }
+
+        if (descriptor.unit() == last_descriptor_->unit()) {
+            mask |= MASK_UNIT;
+        }
+        
+        // include excludedTargets and tagCount bits for compatibility purposes
+        mask |= MASK_EXCLUDED_TARGETS;
+        mask |= MASK_TAG_COUNT;
```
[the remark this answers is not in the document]
yemreinci: If the last_descriptor_ isn't set yet the resulting mask is 0.

hazelcast/include/hazelcast/client/impl/metrics/metrics_compressor.h
yemreinci: getDictionaryId is how it's named in other clients and I don't think `get_or_put_into_dictionary` is more intuitive either.

hazelcast/include/hazelcast/client/impl/metrics/metrics_dictionary.h
yemreinci: begin() and end() calls are used by the for loop in metrics_compressor::write_dictionary.

hazelcast/include/hazelcast/client/impl/metrics/metrics_dictionary.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace hazelcast {
+namespace client {
+namespace impl {
+namespace metrics {
+
+class HAZELCAST_API metrics_dictionary
+{
+private:
+    using map = std::map<std::string, int>;
```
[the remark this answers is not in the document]
yemreinci: We need words to be ordered. Java implementation is using a TreeMap for this.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+    buffer_.insert(buffer_.end(), vec.begin(), vec.end());
+}
+
+const std::vector<byte> &output_buffer::content() const {
+    return buffer_;
+}
+
+void metrics_compressor::add_long(const metric_descriptor &descriptor, int64_t value) {
+    write_descriptor(descriptor);
+    metrics_buffer_.write(static_cast<byte>(VALUE_TYPE_LONG));
```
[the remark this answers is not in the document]
yemreinci: casting makes it obvious that a byte is being written into the buffer. VALUE_TYPE_LONG is defined constexpr anyways, the compiled code wont have any type conversion, they will be optimized.

hazelcast/src/hazelcast/client/stats.cpp
```diff
@@ -162,45 +170,148 @@ namespace hazelcast {
...
-                        }
-                    }
+                void Statistics::PeriodicStatistics::add_near_cache_metric(std::ostringstream &stats, 
+                                                                           metrics::metrics_compressor &compressor,
+                                                                           const std::string &metric,
+                                                                           const std::string &near_cache_name,
+                                                                           const std::string &near_cache_name_with_prefix,
+                                                                           int64_t value,
+                                                                           metrics::probe_unit unit) {
```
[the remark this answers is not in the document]
yemreinci: forgot to included that in the descriptor below. fixed.

hazelcast/src/hazelcast/client/stats.cpp
```diff
@@ -142,7 +149,8 @@ namespace hazelcast {
                 }
 
                 void Statistics::PeriodicStatistics::fill_metrics(std::ostringstream &stats,
-                                                                 const std::shared_ptr<connection::Connection> &connection) {
+                                                                  metrics::metrics_compressor &compressor,
+                                                                  const std::shared_ptr<connection::Connection> &connection) {
                     stats << "lastStatisticsCollectionTime" << KEY_VALUE_SEPARATOR << util::current_time_millis();
                     add_stat(stats, "enterprise", false);
                     add_stat(stats, "clientType", protocol::ClientTypes::CPP);
```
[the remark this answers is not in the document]
yemreinci: No, these should be sent only in `clientAttributes`.

hazelcast/src/hazelcast/client/metrics.cpp
yemreinci: BINARY_FORMAT_VERSION is an int.

hazelcast/src/hazelcast/client/metrics.cpp
yemreinci: Call to compress2 above can only return Z_MEM_ERROR and its checked. assert here is just for safety. not necessary for release builds.

hazelcast/src/hazelcast/client/metrics.cpp
yemreinci: bad_alloc is the exception you would get when you don't have enough memory. Hazelcast c++ client functions may as well throw this exception when memory is full.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+void metrics_compressor::write_dictionary() {
+    std::string last_word{ "" };
+
+    dictionary_buffer_.write(static_cast<int32_t>(dictionary_.size()));
+
+    for (const auto &item : dictionary_) {
+        const auto &word = item.first;
+        const auto &id = item.second;
+
+        int common_len = find_common_prefix_length(last_word, word);
```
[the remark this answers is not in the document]
yemreinci: defined common_len with auto.

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+    output_buffer blob;
+
+    blob.write(static_cast<byte>((BINARY_FORMAT_VERSION >> BITS_IN_BYTE) & BYTE_MASK));
+    blob.write(static_cast<byte>(BINARY_FORMAT_VERSION & BYTE_MASK));
+    blob.write(static_cast<int32_t>(compressed_dictionary.size()));
+    blob.write(compressed_dictionary);
+    blob.write(static_cast<int32_t>(metrics_count));
+    blob.write(compressed_metrics);
+
+    return blob.content();
```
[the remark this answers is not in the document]
yemreinci: it doesn't. made .content() moveable added returned by moving.

hazelcast/src/hazelcast/client/stats.cpp
yemreinci: deleted.

hazelcast/test/src/HazelcastTests1.cpp
yemreinci: deleted

hazelcast/include/hazelcast/client/impl/metrics/metric_descriptor.h
yemreinci: why not do the same for all classes? :)

hazelcast/src/hazelcast/client/metrics.cpp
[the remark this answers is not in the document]
ihsandemir: yes, lets remove if unused.

hazelcast/src/hazelcast/client/stats.cpp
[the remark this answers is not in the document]
ihsandemir: OK, it seems like it is only filled if nearcache preloader exists and we do not have it implemented yet. We should not forget it when preloader is implemented.

hazelcast/src/hazelcast/client/metrics.cpp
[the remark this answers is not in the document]
ihsandemir: i still don't see the added value to define it as in and do casting. you can always define `constexpr byte VALUE_TYPE_LONG = 0;` and this is as obvious, no need for compiler optimization.

hazelcast/src/hazelcast/client/stats.cpp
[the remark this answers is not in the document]
ihsandemir: does this mean that they can not get rid of `clientAttributes`(and only use binary blob) even at 5.0 unless they change this?

hazelcast/src/hazelcast/client/metrics.cpp
ihsandemir: does it have to be int? I look at used places and byte seems to be enough.

hazelcast/include/hazelcast/client/impl/metrics/metric_descriptor.h
ihsandemir: This class does not have any method with some logic in it, hence, i think it is better for such cases to use struct. Some older classes reflected Java but I prefer this lately if no method with specific logic in it.

ihsandemir:
Based on our meeting with non-java clients it is decided that we go with the binary compressed format. Hence, let's make explicit external dependency to zlib library and provide a cmake flag compile like `WITH_STATS` (disabled by default) to enable the stats.

yemreinci:
> Based on our meeting with non-java clients it is decided that we go with the binary compressed format. Hence, let's make explicit external dependency to zlib library and provide a cmake flag compile like `WITH_STATS` (disabled by default) to enable the stats.

OK. I'll do that ask for a re-review.

hazelcast/src/hazelcast/client/stats.cpp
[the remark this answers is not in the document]
yemreinci: I guess.

hazelcast/src/hazelcast/client/metrics.cpp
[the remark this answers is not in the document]
yemreinci: removed

hazelcast/src/hazelcast/client/metrics.cpp
```diff
@@ -0,0 +1,317 @@
...
+}
+
+int metrics_compressor::calculate_descriptor_mask(const metric_descriptor &descriptor) {
```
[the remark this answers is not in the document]
yemreinci: changed it to return byte

hazelcast/src/hazelcast/client/metrics.cpp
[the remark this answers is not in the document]
yemreinci: changed type to byte.

hazelcast/src/hazelcast/client/metrics.cpp
yemreinci: It's `int` in Java and 2 bytes are reserved to serialize it in the binary format. So theoretically it can be up to 2^16-1. Anyways, changed it to byte.

hazelcast/include/hazelcast/client/impl/metrics/metric_descriptor.h
yemreinci: I think there is "some" logic in this class. By making it a class and specifying defining ctors we enforce the ways how this type can be initialized. E.g.: if a descriptor name is given its value should also be given (it can't be null), the ctors are ensuring that.

yemreinci:
Note: In further discussions, we decided to go with implementing the level 0 compression of ZLIB ourselves and use it to avoid introducing dependencies.

yemreinci:
verify-linux

ihsandemir:
latest decision: Do not add dependency to zlib at all but use simple level-0 implementation of zlib compression.

ihsandemir (approved):
Can you please update the PR description with latest solution of no zlib dependency?

PR #898, opened 2021-07-07
Fix multiple definitions error

### LINKED ISSUE #895: C++ 17 vs C++ 11 incompatibility
C++ compiler version: 9.3.0
Hazelcast Cpp client version: 4.1.1
Hazelcast server version: 4.2
Number of the clients: 1
OS version (Windows/Linux/OSX): Ubuntu 20.04

#### Expected behaviour
Normally the Cpp-client is compiled with C++-11 version. With earlier versions it could then be used in C++17 code.

#### Actual behaviour
Now the compiling terminates with an error "multiple definition of hazelcast::client::imap::UNSET".

Change in examples/CMakeLists.txt the CMAKE_CXX_STANDARD 17
and try to compile with BUILD_EXAMPLES=ON and BUILD_SHARED_LIBS=OFF

yemreinci:
The reported issue is due to a bug in GCC or the linker. See the bug [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101957).

As a workaround, PR changes some `constexpr` usages with `const`. With this change, all examples compile in C++17 mode and link against a C++11-compiled client library without errors.

Fixes #895.

### DIFF
--- a/hazelcast/include/hazelcast/client/imap.h
+++ b/hazelcast/include/hazelcast/client/imap.h
@@ -1001,7 +1001,7 @@ namespace hazelcast {
             /**
              * Default TTL value of a record.
              */
-            static constexpr std::chrono::milliseconds UNSET{-1};
+            static const std::chrono::milliseconds UNSET;
 
             monitor::impl::LocalMapStatsImpl local_map_stats_;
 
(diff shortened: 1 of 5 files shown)


yemreinci:
verify-linux

Issue #337, opened 2017-10-24
SSL FIPS compliance: AWS auto discovery, context: asio.ssl error


aparnam-asg:
Hi,

 We have java application that uses JNI. This JNI code talks to Hazelcast using C++ client. On Amazon, we are using auto discovery using tag name/value. First time application seems to connect fine. But after that, I get the following error:
Error while connecting to cluster! =>context: asio.ssl error

I don't see anything else. Once the application fails, it never reconnects again. It seems like it works first time and that is it. Any idea on why this error happens and if there is anything else I can do to get more details on why ssl works once but fails after?

Thanks for your time

aparnam-asg:
This application is not reusing connection. It shuts down the hazelcast client after use till next request comes in, at which point it starts a fresh client.

aparnam-asg:
Will you be able to provide any insights on how to resolve this type of issue? I tried to reproduce this with just the relevant code in a smaller application but that one works in the exact same environment. I seem to have this issue only when the exact same code is run inside our product. If there is anyway to produce more logs on asio:ssl error, that should help immensely. Thanks for your help

ihsandemir:
Hello,

Are you sure that there is no other log information on the error. Yes, we use boost::asio library to connect https for discovering the AWS member information. I checked at the AWSClient code and here is the place where we throw and IOException: https://github.com/hazelcast/hazelcast-cpp-client/blob/master/hazelcast/src/hazelcast/util/SyncHttpClient.cpp#L100  Can you look again to see if you observe this exception line (assuming that you catch and log this exception)?

A reproducer code would help a lot.

aparnam-asg:
Thanks. Yes, there is nothing else in logs. I catch IException and log it. That just says "HazelcastClient could not be started!". I turned FINEST log level. And on console I see the other log that I mentioned before "Error while connecting to cluster! =>context: asio.ssl error". 

I know, at the minimum I need to provide reproducer code. Funny thing is when I try to extract out the code to a test program, that works, so as of now I can't reproduce it outside of our product. But I did have some progress. There is some other code in that execution path which sends ssl request to another server. And after that hazelcast stuff stops working. Somehow this other request seems to mess up AWS communication after. We enable fips in open ssl in our code. I am not sure if that is any issue. I am trying few more experiments on my end.

ihsandemir:
[...]

PR #908, opened 2021-08-06
Fix Openssl 1.1.x FIPS compatibility

### LINKED ISSUE #906: Won't compile with FIPS-enabled OpenSSL 1.1.x
Try to compile the project with openssl-devel on Fedora 34 to reproduce.

yemreinci:
A different implementation of `sha256_hashhex` that uses OpenSSL's EVP library was added for issue #337. However, the implementation only compiles when Openssl 1.0.x is used and not with 1.1.x as the EVP API was changed. This PR adds `#if` directives based on `OPENSSL_VERSION_NUMBER` to handle that. The EVP library seems to always exists, so I've also removed the implementation that uses SHA256_* calls. 

Also, the version check should be `OPENSSL_VERSION_NUMBER >= 0x10100000L` (not `>`) as the 1.1.x series start with 1.1.0.

I've also updated the dockerfiles to use the newest version of Fedora with OpenSSL 1.1.x installed.

Fixes #906.


docker/hazelcast-fedora-i386.dockerfile
```diff
@@ -1,6 +1,6 @@
-FROM fedora:33
+FROM fedora:34
```
ihsandemir, line 1:
```suggestion
FROM fedora:latest
```

docker/hazelcast-fedora-x86_64.dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM fedora:33
+FROM fedora:34
```
ihsandemir, line 1:
```suggestion
FROM fedora:latest
```

ihsandemir (approved):

Issue #899, opened 2021-07-09
std::hash<member> should hash uuid + address


ihsandemir:
See [Java member impl hashCode](https://github.com/hazelcast/hazelcast/blob/4.2.1/hazelcast/src/main/java/com/hazelcast/cluster/impl/AbstractMember.java#L212). We also need to do the same.

ihsandemir:
fixed by https://github.com/hazelcast/hazelcast-cpp-client/pull/909

PR #909, opened 2021-08-06
[API-597] Implementation external public address usage support

ihsandemir:
Initial implementation of external public address usage support for the C++ client. This feature is needed for external Kubernetes client to be able to connect to the Hazelcast cluster.

related to https://hazelcast.atlassian.net/browse/API-597

Verification:[ Sample kubernetes test run](https://github.com/alisengul53/GHActionTestRepo/runs/3289096937?check_suite_focus=true)

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/899

### DIFF
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -1206,7 +1206,7 @@ In order to use this feature, make sure your cluster members are accessible from
 
 ```c++
     hazelcast::client::client_config config;
-    constexpr const int32_t server_port = 5701;
+    constexpr int server_port = 5701;
     constexpr const char *server_public_address = "myserver.publicaddress.com";
     config.get_network_config().use_public_address(true).add_address(
             hazelcast::client::address{server_public_address, server_port});
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -1198,11 +1198,20 @@ Please check the code sample at `examples/cloud-discovery/ssl-connect-cloud.cpp`
 
 > **NOTE: This feature requires Hazelcast IMDG 4.2 or higher version.**
 
-The client sends requests directly to cluster members in the smart client mode (default) in order to reduce hops to accomplish operations. Because of that, the client should know the addresses of members in the cluster.
+The client sends requests directly to cluster members in the smart client mode (default) in order to reduce hops to 
+accomplish operations. Because of that, the client should know the addresses of members in the cluster.
 
-In cloud-like environments, or Kubernetes, there are usually two network interfaces, the private and the public network. When the client is in the same network as the members, it uses their private network addresses. Otherwise, if the client and the Hazelcast cluster are on different networks, the client cannot connect to members using their private network addresses. Hazelcast 4.2 introduced External Smart Client Discovery to solve that issue. The client needs to communicate with all cluster members via their public IP addresses in this case. Whenever Hazelcast cluster members are able to resolve their own public external IP addresses, they pass this information to the client. As a result, the client can use public addresses for communication.
+In cloud-like environments, or Kubernetes, there are usually two network interfaces: the private and public network 
+interfaces. When the client is in the same network as the members, it uses their private network addresses. Otherwise, 
+if the client and the Hazelcast cluster are on different networks, the client cannot connect to members using their 
+private network addresses. Hazelcast 4.2 introduced External Smart Client Discovery to solve that issue. The client 
+needs to communicate with all cluster members via their public IP addresses in this case. Whenever Hazelcast cluster 
+members are able to resolve their own public external IP addresses, they pass this information to the client. 
+As a result, the client can use public addresses for communication.
 
-In order to use this feature, make sure your cluster members are accessible from the network the client resides in, then set config `client_network_config()::use_public_address(true)` to true. You should specify the public address of at least one member in the configuration:
+In order to use this feature, make sure your cluster members are accessible from the network the client resides in, 
+then set config `client_network_config()::use_public_address(true)` to true. You should specify the public address of 
+at least one member in the configuration:
 
 ```c++
     hazelcast::client::client_config config;
@@ -1212,7 +1221,8 @@ In order to use this feature, make sure your cluster members are accessible from
             hazelcast::client::address{server_public_address, server_port});
 ```
 
-This solution works everywhere without further configuration: Kubernetes, AWS, GCP, Azure, etc. as long as the corresponding plugin is enabled in Hazelcast server configuration.
+This solution works everywhere without further configuration (Kubernetes, AWS, GCP, Azure, etc.) as long as the 
+corresponding plugin is enabled in Hazelcast server configuration.
 
 ## 5.10. Authentication
 
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -1196,9 +1196,11 @@ Please check the code sample at `examples/cloud-discovery/ssl-connect-cloud.cpp`
 
 ## 5.9. External Smart Client Discovery
 
+> **NOTE: This feature requires Hazelcast IMDG 4.2 or higher version.**
+
 The client sends requests directly to cluster members in the smart client mode (default) in order to reduce hops to accomplish operations. Because of that, the client should know the addresses of members in the cluster.
 
-In cloud-like environments, or Kubernetes, there are usually two network interfaces, the private and the public network. When the client is in the same network as the members, it uses their private network addresses. Otherwise, if the client and the Hazelcast cluster are on different networks, the client cannot connect to members using their private network addresses. Hazelcast 4.2 introduced External Smart Client Discovery to solve that issue.
+In cloud-like environments, or Kubernetes, there are usually two network interfaces, the private and the public network. When the client is in the same network as the members, it uses their private network addresses. Otherwise, if the client and the Hazelcast cluster are on different networks, the client cannot connect to members using their private network addresses. Hazelcast 4.2 introduced External Smart Client Discovery to solve that issue. The client needs to communicate with all cluster members via their public IP addresses in this case. Whenever Hazelcast cluster members are able to resolve their own public external IP addresses, they pass this information to the client. As a result, the client can use public addresses for communication.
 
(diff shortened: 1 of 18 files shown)


ihsandemir:
verify-windows

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -840,7 +842,22 @@ namespace hazelcast {
...
+                            auto found = address_map.find(CLIENT);
+                            address member_address;
+                            if (found != address_map.end()) {
+                                member_address = found->second;
+                            } else {
+                                found = address_map.find(MEMBER);
+                                assert(found != address_map.end());
+                                member_address = found->second;
+                            }
+                            member new_member(member_address, m.get_uuid(), m.is_lite_member(), m.get_attributes(), m.address_map());
```
yemreinci, line 858: I realize there isn't a check for the `client_network_config::use_public_address` configuration here. So, the address of a member that is passed to a membership event handler can change with the new implementation. Isn't this a breaking change? There is already an address translation logic implemented in `ClientConnectionManagerImpl::translate`. Why is this also necessary?

hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -1763,6 +1764,99 @@ namespace hazelcast {
...
             }
+
+            class connection_manager_translate : public ClientTestSupport {
+            public:
+                static const address private_address;
+                static const address public_address;
+
+                class test_address_provider : public connection::AddressProvider {
+                    public:
+                        test_address_provider(bool shouldTranslate) : should_translate(shouldTranslate) {}
```
yemreinci, line 1775:
```suggestion
                        test_address_provider(bool should_translate) : should_translate_(should_translate) {}
```

None of the tests below contruct this class with should_translate as false BTW.

hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -1763,6 +1764,99 @@ namespace hazelcast {
...
+                config.get_network_config().use_public_address(false);
+                config.get_connection_strategy_config().get_retry_config().set_cluster_connect_timeout(
+                        std::chrono::seconds(1));
+                auto client = new_client(std::move(config)).get();
+                spi::ClientContext ctx(client);
+
+                member m(public_address, ctx.get_hazelcast_client_implementation()->random_uuid(), false, {},
+                         {{endpoint_qualifier{1, "public"}, address{"127.0.0.1", 5701}}});
+
+                ASSERT_THROW(ctx.get_connection_manager().get_or_connect(m), std::exception);
```
yemreinci, line 1854: Is there a more specific exception type that is expected here? I think `std::exception` is too general.

yemreinci (changes requested):

Reference_Manual.md
```diff
@@ -1190,7 +1194,27 @@ As you see in the code snippet, we provide a `boost::asio::ssl::context` in the
...
+
+In cloud-like environments, or Kubernetes, there are usually two network interfaces, the private and the public network. When the client is in the same network as the members, it uses their private network addresses. Otherwise, if the client and the Hazelcast cluster are on different networks, the client cannot connect to members using their private network addresses. Hazelcast 4.2 introduced External Smart Client Discovery to solve that issue. The client needs to communicate with all cluster members via their public IP addresses in this case. Whenever Hazelcast cluster members are able to resolve their own public external IP addresses, they pass this information to the client. As a result, the client can use public addresses for communication.
+
+In order to use this feature, make sure your cluster members are accessible from the network the client resides in, then set config `client_network_config()::use_public_address(true)` to true. You should specify the public address of at least one member in the configuration:
+
+```c++
+    hazelcast::client::client_config config;
+    constexpr const int32_t server_port = 5701;
```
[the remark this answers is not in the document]
ihsandemir: There was one other place similar in this file and i changed that as well.

hazelcast/src/hazelcast/client/cluster.cpp
```diff
@@ -234,12 +244,28 @@ namespace hazelcast {
...
 
 namespace std {
-    std::size_t hash<hazelcast::client::member>::operator()(const hazelcast::client::member &k) const noexcept {
-        return boost::hash<boost::uuids::uuid>()(k.get_uuid());
+    std::size_t hash<hazelcast::client::member>::operator()(const hazelcast::client::member &m) const noexcept {
+        std::size_t seed = 0;
+        boost::hash_combine(seed, std::hash<hazelcast::client::address>()(m.get_address()));
+        boost::hash_combine(seed, m.get_uuid());
+        return seed;
+    }
```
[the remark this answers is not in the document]
ihsandemir: fixed.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -767,6 +740,18 @@ namespace hazelcast {
...
+            address ClientConnectionManagerImpl::translate(const member &m) {
+                if (use_public_address_) {
+                    auto public_addr_it = m.address_map().find(PUBLIC_ENDPOINT_QUALIFIER);
+                    if (public_addr_it != m.address_map().end()) {
+                        return public_addr_it->second;
+                    }
+                    return m.get_address();
+                }
+
+                return *address_provider_->translate(m.get_address());
```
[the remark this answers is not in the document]
ihsandemir: They are exclusive. The Java side has [this line](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/TranslateToPublicAddressProvider.java#L66) for the check. I added a method `bool is_default_provider()` to the `AddressProvider` interface so that we can do this check while constructing the connection manager and i set the `use_public_address_` bool flag according to it to be consistent with Java.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -840,7 +842,22 @@ namespace hazelcast {
...
+                        if (address_map.empty()) {
+                            result.members.insert({m.get_uuid(), m});
+                        } else {
+                            auto found = address_map.find(CLIENT);
+                            address member_address;
+                            if (found != address_map.end()) {
+                                member_address = found->second;
+                            } else {
+                                found = address_map.find(MEMBER);
+                                assert(found != address_map.end());
```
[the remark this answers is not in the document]
ihsandemir: Based on the [Java implementation ](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/client/impl/spi/impl/ClientClusterServiceImpl.java#L282) and [this line](https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/cluster/impl/MemberImpl.java#L214), I thought that it should exists and hence I put the assertion here. Is the assertion line not self describing, do we still need a comment line even with this assertion line?

hazelcast/src/hazelcast/client/spi.cpp
ihsandemir: Our implementation follows the Java client reference impl. I believe the logic is this: If the use_public_address is not configured (i.e. it is false), use the address coming in the address_map if address_map is sent from the server. It is given higher priority than the server `address` field inside the member_info. Hence, if there is an address_map, the address field is ignored, and this is done regardless of the `use_public_address` configuration.

hazelcast/test/src/HazelcastTests8.cpp
ihsandemir: Removed the unused parameter. It was initially there to include more test cases but I later on decided that i will not have those test cases.

hazelcast/test/src/HazelcastTests8.cpp
ihsandemir: We receive boost::system_error with error_code `boost::asio::error::network_unreachable` on socket connect. This exception do not normally propagate to the user but we always wrap it for user API calls. But in this test scenario, we access an internal API and test it and hence we internally receive a boost exception. I changed to test for this specific error_code hoping that it will work for windows as well :)

Reference_Manual.md
```diff
@@ -1190,7 +1194,27 @@ As you see in the code snippet, we provide a `boost::asio::ssl::context` in the
...
 
-## 5.9. Authentication
+## 5.9. External Smart Client Discovery
+
+> **NOTE: This feature requires Hazelcast IMDG 4.2 or higher version.**
+
+The client sends requests directly to cluster members in the smart client mode (default) in order to reduce hops to accomplish operations. Because of that, the client should know the addresses of members in the cluster.
+
+In cloud-like environments, or Kubernetes, there are usually two network interfaces, the private and the public network. When the client is in the same network as the members, it uses their private network addresses. Otherwise, if the client and the Hazelcast cluster are on different networks, the client cannot connect to members using their private network addresses. Hazelcast 4.2 introduced External Smart Client Discovery to solve that issue. The client needs to communicate with all cluster members via their public IP addresses in this case. Whenever Hazelcast cluster members are able to resolve their own public external IP addresses, they pass this information to the client. As a result, the client can use public addresses for communication.
```
Serdaro, line 1203: ... network interfaces: private and public...

Reference_Manual.md
```diff
@@ -1190,7 +1194,27 @@ As you see in the code snippet, we provide a `boost::asio::ssl::context` in the
...
+
+```c++
+    hazelcast::client::client_config config;
+    constexpr int server_port = 5701;
+    constexpr const char *server_public_address = "myserver.publicaddress.com";
+    config.get_network_config().use_public_address(true).add_address(
+            hazelcast::client::address{server_public_address, server_port});
+```
+
+This solution works everywhere without further configuration: Kubernetes, AWS, GCP, Azure, etc. as long as the corresponding plugin is enabled in Hazelcast server configuration.
```
Serdaro, line 1215: ... configuration (Kubernetes, AWS, GCP, Azure, etc.) as long as ...

Serdaro (commented):
Tiny edits.

Reference_Manual.md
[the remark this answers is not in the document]
yemreinci: There are more though.

yemreinci (dismissed):

Serdaro (approved):

Issue #900, opened 2021-07-13
Lockup in library when client shuts down


andysCaplin:
C++ compiler version: 
Hazelcast Cpp client version:  4.0.1
Hazelcast server version: N/A
Number of the clients:
Cluster size, i.e. the number of Hazelcast cluster members:
OS version (Windows/Linux/OSX): Linux

Please attach relevant logs and files for client and server side.

#### Expected behaviour
When a SIGTERM is handled the binary should shut down

#### Actual behaviour
When the SIGTERM is received the Hazelcast library gets stuck sometimes which prevents the binary shutdown
We intercept SIGTERM and write to a pipe to call our specific signal handling code - i.e. none of our signal handling code, including the call to Hazelcast shutdown is called in the signal handler itself.

You can see that in thread 41 of the attached file. I don't know why frame 3 isn't shown but it's a call to a routine in our code that calls Hazelcast shutdown. When that returns we exit.
I can see from the logs that we've called the Hazelcast shutdown
```
2021/07/09-22:49:19.542328 +0000: (rttpd): NOTIFY: Discovery: Shutting down client
2021/07/09-22:49:19.542349 +0000: (rttpd): NOTIFY: Discovery: Peer - Removing entry <ccbbfb16-e5be-4c1a-af5b-840492060004>
2021/07/09-22:49:19.542465 +0000: (4276): INFO: Discovery: Library - <hz.client_1><soaktest> LifecycleService::LifecycleEvent SHUTTING_DOWN
2021/07/09-22:49:19.542631 +0000: (4276): INFO: Discovery: Library - <hz.client_1><soaktest> Removed connection to endpoint: Address[10.42.3.224:5701], connection: ClientConnection{alive=0, connectionId=3, remoteEndpoint=Address[10.42.3.224:5701], lastReadTime=2021-07-09 22:49:19.0-5, closedTime=2021-07-09 22:49:19.000, connected server version=4.0.3}
2021/07/09-22:49:19.542676 +0000: (4276): INFO: Discovery: Library - <hz.client_1><soaktest> Removed connection to endpoint: Address[10.42.3.226:5701], connection: ClientConnection{alive=0, connectionId=1, remoteEndpoint=Address[10.42.3.226:5701], lastReadTime=2021-07-09 22:49:19.0-5, closedTime=2021-07-09 22:49:19.000, connected server version=4.0.3}
[...]

PR #916, opened 2021-08-20
Do not re-schedule backup event timer if connection is closing and client is shutting down

ihsandemir:
Added an extra check for re-scheduled backup event timer so that it is not being re-scheduled when the client being shutdown. Failing to do so causes the shutdown to hang up. Possible scenario for generating the problem:

1. Trigger client shutdown.
2. The backup timer expired and its lambda code is being executed.
3. The connection close cancels the timer but it is not effective since the time is not in the wait queue.
4. The timer is being rescheduled.

Also, checked all the other timers, they all have similar checks. Changed the location of the check for the `ClientConnectionManagerImpl::schedule_connect_to_all_members` method.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/900

Please see reproducer test at [reproducer branch](https://github.com/ihsandemir/hazelcast-cpp-client/tree/issue_900_master/examples/issue_900).

### DIFF
--- a/hazelcast/src/hazelcast/client/network.cpp
+++ b/hazelcast/src/hazelcast/client/network.cpp
@@ -132,16 +132,16 @@ namespace hazelcast {
             }
 
             void ClientConnectionManagerImpl::schedule_connect_to_all_members() {
+                if (!client_.get_lifecycle_service().is_running()) {
+                    return;
+                }
+
                 connect_to_members_timer_->expires_from_now(boost::asio::chrono::seconds(1));
                 connect_to_members_timer_->async_wait([=](boost::system::error_code ec) {
                     if (ec == boost::asio::error::operation_aborted) {
                         return;
                     }
                     connect_to_all_members();
-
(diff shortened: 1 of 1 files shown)


yemreinci (approved):

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -132,17 +132,17 @@ namespace hazelcast {
             }
 
             void ClientConnectionManagerImpl::schedule_connect_to_all_members() {
+                if (!client_.get_lifecycle_service().is_running()) {
+                    return;
+                }
+
```
yemreinci, line 138: This change doesn't actually affect anything, right? I assume it's just for consistency with the other one.

ihsandemir: yes, it doesn't change anything.

ihsandemir:
verify

PR #918, opened 2021-08-20
Fix Windows GHA builds

yemreinci:
The default user agent of `Invoke-WebRequest` in PowerShell is `Mozilla` and JFrog started serving HTML pages for download requests made with browser agents. Passing an empty user agent makes it work as it used be.

See [the run](https://github.com/yemreinci/hazelcast-cpp-client/actions/runs/1151372489) from my fork.


yemreinci:
verify

ihsandemir (approved):

PR #923, opened 2021-08-25
Fix missing .get() in test

yemreinci:
I had [this](https://github.com/yemreinci/hazelcast-cpp-client/runs/3416772846?check_suite_focus=true) ASAN error report on a GHA run on my fork. (I have modified the workflow to use ASAN). 

Apparently, there can be future handlers still running even though the client object was destroyed and those handlers can access destroyed objects. 

This PR doesn't solve the heap-use-after-free but only fixes a trivial mistake in the test.


ihsandemir (approved):
We need to think about the lifecycle of the client and such async invocations to solve the root cause. An idea: we can keep a shared_ptr to client in future retuning `SerializingProxy::to_object` method lambda, but need to check if any cyclic dependency.

PR #929, opened 2021-09-16
Fix init order of default log handler's lock

yemreinci:


yemreinci:
I tried something in this PR but it didn't work. Closing.

PR #930, opened 2021-09-17
Initial draft POC implementation of C API to be used by Lua language

ihsandemir:
Initial draft POC implementation of C API to be used by Lua language.

Use example in lua_example directory after [installing the library](https://github.com/hazelcast/hazelcast-cpp-client/blob/v4.2.0/Reference_Manual.md#113-install-from-source-code-using-cmake) from source code.

Run the example as `lua hello.lua`. Before running the example, create a symbolic link named `hazelcast.so` in the directory to the installed `libhazelcast-cpp-client.so` file.

Prerequisites:
1. Install lua
2. Install boost library (see [installing the library](https://github.com/hazelcast/hazelcast-cpp-client/blob/v4.2.0/Reference_Manual.md#113-install-from-source-code-using-cmake) )


ihsandemir:
Closing this PR, since we decided to continue with a different solution which works async with co-socket API.

PR #933, opened 2021-09-22
Add vcpkg manifest mode usage (vcpkg.json)

ihsandemir:
Updated the documentation to detail how the user can work with [vcpkg manifest mode](https://github.com/microsoft/vcpkg/blob/master/docs/users/manifests.md) (using application specific vcpkg.json).


ihsandemir:
closing for now. To be done int he future.

Issue #874, opened 2021-05-11
Tests should fail if client cannot connect 


yemreinci:
Since we have merged #767, if a the client instance in a test can't connect to the cluster for some reason, the test hangs indefinitely. This makes the development hard and causes long-running CI jobs (e.g. [this ](https://github.com/yemreinci/hazelcast-cpp-client/runs/2559505596?check_suite_focus=true)happened when I was working on windows GA builds.). 

Tests should simply fail if the client couldn't connect for some time.

ihsandemir:
Proposal: mWe can use the `ClientTestSupportBase::get_config` in all tests where we instantiate client_config and build on top of it. We can set the cluster connect timeout to a lower value suitable for the tests in this factory class.

PR #939, opened 2021-10-17
Run tests without RC

yemreinci:
Depends on #938 

Currently, the `client_test` executable is useless if RC is not running. However, we have tests that don't necessarily require RC (e.g. pure unit tests) that we may want to run during development. Also, googletest utilities such as `--gtest_list_tests` don't need RC to produce output. 

This PR allows `client_test` to run tests even when an RC connection wasn't made. The test cases that require RC will fail individually when RC wasn't started. For example:
```
-> ./build-ssl/hazelcast/test/src/client_test --gtest_filter=ClientMultiMapTest.testPutGetRemove
Running main() from /home/yunus/.local/src/googletest-release-1.11.0/googletest/src/gtest_main.cc
Note: Google Test filter = ClientMultiMapTest.testPutGetRemove
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ClientMultiMapTest
Thrift: Mon Oct 18 11:11:29 2021 TSocket::open() connect() <Host: 127.0.0.1 Port: 9701>: Connection refused
unknown file: Failure
C++ exception with description "connect() failed: Connection refused" thrown in SetUpTestSuite().
[ RUN      ] ClientMultiMapTest.testPutGetRemove
Thrift: Mon Oct 18 11:11:29 2021 TSocket::open() connect() <Host: 127.0.0.1 Port: 9701>: Connection refused
unknown file: Failure
C++ exception with description "connect() failed: Connection refused" thrown in the test fixture's constructor.
[  FAILED  ] ClientMultiMapTest.testPutGetRemove (0 ms)
[----------] 1 test from ClientMultiMapTest (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ClientMultiMapTest.testPutGetRemove

 1 FAILED TEST
[  FAILED  ] ClientMultiMapTest: SetUpTestSuite or TearDownTestSuite

 1 FAILED TEST SUITE
```
Previously, the `client_test` executable quit immediately after realizing RC wasn't started, now it will run the test but they will fail.

Summary of changes:
  1. Removed custom googletest main (`test/src/main.cpp`) as it wasn't necessary anymore. Added `gtest_main` as a dependency in CMake.
  2. Moved content of ClientTestSupportBase to ClientTestSupport. Renamed `ClientTestSupport` to `ClientTest`.
  3. Converted RC client instance to a Meyer's singleton, accessible with `remote_controller_client()`.
  4. Converted g_srvFactory to a singleton, accessible with `ClientTest::default_server_factory()`.
  5. `TEST_F` uses that don't actually require the associated fixture to work are converted to `TEST`. Example test names are: `AwsConfigTest` and `ssl_config_test`
  6. Some test fixtured derived from `ClientTestSupport` although they didn't require any client or HZ instances. e.g: CallIdSequenceWithoutBackpressureTest. These test fixtures updated to derive from `::testing::Test`.


hazelcast/test/src/remote_controller_client.cpp
```diff
@@ -0,0 +1,55 @@
...
+    using namespace apache::thrift::protocol;
+    using namespace apache::thrift::transport;
+
+    static auto socket = std::make_shared<TSocket>(remote_controller_address(), 9701);
+    static auto transport = std::make_shared<TBufferedTransport>(socket);
+    static auto protocol = std::make_shared<TBinaryProtocol>(transport);
+
+    static remote::RemoteControllerClient client{ protocol };
+
+    transport->open();
```
sancar, line 48: Isn't this `open` called every call to the remote_controller client ? And how it is not a problem ?

hazelcast/test/src/ClientTest.cpp
```diff
@@ -0,0 +1,148 @@
...
+        config.set_cluster_name(get_ssl_cluster_name());
+        boost::asio::ssl::context ctx(boost::asio::ssl::context::method::tlsv12_client);
+        ctx.set_default_verify_paths();
+        ctx.load_verify_file(get_ca_file_path());
+
+        config.get_network_config().get_ssl_config().set_context(std::move(ctx));
+    }
+#endif
+
+    config.get_connection_strategy_config().get_retry_config().set_cluster_connect_timeout(
```
sancar, line 81: Why do we set default timeout here ?

hazelcast/test/src/remote_controller_client.cpp
yemreinci: Good question, I've missed this. It's not a problem because it ignores the call if it was already opened.

The better way of doing it would be wrapping TSocket or TTransport in a RAII type that will do the ::open call during construction. I'll think about this.

https://github.com/apache/thrift/blob/98be76fc033f1d66bcfd09d4a22b86e8061e89c2/lib/cpp/src/thrift/transport/TSocket.cpp#L411-L420

hazelcast/test/src/ClientTest.cpp
yemreinci: https://github.com/hazelcast/hazelcast-cpp-client/issues/874#issuecomment-839568694

sancar: Ok. I leave the decision to you guys. 
Additional note, 5 seconds seems too low to me. I would set it something like a minute to avoid false positives.

sancar (dismissed):

hazelcast/test/src/remote_controller_client.cpp
yemreinci: I've changed it to

```
    // if this is the first time this function is called,
    // open the transport.
    if (!transport->isOpen()) {
        transport->open();
    }
```

Now it doesn't call open after the first time (it would be fine if it did as I said earlier, but I still put the if there to make it clear to the reader)

hazelcast/test/src/ClientTest.cpp
yemreinci: >  Additional note, 5 seconds seems too low to me. I would set it something like a minute to avoid false positives.

Made it 20. If nightly tests starts failing I'll increase it even more.

yemreinci:
@sancar I've addressed your comments, can you please re-approve?

@ihsandemir Can you please review?

hazelcast/test/src/ClientTest.cpp
ihsandemir: We have 120secs for assert eventually checks, may be we can have the same timeout for here as well. Or you can wait until it fails. It will be more visible when the compatibility tests run which usually cause timing problems and we may not apply the fix for backwards, hence I would take the safe path and put a higher boundary.

hazelcast/test/src/HazelcastServerFactory.h
```diff
@@ -57,14 +48,10 @@ namespace hazelcast {
 
             private:
                 std::shared_ptr<logger> logger_;
-                std::string server_address_;
```
ihsandemir, line 60: The initial aim for this address was that we could have started the rc on another machine, later on we seem to not use it much. Do we have such needs at all?

yemreinci: This variable is always set to `g_srvFactory->get_server_address()` so I removed it. All of its uses are replaced with `remote_controller_address()`. 

Current codebase doesn't support running RC on anywhere other than localhost because its address is statically set to 127.0.0.1 [here](https://github.com/hazelcast/hazelcast-cpp-client/blob/f3991d8663eaa61ec1d38de90c1aae18a8fd2287/hazelcast/test/src/main.cpp#L51). I have extracted this value and put it in the global function `remote_controller_address()` and left a TODO there saying this value can be read from a env variable or a command line argument some day to support this feature. But even then, it would require extra refactoring. Because some tests specifically expect the member to be on 127.0.0.1. (see [this](https://github.com/hazelcast/hazelcast-cpp-client/blob/f3991d8663eaa61ec1d38de90c1aae18a8fd2287/hazelcast/test/src/HazelcastTests1.cpp#L160) for example, it doesn't do any networking config, so the default 127.0.0.1 is used. If the RC is on a different host than tests using this function will fail.)

hazelcast/test/src/ClientTest.cpp
yemreinci: it is 120 now. c74ffb76cd4e109fd879ab7718fa6a84b2e6eb1a

yemreinci:
verify-windows

ihsandemir (approved):

sancar (approved):

PR #945, opened 2021-12-17
Remove redundant `strerror_s` and `tokenize_version_string`

### LINKED ISSUE #942: Doesn't compile with musl [API-1026]
Compiling on a linux system with musl libc (e.g. Alpine) produce the following:

```
/data/hazelcast-cpp-client-4.2.0/hazelcast/src/hazelcast/util/util.cpp: In function 'int hazelcast::util::strerror_s(int, char*, size_t, const char*)':
/data/hazelcast-cpp-client-4.2.0/hazelcast/src/hazelcast/util/util.cpp:417:40: error: invalid conversion from 'int' to 'char*' [-fpermissive]
  417 |             char *errStr = ::strerror_r(errnum, strerrbuf + numChars, buflen - numChars);
      |                            ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                        |
      |                                        int
```
[
musl's strerror_r](http://git.musl-libc.org/cgit/musl/tree/src/string/strerror_r.c) returns `int`, not `char*`.

[See](https://linux.die.net/man/3/strerror_r) for more info.

The fix is just to remove `hazelcast::util::strerror_s`, it's not used anywhere other than tests.

yemreinci:
Fixes #942.

### DIFF
--- a/hazelcast/include/hazelcast/util/Util.h
+++ b/hazelcast/include/hazelcast/util/Util.h
@@ -64,12 +64,6 @@ namespace hazelcast {
          */
         HAZELCAST_API int64_t current_time_nanos();
 
-        /**
-         * @return 0 if error string could be obtained, non-zero otherwise
-         */
-        HAZELCAST_API int
-        strerror_s(int errnum, char *strerrbuf, size_t buflen, const char *msg_prefix = (const char *) NULL);
-
         int32_t HAZELCAST_API get_available_core_count();
 
         template <typename T>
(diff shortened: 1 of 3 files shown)


mdumandag (approved):

ihsandemir (approved):
Is method [`sleepmillis`](https://github.com/hazelcast/hazelcast-cpp-client/blob/master/hazelcast/include/hazelcast/util/Util.h#L38) also unused? If so, lets remove that as well.

PR #946, opened 2022-01-06
Add .clang-format, Format files

yemreinci:
I've added a .clang-format file and applied it on every .cpp and .h file. 

The first commit 1165b18 adds the .clang-format file, which I generated using the following command:
```
clang-format -style="{BasedOnStyle: Mozilla, FixNamespaceComments: true,  IndentWidth: 4, AccessModifierOffset: -4, SortIncludes: false}" -dump-config > .clang-format
```
I used Clang Format version 13.0.0. I customized 3 options of the Mozilla preset:
   1. `FixNamespaceComments: true`: This makes the clang-format to add a comment after the namespace closing bracket like this:
       ```
       namespace foo {
       // here be code
       } // namespace foo
      ```
      This is useful as it can help you see which namespace are you in at some line in the code.
   2. `IndentWidth: 4, AccessModifierOffset: -4` : The Mozilla preset by default has an indent width of 2. I have changed it to 4 based purely on personal preference. But I don't have a strong opinion on this.
   3. `SortIncludes: false` This sorts the includes alphabetically and should ideally be set to true. However, a lot of header files in the codebase don't follow the "Include What You Use" concept and heavily depend on transitive includes. And if we sort them, the code doesn't compile because of missing includes. So I am setting this to false temporarily as it would require some manual refactoring and time to get it right. 

The second commit 1567490 applies clang-format on every file, using the following one-liner:
```
find hazelcast examples | grep -E '(cpp|h)$' | xargs clang-format -i
```

Both commits don't contain any manual refactoring or change, all changes are automatically generated by ClangFormat. So a quick skim through the files instead of a detailed review will suffice.

I can customize more configuration options if we don't like any specific formatting.


sancar:
Can we add the formatting one liner as a bash script to the project somewhere ?

sancar (dismissed):

yemreinci:
> Can we add the formatting one liner as a bash script to the project somewhere ?

@sancar, Is this so that we can run it before committing our changes?

sancar:
@yemreinci It could be, especially for the ones that want to run this outside of the IDE.

yemreinci:
> @yemreinci It could be, especially for the ones that want to run this outside of the IDE.

One should probably use `git-clang-format` for this purpose. It is a python script that runs clang-format on a set of commits or your staged but not yet committed changes. It is included in  the `clang-format` package on ubuntu apt repositories for ex.

Still, added the command as a script, `script/format-all.sh`. See 8287006

sancar (dismissed):

hazelcast/include/hazelcast/client/imap.h
```diff
@@ -37,1132 +37,1381 @@
...
+#pragma warning(disable : 4251) // for dll export
 #endif
 
 namespace hazelcast {
-    namespace client {
-        class hazelcast_client;
+namespace client {
+class hazelcast_client;
 
-        namespace spi {
```
ihsandemir: Do we really want namepsace and class definition alignments all to left with no indentation? By the way, I was hoping that you auto-generated the namespace closing comment (as you put in the PR description) but I can not see them here.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -64,1271 +64,1639 @@
...
+    CLIENT,
+    "public"
+};
+
+ClientConnectionManagerImpl::ClientConnectionManagerImpl(
+  spi::ClientContext& client,
+  std::unique_ptr<AddressProvider> address_provider)
+  : alive_(false)
+  , logger_(client.get_logger())
+  , connection_timeout_millis_((std::chrono::milliseconds::max)())
```
ihsandemir: do we really want these `,` characters at the beginning?

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -64,1271 +64,1639 @@
...
+        auto cluster_id_changed = !current_cluster_id.is_nil() &&
+                                  !(new_cluster_id == current_cluster_id);
+        if (cluster_id_changed) {
+            HZ_LOG(
+              logger_,
+              warning,
+              boost::str(
+                boost::format(
+                  "Switching from current cluster: %1%  to new cluster: %2%") %
+                current_cluster_id % new_cluster_id));
```
ihsandemir: Do we really want this log line span these lines?

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -64,1271 +64,1639 @@
...
-                    }
-                    auto connection = get_connection(member->get_uuid());
-                    if (connection) {
-                        return connection;
-                    }
-                }
+        auto local_address = connection->get_local_socket_address();
+        if (local_address) {
+            HZ_LOG(
+              logger_,
```
ihsandemir: Is there a problem with MACRO formatting?

hazelcast/include/hazelcast/client/imap.h
yemreinci: > Do we really want namepsace and class definition alignments all to left with no indentation? 

I think, yes. Because most code in a file generally lives in the same namespace, and adding an indent for every namespace is wasting valuable space on the left side of the code. This makes lines long and hard to read.

> By the way, I was hoping that you auto-generated the namespace closing comment (as you put in the PR description) but I can not see them here.

If the namespace's content is only one line it doesn't generate a closing comment, which is the case for this particular line. There is a new parameter in clang-format 14 called ShortNamespaceLines which allows you to control exactly this but it's very new an might not available on everywhere yet - as a matter of fact LLVM 14 was released today -. Anyway, I think this should not be important, it won't put comments if there is only one line.

hazelcast/src/hazelcast/client/network.cpp
yemreinci: What do you suggest?

hazelcast/src/hazelcast/client/network.cpp
yemreinci: If it is longer than 80 characters, yes.

hazelcast/src/hazelcast/client/network.cpp
yemreinci: This the Mozilla's default and personally, I like it. Every member is on a different line, so it is easier the see what are the members and how are they being initialized. The commas at the beginning makes everything nicely aligned with the initial line starting with a colon. I agree that it might look a bit weird at first but once you've seen a few, it's clear.

Please note that you don't have to write your code like this. You just write however you like and press some shortcut on your IDE and it will make it this way.

hazelcast/src/hazelcast/client/network.cpp
yemreinci: I don't know what do you mean by problem? Is it inconsistent? The way it works is that if a function call is too long to fit into the column limit, it puts every parameter into a different line.

yemreinci:
Since I've merged a few PR's while this was still here, I had to rebase and rerun the formatting script. We still have 3 commits:
1. 939ac5b29efd66c05f9a1cdb2b266d33d2f553ad adds the .clang-format file.
2. b9ebd33cfcd4860b642f6ad0bdb4e44d2bd47377 adds the format-all.sh script.
3. bdf8352dcd5cc1442d69fcf2e370acd99b9b8380 is the huge commit changing to the new format, prepared automatically by running the format-all.sh script.

ihsandemir (approved):

sancar (approved):

PR #948, opened 2022-01-13
run toObject with sync launch policy

sancar:


yemreinci (approved):
I think we should merge this regardless of Compact.

ihsandemir (approved):

Issue #418, opened 2018-05-23
Azure auto discovery


aparnam-asg:
C++ client doesn't seem to support auto discovery on Azure, which is a requirement to run our application on Azure.

PR #949, opened 2022-01-14
Fix macOS builds

yemreinci:
Homebrew started to install OpenSSL 3 by default. And the Mac builds fail because some functions the project is using was deprecated. ([see](https://github.com/hazelcast/hazelcast-cpp-client/runs/4810898840?check_suite_focus=true))

We should ideally replace deprecated methods, but this PR pins the version to 1.1 to make the builds green for now.

### DIFF
--- a/.github/workflows/nightly-macos-x86_64.yml
+++ b/.github/workflows/nightly-macos-x86_64.yml
@@ -40,7 +40,7 @@ jobs:
     steps:
       - name: Install Dependencies
         run: |
-          brew install boost openssl thrift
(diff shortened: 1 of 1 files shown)


ihsandemir (approved):
Is there an issue we opened that we can now completely remove the aws discovery since the member reports the public private pairs? Do you think that we can remove those codes for aws discovery, or are they still needed? By the way, regarding [this](https://github.com/hazelcast/hazelcast-cpp-client/issues/418) issue, with public private ip pairs from the server, we may have it already working.

mdumandag (approved):

PR #952, opened 2022-01-20
Hide to_objects without future parameter

sancar:
This is the prepartion pr for compact.
We will have special kind of handling on the future continuations
for compact schemas. to_object without a future should not be
called from any proxy.

### DIFF
--- a/hazelcast/include/hazelcast/client/imap.h
+++ b/hazelcast/include/hazelcast/client/imap.h
@@ -738,7 +738,7 @@ public:
       query::paging_predicate<K, V>& predicate)
     {
         predicate.set_iteration_type(query::iteration_type::KEY);
-        auto futures =
+        auto future =
           key_set_for_paging_predicate_data(
             protocol::codec::holder::paging_predicate_holder::of(
               predicate, serialization_service_))
@@ -751,7 +751,7 @@ public:
                   predicate.set_anchor_data_list(std::move(result.second));
                   return result.first;
               });
-        return to_object_vector<K>(std::move(futures));
+        return to_object_vector<K>(std::move(future));
     }
 
     /**
@@ -803,7 +803,7 @@ public:
       query::paging_predicate<K, V>& predicate)
     {
         predicate.set_iteration_type(query::iteration_type::VALUE);
-        auto futures =
+        auto future =
           values_for_paging_predicate_data(
             protocol::codec::holder::paging_predicate_holder::of(
               predicate, serialization_service_))
@@ -816,7 +816,7 @@ public:
                   predicate.set_anchor_data_list(std::move(result.second));
                   return result.first;
               });
-        return to_object_vector<V>(std::move(futures));
+        return to_object_vector<V>(std::move(future));
     }
 
     /**
@@ -869,7 +869,7 @@ public:
       query::paging_predicate<K, V>& predicate)
     {
         predicate.set_iteration_type(query::iteration_type::ENTRY);
-        auto futures =
+        auto future =
           entry_set_for_paging_predicate_data(
             protocol::codec::holder::paging_predicate_holder::of(
               predicate, serialization_service_))
@@ -882,7 +882,7 @@ public:
                   predicate.set_anchor_data_list(std::move(result.second));
                   return result.first;
               });
-        return to_entry_object_vector<K, V>(std::move(futures));
+        return to_entry_object_vector<K, V>(std::move(future));
     }
 
     /**
--- a/hazelcast/include/hazelcast/client/imap.h
+++ b/hazelcast/include/hazelcast/client/imap.h
@@ -634,7 +634,7 @@ public:
       const K& key)
     {
         auto f = proxy::IMapImpl::get_entry_view_data(to_data(key));
-        return to_object_entry_view<K, V>(std::move(f), std::move(key));
+        return to_object_entry_view<K, V>(std::move(f));
     }
 
     /**
--- a/hazelcast/include/hazelcast/client/imap.h
+++ b/hazelcast/include/hazelcast/client/imap.h
@@ -633,8 +633,8 @@ public:
     boost::future<boost::optional<entry_view<K, V>>> get_entry_view(
       const K& key)
     {
-        return to_object_entry_view<K, V>(
-          proxy::IMapImpl::get_entry_view_data(to_data(key)), key);
+        auto f = proxy::IMapImpl::get_entry_view_data(to_data(key));
+        return to_object_entry_view<K, V>(std::move(f), std::move(key));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


yemreinci (dismissed):

yemreinci (approved):

yemreinci:
verify-linux

hazelcast/include/hazelcast/client/proxy/SerializingProxy.h
```diff
@@ -215,6 +202,25 @@ class HAZELCAST_API SerializingProxy
...
 
+    template<typename K, typename V>
+    inline boost::future<boost::optional<entry_view<K, V>>>
+    to_object_entry_view(
+      boost::future<boost::optional<map::data_entry_view>> data_future,
+      const K& key)
+    {
+        return data_future.then(
+          boost::launch::sync,
+          [=](boost::future<boost::optional<map::data_entry_view>> f) {
```
ihsandemir, line 213:
```suggestion
          [this](boost::future<boost::optional<map::data_entry_view>> f) {
```

hazelcast/include/hazelcast/client/proxy/SerializingProxy.h
```diff
@@ -215,6 +202,25 @@ class HAZELCAST_API SerializingProxy
           });
     }
 
+    template<typename K, typename V>
+    inline boost::future<boost::optional<entry_view<K, V>>>
+    to_object_entry_view(
+      boost::future<boost::optional<map::data_entry_view>> data_future,
+      const K& key)
```
ihsandemir, line 209:
```suggestion
      K key)
```
Eliminate copy and use move at caller since most probably that key will not be used after that point on the caller, right?

hazelcast/include/hazelcast/client/proxy/SerializingProxy.h
```diff
@@ -92,41 +93,46 @@ class HAZELCAST_API SerializingProxy
...
+      std::vector<T>& elements)
     {
-        return to_object<T>(data.get_ptr());
+        return f.then(
+          boost::launch::sync,
+          [&elements,
+           this](boost::future<std::vector<serialization::pimpl::data>> f) {
+              auto datas = f.get();
+              auto size = datas.size();
+              elements.reserve(elements.size() + size);
```
ihsandemir, line 130: There may be a bug here. According to javadoc, it says `Removes all available elements from this queue and adds them to the given collection`
Can you double check on this?

hazelcast/test/src/HazelcastTests7.cpp
```diff
@@ -738,6 +738,18 @@ TEST_F(ClientQueueTest, testDrain)
     ASSERT_EQ("item1", list2[0]);
     ASSERT_EQ("item2", list2[1]);
     ASSERT_EQ("item3", list2[2]);
+
+    offer(2);
+    // drain to a non-empty vector
+    result = q->drain_to(list2).get();
+    ASSERT_EQ(2U, result);
+    // existing items
+    ASSERT_EQ("item1", list2[0]);
```
ihsandemir, line 747: let's double check the expected behaviour at Java.

hazelcast/include/hazelcast/client/proxy/SerializingProxy.h
sancar: I double checked, this is the correct behaviour.

sancar: Note that we are not changing the behaviour. This is just more optimized version where we resize the elements vector given from user once. In existing impl, push_back my had to resize more than once.

hazelcast/test/src/HazelcastTests7.cpp
sancar: Also note that, this is just testing the existing behavior. I added the test just because I felt that it is missing. This test already passes on the current master

hazelcast/include/hazelcast/client/imap.h
```diff
@@ -763,23 +738,20 @@ class HAZELCAST_API imap : public proxy::IMapImpl
...
+            .then(
+              boost::launch::sync,
+              [&predicate](
+                boost::future<std::pair<std::vector<serialization::pimpl::data>,
+                                        query::anchor_data_list>> f) {
+                  auto result = f.get();
+                  predicate.set_anchor_data_list(std::move(result.second));
+                  return result.first;
+              });
+        return to_object_vector<K>(std::move(future));
```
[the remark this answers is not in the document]
sancar: Moving all then block to SerializingProxy to hide sync to_object methods without adding continuation.

hazelcast/include/hazelcast/client/proxy/SerializingProxy.h
```diff
@@ -206,11 +275,11 @@ class HAZELCAST_API SerializingProxy
     inline boost::future<boost::optional<entry_view<K, V>>>
     to_object_entry_view(
       boost::future<boost::optional<map::data_entry_view>> data_future,
-      const K& key)
+      K key)
```
yemreinci, line 278: Let's count how many times the `key` is copied in this commit and the one before:

**Before** (parameter was `const K&`):
1. The parameter is passed as an lvalue reference so no copies there. (0)
2. The parameter is captured by the lambda by-value ([=]), so one copy. (1).
3. Upon completion, the lambda calls entry_view<K, V>'s ctor, which takes the key by-value, so one copy. (1).

So in total, 2 copies.

**Now** (parameter is `K`):
1. Depending on whether the call site uses an lvalue reference or a nonconst rvalue reference, there may be 1 or 0 copies. The way you call this in get_entry_view will make it copy as I explained in the comment, but assuming that you'll correct it, we can put 0 copies. (0)
2. The parameter is captured by the lambda by-value ([key]), so one copy. (1)
3. Upon completion, the lambda calls entry_view<K, V>'s ctor, which takes the key by-value, so one copy. (1).

So in total, still 2 copies. (at best, if don't change how you call to_object_entry_view it's 3)

A few words on how to optimize the number of copies:
* First of all, AFAIA this PR didn't introduce any extra copies compared to master. The [get_entry_view at master](https://github.com/hazelcast/hazelcast-cpp-client/blob/e8abe511b16e39d89e7c771477b62fb11c575f15/hazelcast/include/hazelcast/client/imap.h#L633-L647) does the same 2 copies: (1) capturing the parameter in lambda, (2) constructing an entry_view. So I think this PR should be fine to merge without any optimization.
* The first copy in the lambda capture is not possible to eliminate in C++11, if we want to keep the lambda. C++14's lambda syntax allows capturing by-move (e.g. [key = std::move(key)]) but C++11's is limited in this sense. However, one could change the lambda to a class with `operator()` overloaded (which what lambdas really are) to achieve the C++14 behavior.
* The copy that happens constructing the `entry_view` is easy to eliminate. Do this:
```C++
[key](...) mutable {
     return entry_view(std::move(key), ...);
}
```
The `mutable` keyword is necessary because by default lambdas are const-qualified, which means any reference to captured variables are also const-qualified and calling `std::move` on const reference has no real effect.

yemreinci (approved):

PR #959, opened 2022-02-25
[API-1205] Implement minimal support for SQL

yemreinci:
Implements limited support for SQL. Limitations:
- Only VARCHAR (std::string) is supported as a column type.
- Only a single page of results is supported. (multi-page support is on the way)


sancar:
Closing in favor of https://github.com/hazelcast/hazelcast-cpp-client/pull/971

PR #967, opened 2022-04-19
Introduce hazelcast::client::decimal [API-1297]

sancar:
This class will be used in SQL and in Compact to match with
Decimal type on SQL and Compact. a.k.a BigDecimal on java.


hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -0,0 +1,89 @@
...
+struct HAZELCAST_API decimal
+{
+    boost::multiprecision::cpp_int unscaled;
+    int32_t scale;
+};
+
+namespace pimpl {
+
+/**
+ * Takes twos complement of giving array where most significant value is first.
```
srknzl, line 52: What is a giving array? Did you mean the given array? We can also say that it modifies the vector in place.

hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -0,0 +1,89 @@
...
+ * @param a the array to take twos complement of
+ */
+
+void HAZELCAST_API
```
srknzl, line 56: What does HAZELCAST_API mean? As I know it makes an API public. I think this method should be internal.

hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -0,0 +1,89 @@
...
+twos_complement(std::vector<int8_t>& a);
+
+/**
+ * Creates cpp_int from vector of int8_t respecting the sign.
```
srknzl, line 60:
```suggestion
 * Creates a cpp_int from a vector of int8_t respecting the sign.
```

hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * boost::import_bits does not respect the sign, so we have to do it manually.
+ * if v represents  a negative number, we take the two's complement of it,
```
srknzl, line 63:
```suggestion
 * if v represents a negative number, we take the two's complement of it,
```

hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -0,0 +1,89 @@
...
+ * to get positive representation of the same number. Then we add the sign
+ * at the end.
+ * @param v int8_t array to read from
+ * @return cpp_int
+ */
+boost::multiprecision::cpp_int HAZELCAST_API
+from_bytes(std::vector<int8_t> v);
```
srknzl, line 70: Again is this method public? If it is, I think users won't about the second paragraph. Same for to_bytes.

hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Creates cpp_int from vector of int8_t respecting the sign.
```
srknzl, line 72:
```suggestion
 * Converts a cpp_int to a vector of int8_t.
```

or something like this

srknzl (changes requested):
The first part of the review just including the review of `.h` file

hazelcast/include/hazelcast/client/decimal.h
sancar: HAZELCAST_API does not mean public. It is visibility modifiers so that Windows will compile. Anything defined in the headers should have this as far as I remember.

sancar:
>  Why about adding this to builtin serializers? Currently what do we have for bigdecimal in cpp? If we are going to do it, can you create an issue in Jira to track it?

I think there is already jira tasks for all missing types in all the languages. See https://hazelcast.atlassian.net/jira/software/projects/API/boards/40/backlog?selectedIssue=API-884&text=serializers

hazelcast/src/hazelcast/client/decimal.cpp
```diff
@@ -0,0 +1,81 @@
...
+bool
+operator==(const decimal& lhs, const decimal& rhs)
+{
+    return lhs.unscaled == rhs.unscaled && lhs.scale == rhs.scale;
+}
+
+namespace pimpl {
+
+void
+twos_complement(std::vector<int8_t>& a)
```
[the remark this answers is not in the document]
sancar: Implementation source is me I guess. I have wrote it myself. The concept of 2's complement on the other hand a common thing that you can find easily everywhere.

hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * boost::import_bits does not respect the sign, so we have to do it manually.
+ * if v represents a negative number, we take the two's complement of it,
+ * to get positive representation of the same number. Then we add the sign
+ * at the end.
+ * @param v int8_t array to read from
+ * @return cpp_int
+ */
+boost::multiprecision::cpp_int HAZELCAST_API
+from_bytes(std::vector<int8_t> v);
```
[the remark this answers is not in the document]
sancar: No, they will be used to serialize deserialize decimal type in SqlPageCodec, builtin serializer and in compact.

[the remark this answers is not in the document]
sancar: signed int was more convenient to check sign of the left most bit. "x < 0" does the job in this case. 
Also now it is more similar to java. Not really sure which one came first when implementing but both came to my mind.

srknzl (dismissed):

hazelcast/src/hazelcast/client/decimal.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    for (auto& item : a) {
+        item = ~item;
+    }
+    int carry = 1;
```
ihsandemir, line 36:
```suggestion
    int8_t carry = 1;
```

ihsandemir, line 36: Can you document what is the purpose of `carry`?

hazelcast/src/hazelcast/client/decimal.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+std::vector<int8_t>
+to_bytes(const boost::multiprecision::cpp_int& i)
+{
+    std::vector<int8_t> v;
+    export_bits(i, std::back_inserter(v), 8);
+    if (i < 0) {
+        twos_complement(v);
+        if (v[0] > 0) {
+            v.insert(v.begin(), 1, -1);
```
ihsandemir, line 70:
```suggestion
            v.insert(v.begin(), -1);
```

hazelcast/src/hazelcast/client/decimal.cpp
```diff
@@ -0,0 +1,81 @@
...
+        }
+    } else {
+        if (v[0] < 0) {
+            v.insert(v.begin(), 1, 0);
```
ihsandemir, line 74:
```suggestion
            v.insert(v.begin(), 0);
```

hazelcast/src/hazelcast/client/decimal.cpp
```diff
@@ -0,0 +1,81 @@
+/*
```
ihsandemir, line 1: Can we delete this cpp file and put it into `client_impl.cpp`?

hazelcast/include/hazelcast/client/decimal.h
```diff
@@ -79,7 +92,8 @@ from_bytes(std::vector<int8_t> v);
  * if i is a negative number, we take the two's complement on resulting vector,
  * to get negative representation of the number.
  * We also add one extra byte to the end of the vector to preserve the sign if
- * necessary.
+ * sign of the integer is not same as the most significant byte's sign.
```
srknzl: `is not the same`

hazelcast/test/src/compact_test.cpp
```diff
@@ -494,6 +495,9 @@ check_schema_field(const schema& schema,
 
 TEST_F(CompactSerializationTest, test_field_order_fixed_size)
 {
+    std::unordered_map<hazelcast::client::decimal, hazelcast::client::decimal> m(4);
+    m.insert(std::make_pair(hazelcast::client::decimal{1, 2},
+                            hazelcast::client::decimal{3, 4}));
```
srknzl: I think these two lines looks weird in this test, we can add this as a new test.

sancar: It is just a leftover. I removed them.

srknzl (approved):

ihsandemir (dismissed):

ihsandemir (approved):

ihsandemir:
Serkan, already approved the PR. Merging the PR.

PR #968, opened 2022-04-20
Compact Nullable Primitive Arrays [API-1297]

sancar:


ihsandemir (dismissed):

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -411,6 +411,96 @@ class HAZELCAST_API compact_reader
...
+
+    /**
+     * Reads a nullable array of nullable 16-bit two's complement signed
+     * integers.
+     *
+     * @param field_name name of the field.
+     *  @return the value of the field.
+     *  @throws hazelcast_serialization if the field does not exist in the
+     *  schema or the type of the field does not match with the one defined
+     *  in the schema.
```
srknzl, line 447:
```suggestion
     * @return the value of the field.
     * @throws hazelcast_serialization if the field does not exist in the
     * schema or the type of the field does not match with the one defined
     * in the schema.
```

Similar for other methods

hazelcast/include/hazelcast/client/serialization/pimpl/compact.i.h
```diff
@@ -193,6 +193,33 @@ compact_reader::read()
...
+      std::vector<boost::optional<bool>>(len));
+    int index = 0;
+    byte current_byte = object_data_input.read<byte>();
+    for (int i = 0; i < len; ++i) {
+        if (index == util::Bits::BITS_IN_BYTE) {
+            index = 0;
+            current_byte = object_data_input.read<byte>();
+        }
+        values.value()[i] =
+          boost::make_optional<bool>(((current_byte >> index) & 1) != 0);
```
srknzl, line 217: Because we have make_optional here, it is better for readability to introduce the `result` variable here like in java.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.i.h
```diff
@@ -289,6 +316,55 @@ compact_reader::read_nullable_primitive(
...
+    util::finally set_position_back(
+      [this, current_pos]() { this->object_data_input.position(current_pos); });
+
+    int32_t pos = read_var_size_position(field_descriptor);
+    if (pos == pimpl::offset_reader::NULL_OFFSET) {
+        return boost::none;
+    }
+    object_data_input.position(pos);
+    int32_t item_count = object_data_input.read<int32_t>();
+    auto values = boost::make_optional<std::vector<boost::optional<T>>>(
```
srknzl, line 350: Can we make this optional just before returning?

hazelcast/include/hazelcast/client/serialization/pimpl/compact.i.h
```diff
@@ -289,6 +316,55 @@ compact_reader::read_nullable_primitive(
...
+
+template<>
+boost::optional<std::vector<boost::optional<bool>>> inline compact_reader::
+  read_primitive_array_as_nullable_array(
+    const pimpl::field_descriptor& field_descriptor)
+{
+    return read_variable_size<std::vector<boost::optional<bool>>>(
+      field_descriptor);
+}
+
```
srknzl, line 367: 🤔 Do we need this? There should not be a need of read_primitive_array_as_nullable_array for nullable boolean

hazelcast/test/src/compact_test.cpp
```diff
@@ -677,14 +843,19 @@ TEST_F(CompactNullablePrimitiveInteroperabilityTest,
...
+    ASSERT_SAME_CONTENT(expected.nullableBytes, actual.bytes);
+    ASSERT_SAME_CONTENT(expected.nullableShorts, actual.shorts);
+    ASSERT_SAME_CONTENT(expected.nullableInts, actual.ints);
+    ASSERT_SAME_CONTENT(expected.nullableLongs, actual.longs);
+    ASSERT_SAME_CONTENT(expected.nullableFloats, actual.floats);
+    ASSERT_SAME_CONTENT(expected.nullableDoubles, actual.doubles);
 }
 
 TEST_F(CompactNullablePrimitiveInteroperabilityTest,
        testWriteNullReadPrimitiveThrowsException)
```
srknzl, line 856: In this test we are not testing try to reading every field throws, we are just asserting trying to read all fields throws or not, can we do anything about it 🤔 ?

hazelcast/include/hazelcast/client/serialization/pimpl/compact.i.h
sancar: Yes, I think we need it. 

getArrayOfNullableBoolean calls getVariableSize in java if field_kind is `ARRAY_OF_BOOLEAN`. 

To have the same behavior. `read_array_of_nullable` should call `read_array_of_nullable` when T is `std::vector<boost::optional<bool>`.

hazelcast/test/src/compact_test.cpp
sancar: I thought the same thing as well. But I could not found a smart way of doing this without copy pasting test for each field type.

srknzl (approved):

ihsandemir (approved):

PR #969, opened 2022-04-23
Introduce controlled serialization

sancar:
PR closed by Hazelcast automation as no activity (>3 months). Please reopen with comments, if necessary. Thank you for using Hazelcast and your valuable contributions

### DIFF
--- a/hazelcast/include/hazelcast/client/imap.h
+++ b/hazelcast/include/hazelcast/client/imap.h
@@ -101,7 +101,8 @@ public:
     template<typename K, typename V>
     boost::future<boost::optional<V>> get(const K& key)
     {
-        return to_object<V>(get_internal(to_data(key)));
+        return controlled_serialization(
+          [&]() { return to_object<V>(get_internal(to_data(key))); });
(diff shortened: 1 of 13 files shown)


hazelcast/include/hazelcast/client/imap.h
```diff
@@ -101,7 +101,8 @@ class HAZELCAST_API imap : public proxy::IMapImpl
     template<typename K, typename V>
     boost::future<boost::optional<V>> get(const K& key)
     {
-        return to_object<V>(get_internal(to_data(key)));
+        return controlled_serialization(
+          [&]() { return to_object<V>(get_internal(to_data(key))); });
```
ihsandemir, line 105: Reference capture of key. This may create a problem if user provided key is deleted? Please also use explicit capture of each variable rather than default capture, for better visibility.

sancar: I believe it should be ok since we will use key only in this thread. It will not be delegated to another thread.

PR #971, opened 2022-04-28
Sql Support  [API-1205]

sancar:
supersedes https://github.com/hazelcast/hazelcast-cpp-client/pull/959

### DIFF
--- a/hazelcast/include/hazelcast/client/sql/impl/query_id.h
+++ b/hazelcast/include/hazelcast/client/sql/impl/query_id.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2008-2021, Hazelcast, Inc. All Rights Reserved.
+ *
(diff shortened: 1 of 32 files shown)


hazelcast/include/hazelcast/client/sql/impl/query_id.h
```diff
@@ -0,0 +1,50 @@
...
+class HAZELCAST_API query_id
+{
+public:
+    query_id(int64_t member_id_high,
+             int64_t member_id_low,
+             int64_t local_id_high,
+             int64_t local_id_low);
+
+    int64_t member_id_high() const;
+    int64_t member_id_low() const;
```
ihsandemir, line 36: This is how Java uses uuid, but we can use boost::uuid which is 16 bytes instead.

ihsandemir:
continue with PR #995

Issue #976, opened 2022-06-01
The static lock used in logger may be destructed before the client is destroyed which may cause   [API-1358]


ihsandemir:
The following test program crashes (tested at Mac OS) with the below message: 

```
01/06/2022 13:50:24.603 INFO: [0x16f227000] hz.client_1[dev] [5.0.0] (Thu Oct 14 10:30:14 2021 +0300:5230f0871) LifecycleService::LifecycleEvent Client (ebacea60-2b4d-4e87-acb6-67376fe94a60) is STARTING
01/06/2022 13:50:24.603 INFO: [0x16f227000] hz.client_1[dev] [5.0.0] (Thu Oct 14 10:30:14 2021 +0300:5230f0871) LifecycleService::LifecycleEvent STARTING
01/06/2022 13:50:24.603 INFO: [0x16f227000] hz.client_1[dev] [5.0.0] LifecycleService::LifecycleEvent STARTED
01/06/2022 13:50:24.604 INFO: [0x16f227000] hz.client_1[dev] [5.0.0] Trying to connect to Address[127.0.0.1:5701].
01/06/2022 13:50:24.617 INFO: [0x16f227000] hz.client_1[dev] [5.0.0] LifecycleService::LifecycleEvent CLIENT_CONNECTED
01/06/2022 13:50:24.617 INFO: [0x16f227000] hz.client_1[dev] [5.0.0] Authenticated with server  Address[10.0.101.18:5701]:5451d9e6-16ba-4f49-b308-c7dad26f9c73, server version: 5.1-SNAPSHOT, local address: Address[127.0.0.1:57921]. Connection{alive=1, connection id=1, remote endpoint=Address[10.0.101.18:5701], last_read_time=2022-06-01 13:50:24.617, last_write_time=2022-06-01 13:50:24.605, closedTime=never, connected server version=5.1-SNAPSHOT}
01/06/2022 13:50:24.619 INFO: [0x16fd17000] hz.client_1[dev] [5.0.0] 

Members [1]  {
	Member[10.0.101.18]:5701 - 5451d9e6-16ba-4f49-b308-c7dad26f9c73
}

Finished
libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
Signal: SIGABRT (signal SIGABRT)
```

```C++
#include <hazelcast/client/hazelcast_client.h>

[...]

PR #977, opened 2022-06-01
Make sure that the logger static lock mutex is not destroyed before any global client instance

ihsandemir:
Changed the logger lock mutex so that we can control its destruction order properly. We do not want it destroyed before the client is destructed. We need to control the order. static initialization and destruction order is not controllable if it is not a member of the logger class. It can be destroyed before the any global client instance is destroyed since global variables are also in static duration (https://www.learncpp.com/cpp-tutorial/introduction-to-global-variables).

See the related issue for a reproducer of the problem. Unfortunately, I could not write a unit test for it since i needed global variable and static variable destruction which would require a separate test binary. Hence, i tests manually using the test code in the issue.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/976

### DIFF
--- a/hazelcast/include/hazelcast/logger.h
+++ b/hazelcast/include/hazelcast/logger.h
@@ -3,6 +3,7 @@
 #include <string>
 #include <functional>
 #include <limits>
+#include <mutex>
 
 #include "hazelcast/util/export.h"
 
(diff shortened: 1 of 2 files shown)


ihsandemir:
verify-linux

srknzl:
I did not quite understand what we are doing. Is the cout_lock global variable now instead of a static variable? If so, how does it help?

yuce (approved):

srknzl (approved):

PR #982, opened 2022-06-13
[API-1379] Update the test server version to update the server letsencrypt certificate

ihsandemir:
Update the test server version to 5.2-SNAPSHOT so that the builds do not fail due to expired letsencrypt certificates.

fixes #984

PR #983, opened 2022-06-13
[Backport] Update the test server version to update the server letsencrypt certificate 

ihsandemir:
Update the test server version to 5.2-SNAPSHOT so that the builds do not fail due to expired letsencrypt certificates.

backports #982


scripts/start-rc.bat
```diff
@@ -1,7 +1,9 @@
 @echo off
 setlocal EnableDelayedExpansion
 
-set HZ_VERSION=4.1.4-SNAPSHOT
+if "%HZ_VERSION%"=="" (
+    set HZ_VERSION=5.2-SNAPSHOT
```
mdumandag, line 5: Should we use 5.2 with the 4.1.z branch? I have backported the certificate PR to maintenance Hazelcast branches as well. So, `4.1.10-SNAPSHOT` would also work for this branch.

ihsandemir: It can be both ways. We should be working in all cases anyway and the during the release compatibility tests will handle it.

mdumandag (approved):

ihsandemir:
verify-linux

ihsandemir:
verify-linux

srknzl (approved):

mdumandag (approved):

PR #994, opened 2022-06-20
[BACKPORT] Make sure that the logger static lock mutex is not destroyed before any global client instance

ihsandemir:
Make sure that the logger static lock mutex is not destroyed before any global client instance (https://github.com/hazelcast/hazelcast-cpp-client/pull/977)

Changed the logger lock mutex so that we can control its destruction order properly. We do not want it destroyed before the client is destructed. We need to control the order. static initialization and destruction order is not controllable if it is not a member of the logger class. It can be destroyed before the any global client instance is destroyed since global variables are also in static duration (https://www.learncpp.com/cpp-tutorial/introduction-to-global-variables).

backports https://github.com/hazelcast/hazelcast-cpp-client/pull/977

### DIFF
--- a/hazelcast/include/hazelcast/logger.h
+++ b/hazelcast/include/hazelcast/logger.h
@@ -3,6 +3,7 @@
 #include <string>
 #include <functional>
 #include <limits>
+#include <mutex>
 
 #include "hazelcast/util/export.h"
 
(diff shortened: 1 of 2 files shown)


srknzl (approved):

ihsandemir:
verify-windows

Issue #944, opened 2021-12-09
Example member configs for authentication are incorrect [API-1080]


yemreinci:
The XML config examples for [token authentication](https://github.com/hazelcast/hazelcast-cpp-client/blob/master/Reference_Manual.md#5102-token-authentication) and [username-password authentication](https://github.com/hazelcast/hazelcast-cpp-client/blob/master/Reference_Manual.md#5101-username-password-authentication) are incorrect:

1. The example for username password authentication should include a `<member-authentication>` option to work.
2. There is no way to configure token authentication without writing Java code.

ihsandemir:
fixed by https://github.com/hazelcast/hazelcast-cpp-client/pull/1011

PR #1004, opened 2022-08-08
Authentication configs are fixed [API-1080]

hakanaktas0:
Fix for [this issue](https://github.com/hazelcast/hazelcast-cpp-client/issues/944)

### DIFF
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -1246,8 +1246,9 @@ If you want to enable authentication, then the client can be configured in one o
 The following is an example configuration where we set the username `test-user` and password `test-pass` to be used while trying to connect to the cluster:
 
 ```c++
-hazelcast_client hz(client_config().set_credentials(
-std::make_shared<security::username_password_credentials>("test-user", "test-pass")));
+hazelcast::client::client_config config;
+config.set_credentials(std::make_shared<hazelcast::client::security::username_password_credentials>("test-user", "test-pass"));
+auto hz = hazelcast::new_client(std::move(config)).get();
 ```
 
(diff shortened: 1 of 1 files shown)


Reference_Manual.md
```diff
@@ -1269,22 +1271,74 @@ The following is an example configuration where we set the secret token bytes to
...
-                    <token encoding="base64">SGF6ZW</token>
-                </identity>
-            </realm>
-        </realms>
-    </security>
+    <realms>
+        <realm name="jaasRealm">
+            <authentication>
+                <jaas>
+                    <login-module class-name="path.to.file.TokenLoginModule" usage="REQUIRED">
```
ihsandemir, line 1331: is this path or package name?

hakanaktas0: path

Reference_Manual.md
```diff
@@ -1269,22 +1271,74 @@ The following is an example configuration where we set the secret token bytes to
...
+                    </login-module>
+                </jaas>
+            </authentication>
+            <identity>
+                <token>SGF6ZW</token>
+            </identity>
+        </realm>
+    </realms>
+    <client-authentication realm="jaasRealm"/>
```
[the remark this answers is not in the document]
hakanaktas0: I am surprised too, but the above one doesn't work with `<client-authentication realm=""/>` for some reason, client just connects whether the credentials are correct or not, when configured that way. I tried them both for both cases and this is how they work.

hakanaktas0:
Replaced with [this](https://github.com/hazelcast/hazelcast-cpp-client/pull/1011)

Issue #1005, opened 2022-08-11
Occasional assertion failure in hazelcast::cp::internal::session::proxy_session_manager::create_new_session() [API-1501]


bjd:
C++ compiler version: g++ 7.3.1
Hazelcast Cpp client version: 5.0.0
OS version (Windows/Linux/OSX):  Linux

```
/hazelcast-cpp-client-5.0.0/hazelcast/src/hazelcast/cp/cp_impl.cpp:73: std::unordered_map<hazelcast::cp::raft_group_id, hazelcast::cp::internal::session::proxy_session_manager::session_state>::iterator hazelcast::cp::internal::session::proxy_session_manager::create_new_session(const hazelcast::cp::raft_group_id&): Assertion `result.second' failed.
```
```
Thread 38 (Thread 0x7f42687f0700 (LWP 300907)):
#0  0x00007f428a35f1ff in select () from /lib64/libc.so.6
#1  0x00000000004630e3 in segv_handler (sig=6, info=<optimized out>)
#2  <signal handler called>
#3  0x00007f428a286a9f in raise () from /lib64/libc.so.6
#4  0x00007f428a259e05 in abort () from /lib64/libc.so.6
#5  0x00007f428a259cd9 in __assert_fail_base.cold.0 () from /lib64/libc.so.6
#6  0x00007f428a27f3f6 in __assert_fail () from /lib64/libc.so.6
#7  0x00007f428c11994c in hazelcast::cp::internal::session::proxy_session_manager::create_new_session(hazelcast::cp::raft_group_id const&) () from /home/user/app/lib/libhazelcast-cpp-client.so.5.0.0
#8  0x00007f428c1197a1 in hazelcast::cp::internal::session::proxy_session_manager::get_or_create_session(hazelcast::cp::raft_group_id const&) () from /home/user/app/lib/libhazelcast-cpp-client.so.5.0.0
#9  0x00007f428c1194c7 in hazelcast::cp::internal::session::proxy_session_manager::acquire_session(hazelcast::cp::raft_group_id const&) () from /home/user/app/lib/libhazelcast-cpp-client.so.5.0.0
#10 0x00007f428c0834e8 in hazelcast::cp::fenced_lock::lock_and_get_fence()::{lambda()#1}::operator()() const () from /home/user/app/lib/libhazelcast-cpp-client.so.5.0.0
[...]

PR #1010, opened 2022-08-23
Assertion Failure is Fixed [API-1501]

hakanaktas0:
solves [this issue](https://github.com/hazelcast/hazelcast-cpp-client/issues/1005)

### DIFF
--- a/hazelcast/src/hazelcast/cp/cp_impl.cpp
+++ b/hazelcast/src/hazelcast/cp/cp_impl.cpp
@@ -82,6 +82,10 @@ proxy_session_manager::create_new_session(const raft_group_id& group_id)
     auto response = request_new_session(group_id);
     auto result = sessions_.emplace(
       group_id, session_state{ response.id, response.ttl_millis });
+    if(!result.second){
+        sessions_.erase(result.first);
+        result = sessions_.emplace(group_id, session_state{response.id, response.ttl_millis});
+    }
     assert(result.second);
     auto session =
       sessions_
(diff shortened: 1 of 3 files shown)


ihsandemir (commented):
Can you add a test for the fix?

ihsandemir:
verify

hakanaktas0:
verify

hakanaktas0:
Because TTL has to be an integer it can't be less than 1, 0 is infinite, I had to add 1 second sleep. Before the fix, this test would give an assertion error, should pass without problems after the fix.

ihsandemir (approved):

OzanCansel:
I have reviewev the PR and here are my findings :

- Firstly, the fix is correct so it works.

But I don't think it is a good implementation because we call `emplace` function 3 times. Indeed, there is no need for that, it can be handled with only one emplace operation..

1. Add copy assingment special member function to the session_state class.
``` C++
    // cp_impl.h#96
    session_state& operator=(const session_state& rhs);
```
2. Make definition of it as :
``` C++
// cp_impl.cpp
proxy_session_manager::session_state&
proxy_session_manager::session_state::operator=(const session_state& rhs)
{
  id  = rhs.id;
  ttl = rhs.ttl;
  creation_time = rhs.creation_time;
  acquire_count = rhs.acquire_count.load();

  return *this;
}
```

3. Define `create_new_session` like this. If there is already an entry at the `sessions_` with the specified `group_id`, then just assign brand-new `session_state` to the `value` of it.

``` C++
   // the lock_ is already acquired as write lock
    auto response = request_new_session(group_id);
    session_state state { response.id , response.ttl_millis };

    auto result = sessions_.emplace( group_id, state );
    
    if(!result.second)
        result.first->second = state;

    schedule_heartbeat_task(response.heartbeat_millis);
    return result.first;
```

PR #1011, opened 2022-08-26
Authentication fix [API-1080]

hakanaktas0:
Replaces [this](https://github.com/hazelcast/hazelcast-cpp-client/pull/1004)
All the recommended changes from the previous PR are implemented.

### DIFF
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -2246,23 +2246,24 @@ See the following example.
 
 ```c++
 int main() {
-    auto hz = hazelcast::new_client().get();
-
-    auto map = hz.get_map("EntryListenerExampleMap").get();
-    
-    auto registrationId = map->add_entry_listener(
-            hazelcast::client::entry_listener().on_added([](hazelcast::client::entry_event &&event) { 
-                std::cout << "Entry added:" << event.get_key().get<int>().value() << " --> " << event.get_value().get<std::string>().value() << std::endl;
-            }).on_removed([](hazelcast::client::entry_event &&event) {
-                std::cout << "Entry removed:" << event.get_key().get<int>().value()
-                          << " --> " << event.get_value().get<std::string>().value() << std::endl;
-            }), true).get();
-
-    map->put(1, "My new entry").get();
-    map->remove<int, std::string>(1).get();
-    /* ... */
-
-    return 0;
+  auto hz = hazelcast::new_client().get();
+  
+  auto map = hz.get_map("EntryListenerExampleMap").get();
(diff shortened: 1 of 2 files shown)


Reference_Manual.md
```diff
@@ -1260,6 +1261,7 @@ Note that the server needs to be configured to use the same username and passwor
                 </identity>
             </realm>
         </realms>
+        <member-authentication realm="usernamePasswordIdentityRealm"/>
```
mdumandag, line 1264: OK, I just looked at the code and as we discussed, we should be using the `client-authentication` here. The problem was that the use of `identity` configuration was wrong. It looks like the server expects you to use `authentication` instead. So, the correct member config should be like

```xml
    <security enabled="true">
        <realms>
            <realm name="usernamePasswordIdentityRealm">
                <authentication>
                    <simple>
                        <user username="test-user" password="test-pass"/>
                    </simple>
                </authentication>
            </realm>
        </realms>
        <client-authentication realm="usernamePasswordIdentityRealm"/>
    </security>
```

Can you try this and update the config with it?

Reference_Manual.md
```diff
@@ -1269,22 +1271,70 @@ The following is an example configuration where we set the secret token bytes to
...
+            return true;
+        }
+        throw new LoginException("Invalid token");
+    }
 
-hazelcast_client hz(client_config().set_credentials(
-std::make_shared<security::token_credentials>(my_token)));
+    @Override
+    protected String getName() {
+        return name;
```
mdumandag, line 1317: The name is always `null` right now. Maybe we can do something similar to `DefaultLoginModule` and assign it to `name = credentials.getName();` in the `onLogin` method

Reference_Manual.md
```diff
@@ -1269,22 +1273,70 @@ The following is an example configuration where we set the secret token bytes to
...
+
+    private String name;
+    private final byte[] expectedToken = "SGF6ZW".getBytes(StandardCharsets.US_ASCII);
 
-hazelcast_client hz(client_config().set_credentials(
-std::make_shared<security::token_credentials>(my_token)));
+    @Override
+    protected boolean onLogin() throws LoginException {
+        CredentialsCallback cb = new CredentialsCallback();
+	name = cb.getCredentials().getName();
```
mdumandag, line 1301: minor: Let's move this line below `Credentials credentials = cb.getCredentials();`, you already call `getCredentials` there, no need to duplicate it.

mdumandag (approved):
approving with a minor comment

Issue #1022, opened 2022-10-06
Release version nightly builds started to fail [API-1625]


ihsandemir:
See release builds at [jenkins](https://jenkins.hazelcast.com/view/C++/) The failures are linkage errors:

Please attach relevant logs and files for client and server side.

```
19:03:27 [100%] [32m[1mLinking CXX executable client_test[0m
19:03:27 cd /appdisk/jenkins/jenkins_slave/workspace/cpp-linux-32-SHARED-Release/build/hazelcast/test/src && /usr/bin/cmake -E cmake_link_script CMakeFiles/client_test.dir/link.txt --verbose=1
[...]

PR #1024, opened 2022-10-24
Unresolved external errors are fixed. [API-1625] (#1022)

OzanCansel:
> But when you define an inline member function (the {...} part), you prepend the member function’s definition with the keyword inline, and you (almost always) put the definition into a header file:

...

> The reason you (almost always) put the definition (the {...} part) of an inline function in a header file is to avoid “unresolved external” errors from the linker. That error will occur if you put the inline function’s definition in a .cpp file and if that function is called from some other .cpp file.

Ref : [isocpp.org](https://isocpp.org/wiki/faq/inline-functions#inline-member-fns)

### DIFF
--- a/hazelcast/src/hazelcast/client/serialization.cpp
+++ b/hazelcast/src/hazelcast/client/serialization.cpp
@@ -626,7 +626,7 @@ data_output::write_zero_bytes(size_t number_of_bytes)
     output_stream_.insert(output_stream_.end(), number_of_bytes, 0);
 }
 
-inline size_t
+size_t
 data_output::position()
 {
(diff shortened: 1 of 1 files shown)


ihsandemir (changes requested):
This seemed to change in f6c8d7a3d70ef3cd2cce495a1d19cf520d2bdf2a commit. Please check that commit and let's also put those functions there as inline functions since they were inline before.

ihsandemir (commented):
Can you do the definitions out of class and add documentation to these methods?

ihsandemir (approved):

Issue #1012, opened 2022-08-26
Nightly Builds Fail [API-1527]


hakanaktas0:
Release builds of all nightly builds fail as seen [here](https://github.com/hazelcast/hazelcast-cpp-client/actions).

hakanaktas0:
Only release builds fail for Ubuntu and MacOS. 
All builds fail for Windows. For Windows problem seems to be related to boost install.
Logs:
[Ubuntu Logs.zip](https://github.com/hazelcast/hazelcast-cpp-client/files/9433965/Ubuntu.Logs.zip)
[MacOS Logs.zip](https://github.com/hazelcast/hazelcast-cpp-client/files/9433966/MacOS.Logs.zip)
[Windows Logs.zip](https://github.com/hazelcast/hazelcast-cpp-client/files/9433967/Windows.Logs.zip)
"

hakanaktas0:
Changing windows version to 2019 fixes boost installation error.

OzanCansel:
- Ubuntu builds fail because of undefined reference error and fix #1022 also solves this issue.
- MacOS release builds fail because we treat a warning as an error and new version of boost which generates a warning message which tells that language standard of Boost.Math module will be increased to C++14 in July 2023.
- Windows builds fail because windows-latest update to windows-2022 and it comes with VS17-2022 toolchain which boost is compilable only after 1.78 version but the action downloads 1.76.0.

PR #1025, opened 2022-10-26
Nightly builds are working again. [API-1527] #1012

OzanCansel:


.github/workflows/nightly-macos-x86_64.yml
```diff
@@ -10,6 +10,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        boost:
+            - version: 1.71.0
+            - version: 1.76.0
```
ihsandemir, line 15: what about with latest boost? what was the reason for 1.76? Was it related to C++14?

.github/workflows/nightly-windows.yml
```diff
@@ -50,16 +81,16 @@ jobs:
...
               -UserAgent "''"
-          tar xzf boost_1_76_0.tar.gz
-          rm boost_1_76_0.tar.gz
-          cd boost_1_76_0
+          tar xzf ${{ matrix.vc_boost.boost_archive_name }}
+          rm ${{ matrix.vc_boost.boost_archive_name }}
+          cd ${{ matrix.vc_boost.boost_folder_name }}
           .\bootstrap.bat
-          .\b2 address-model=${{ matrix.arch.address_model }} --with-thread --with-chrono install
+          .\b2 address-model=${{ matrix.arch.address_model }} --with-atomic --with-thread --with-chrono install
```
ihsandemir, line 91: what was the reason for atomic?

ihsandemir (commented):
Can you also share your fork repo github actions run link that shows that the builds are green with these changes?

.github/workflows/nightly-macos-x86_64.yml
OzanCansel: Yes, it is related with C++14.

.github/workflows/nightly-windows.yml
OzanCansel: We have dependency to `Boost.Thread` which transitively depends on `Boost.Atomic`. At olders versions, Boost doesn't complain about it but with newer versions it complains so that's why I added this flag.

.github/workflows/nightly-macos-x86_64.yml
OzanCansel: It doesn't affect the users of the library, the reason why this build is we treat the warnings as an error at Release build. We can do code change to fix that issue but i don't think it is required because there is nothing wrong with the code, it is just a preliminary warning for developers which tells `Boost.Math` will increase minimum C++ stantard to C++14.

OzanCansel:
> Can you also share your fork repo github actions run link that shows that the builds are green with these changes?

[macOs](https://github.com/OzanCansel/hazelcast-cpp-client/actions/runs/3366179212)
[Ubuntu-i386](https://github.com/OzanCansel/hazelcast-cpp-client/actions/runs/3365145880)
[Ubuntu-x86_64](https://github.com/OzanCansel/hazelcast-cpp-client/actions/runs/3364514662)
[Windows](https://github.com/OzanCansel/hazelcast-cpp-client/actions/runs/3362276418)

ihsandemir (approved):

PR #1027, opened 2022-10-28
[API-1501] Fix assert fails. (#1005)

### LINKED ISSUE #1005: Occasional assertion failure in hazelcast::cp::internal::session::proxy_session_manager::create_new_session() [API-1501]

OzanCansel:
- Removed assert statement.
- If there is an entry at the sessions then emplace doesn't happen so I changed it to something like 'upsert' operation. If the entry is already there I just update the 'session_state' to the brand-new one.

Fixes #1005

### DIFF
--- a/hazelcast/src/hazelcast/cp/cp_impl.cpp
+++ b/hazelcast/src/hazelcast/cp/cp_impl.cpp
@@ -80,18 +80,15 @@ proxy_session_manager::create_new_session(const raft_group_id& group_id)
 {
     // the lock_ is already acquired as write lock
     auto response = request_new_session(group_id);
-    auto result = sessions_.emplace(
-      group_id, session_state{ response.id, response.ttl_millis });
-    if(!result.second){
-        sessions_.erase(result.first);
-        result = sessions_.emplace(group_id, session_state{response.id, response.ttl_millis});
-    }
-    auto session =
-      sessions_
-        .emplace(group_id, session_state{ response.id, response.ttl_millis })
-        .first;
+    session_state state { response.id , response.ttl_millis };
+
+    auto result = sessions_.emplace( group_id, state );
+
+    if(!result.second)
+        result.first->second = state;
+
     schedule_heartbeat_task(response.heartbeat_millis);
-    return session;
+    return result.first;
 }
 
 proxy_session_manager::session_response
@@ -329,6 +326,17 @@ proxy_session_manager::session_state::session_state(const session_state& rhs)
   , creation_time(rhs.creation_time)
   , acquire_count(rhs.acquire_count.load())
 {}
+
+proxy_session_manager::session_state&
+proxy_session_manager::session_state::operator=(const session_state& rhs)
+{
+  id  = rhs.id;
+  ttl = rhs.ttl;
+  creation_time = rhs.creation_time;
+  acquire_count = rhs.acquire_count.load();
+
+  return *this;
+}
 
 int64_t
 proxy_session_manager::session_state::acquire(int32_t count)
--- a/hazelcast/include/hazelcast/cp/cp_impl.h
+++ b/hazelcast/include/hazelcast/cp/cp_impl.h
@@ -93,6 +93,7 @@ private:
     {
         session_state(int64_t id, int64_t ttl_millis);
         session_state(const session_state& rhs);
+        session_state& operator=(const session_state& rhs);
 
         int64_t id;
         std::chrono::milliseconds ttl;


hazelcast/src/hazelcast/cp/cp_impl.cpp
```diff
@@ -80,18 +80,15 @@ proxy_session_manager::create_new_session(const raft_group_id& group_id)
...
-    auto session =
-      sessions_
-        .emplace(group_id, session_state{ response.id, response.ttl_millis })
-        .first;
+    session_state state { response.id , response.ttl_millis };
+
+    auto result = sessions_.emplace( group_id, state );
+
+    if(!result.second)
+        result.first->second = state;
```
ihsandemir, line 88:
```suggestion
    if(!result.second) {
        result.first->second = session_state state { response.id , response.ttl_millis };
   }
```
Can you use move operator and add move assignment operator instead?

ihsandemir (changes requested):
Looks good.

Is there a unit test? If missing, can we add a test?

OzanCansel:
> Looks good.
> 
> Is there a unit test? If missing, can we add a test?

Yes it is already written at [here](https://github.com/hazelcast/hazelcast-cpp-client/blob/master/hazelcast/test/src/HazelcastTests8.cpp#L1827)

ihsandemir (approved):

PR #1032, opened 2022-11-09
C++17 support for Hazelcast client

cngzhnp:
Currently, C++ version was set to `C++11`. However, repository could be compiled with `C++17` as well even though some features are still taken from `Boost` libraries.

In this PR, I just focus on without breaking backward compability add `C++17` support and replace `Boost::Optional` with `std::optional` which was introduced in `C++17` standards.

Please review my changes and do not hesitate to contact me about your questions. 

Thanks in advance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,9 +23,9 @@ project(hazelcast-cpp-client
 
 # Set the C++ standard
 set(HZ_HAS_CXX17_SUPPORT OFF)
-if( "${HZ_HAS_CXX17_SUPPORT}" STREQUAL "ON" )
+if("${HZ_HAS_CXX17_SUPPORT}" STREQUAL "ON")
     set(CMAKE_CXX_STANDARD 17)
-elif( "${HZ_HAS_CXX17_SUPPORT}" STREQUAL "OFF" )
+elseif("${HZ_HAS_CXX17_SUPPORT}" STREQUAL "OFF")
     set(CMAKE_CXX_STANDARD 11)
 endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,12 @@ project(hazelcast-cpp-client
         LANGUAGES CXX)
 
 # Set the C++ standard
-set(CMAKE_CXX_STANDARD 11)
+set(HZ_HAS_CXX17_SUPPORT OFF)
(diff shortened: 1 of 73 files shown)


OzanCansel:
Thanks for your contribution. I have reviewed the PR, it compiles successfully for `boost::optional`, probably runs as well. One required change is that the variable you added to `CMakelists.txt` should be an `option` and needs to be printed with other options. Also `std::nullopt` and `boost::none` should be handled as well.

But the question trickles my mind is that is this `option` really desired by user perspective?

- Maybe a user could have already used `std::optional` so probably is not willing to mix `boost::optional` and `std::optional` and stick to one of it.
- Probably `optional` header brings less overhead that `boost/optional.h`

But I don't know what is your opinion about these ?

cngzhnp:
Thanks for reviewing @OzanCansel. The main idea behind on these changes, removing Boost dependency step by step without breaking any functionality of Hazelcast library itself.

Currently, Hazelcast has huge dependency to the Boost library which cause a lot of problems when have to link Boost library statically. 

First problem, if I want to update Boost library, I have to update my application as well which cause deployment and maintenance cost for me on each Boost release. Secondly, due to static linking, executable size is getting bigger and bigger which is not quite good if I want to run the executable on small IoT devices. 

So, actually directly `std::optional` could be replaced with `boost::optional` as a first step to get rid of this dependency which brings `C++17` on the table. Maybe, some users do not want to build Hazelcast library with the latest versions of C++ compilers which does not support `C++17` clearly. IMHO, it could be kept as an option in CMakeLists.txt file.

Of course, if you have a plan to make standard C++ version of the library `C++17` in the future release, we do not need to keep `boost::optional`. `std::optional` could be used as a substitute.

OzanCansel:
I also complain about `boost` dependency because it brings ~350k lines of code for every translation unit that uses `hazelcast_client` so it makes compile time longer. Also we force our user-base to use `boost` library. But to be honest it is the framework which the `cpp-client` is built on top of it and i don't think that we are able to get rid of it.

ihsandemir:
> 

@cngzhnp I totally understand your point. We had zero dependency in very earlier releases but this is also a problem when we like to provide nice features such as continuations for futures, utilize lang. features non-existent in C++11 but in higher standard levels.   

If we can get rid of all the dependencies using C++17 or C++20, it would be nice to have this optional configuration in cmake but can we really get rid of all dependent features?

ihsandemir:
verify

OzanCansel:
@cngzhnp This change might break some code of our clients also introduces some maintenance cost. So we will reconsider this in `6.0.0`. So it can be closed.

PR #1033, opened 2022-11-11
rabin_finger_print unit tests are added. Some schema_tests added. [API-1297]

OzanCansel:
Extra rabin_finger_print units tests and schema test for all types are added.

### DIFF
--- a/hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
+++ b/hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
@@ -1407,6 +1407,13 @@ private:
 
 struct HAZELCAST_API field_descriptor
 {
+    field_descriptor(
+        enum field_kind f = static_cast<enum field_kind>( -1 ),
+        int32_t i = -1,
+        int32_t o = -1,
+        int8_t b = -1
+    );
+
     enum field_kind field_kind;
     /**
      * Index of the offset of the non-primitive field. For others, it is -1
@@ -1521,6 +1528,29 @@ struct HAZELCAST_API field_operations
 {
     static field_kind_based_operations get(enum field_kind field_kind);
 };
+
+namespace rabin_finger_print
+{
+
+/**
+ * We use uint64_t for computation to match the behaviour of >>> operator
+ * on java. We use >> instead.
+ */
+constexpr uint64_t INIT = 0xc15d213aa4d7a795L;
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, byte b);
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, int v);
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, const std::string& value);
+
+int64_t HAZELCAST_API
+fingerprint64(const std::string& type_name,
+              std::map<std::string, field_descriptor>& fields);
+}
 
 } // namespace pimpl
 } // namespace serialization
--- a/hazelcast/src/hazelcast/client/compact.cpp
+++ b/hazelcast/src/hazelcast/client/compact.cpp
@@ -1475,12 +1475,14 @@ default_compact_writer::set_position_as_null(const std::string& field_name,
     field_offsets[index] = -1;
 }
 
+field_descriptor::field_descriptor(enum field_kind f, int32_t i, int32_t o, int8_t b)
+    :   field_kind { f }
+    ,   index { i }
+    ,   offset { o }
+    ,   bit_offset { b }
+{}
+
 namespace rabin_finger_print {
-/**
- * We use uint64_t for computation to match the behaviour of >>> operator
- * on java. We use >> instead.
- */
-constexpr uint64_t INIT = 0xc15d213aa4d7a795L;
 
 static std::array<uint64_t, 256>
 init_fp_table()
@@ -1495,6 +1497,7 @@ init_fp_table()
     }
     return FP_TABLE;
 }
+
 uint64_t
 FP_TABLE_AT(int index)
 {
@@ -1682,7 +1685,7 @@ void
 schema_writer::add_field(std::string field_name, enum field_kind kind)
 {
     field_definition_map[std::move(field_name)] =
-      field_descriptor{ kind, -1, -1, -1 };
+      field_descriptor{ kind };
 }
 
 schema
--- a/hazelcast/test/src/compact_test.cpp
+++ b/hazelcast/test/src/compact_test.cpp
@@ -17,6 +17,8 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <tuple>
+#include <limits>
 
 #include <gtest/gtest.h>
 
@@ -46,6 +48,7 @@ struct bits_dto
     bool f = false;
     bool g = false;
     bool h = false;
+    bool i = false;
     int id = 0;
     boost::optional<std::vector<bool>> booleans;
 };
@@ -55,8 +58,8 @@ operator==(const bits_dto& lhs, const bits_dto& rhs)
 {
     return lhs.a == rhs.a && lhs.b == rhs.b && lhs.c == rhs.c &&
            lhs.d == rhs.d && lhs.e == rhs.e && lhs.f == rhs.f &&
-           lhs.g == rhs.g && lhs.h == rhs.h && lhs.id == rhs.id &&
-           lhs.booleans == rhs.booleans;
+           lhs.g == rhs.g && lhs.h == rhs.h && lhs.i == rhs.i &&
+           lhs.id == rhs.id && lhs.booleans == rhs.booleans;
 }
 
 struct node_dto
@@ -173,6 +176,16 @@ operator==(const main_dto& lhs, const main_dto& rhs)
            lhs.nullableF == rhs.nullableF && lhs.nullableD == rhs.nullableD;
 }
 
+struct wrong_field_name_read_obj
+{
+    int value;
+};
+
+struct type_mistmatch_obj
+{
+    int value;
+};
+
 hazelcast::client::local_time
 current_time()
 {
@@ -199,6 +212,7 @@ current_timestamp()
 {
     return hazelcast::client::local_date_time{ current_date(), current_time() };
 }
+
 hazelcast::client::offset_date_time
 current_timestamp_with_timezone()
 {
@@ -361,6 +375,7 @@ struct hz_serializer<compact::test::bits_dto> : public compact_serializer
         writer.write_boolean("f", dto.f);
         writer.write_boolean("g", dto.g);
         writer.write_boolean("h", dto.h);
+        writer.write_boolean("i", dto.i);
         writer.write_int32("id", dto.id);
         writer.write_array_of_boolean("booleans", dto.booleans);
     }
@@ -376,6 +391,7 @@ struct hz_serializer<compact::test::bits_dto> : public compact_serializer
         dto.f = reader.read_boolean("f");
         dto.g = reader.read_boolean("g");
         dto.h = reader.read_boolean("h");
+        dto.i = reader.read_boolean("i");
         dto.id = reader.read_int32("id");
         dto.booleans = reader.read_array_of_boolean("booleans");
         return dto;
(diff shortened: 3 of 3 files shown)


hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1407,6 +1407,13 @@ class HAZELCAST_API default_compact_writer
 
 struct HAZELCAST_API field_descriptor
 {
+    field_descriptor(
+        enum field_kind f = static_cast<enum field_kind>( -1 ),
```
akeles85, line 1411: As I see, the enum values differ in C++ and Java APIs. In C++, it is one less than Java. For example, BOOLEAN is 1 in Java and 0 in C++. That makes NOT_AVAILABLE is missing in C++. I think that is why we used -1 as a default value. As I see from the other part of the code, there is a possibility to access arrays with this field_kind (field_operations::get(enum field_kind field_kind)). So maybe we it might be better to add an enumaration value with a positive value.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1522,6 +1529,29 @@ struct HAZELCAST_API field_operations
...
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, byte b);
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, int v);
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, const std::string& value);
+
+int64_t HAZELCAST_API
```
akeles85, line 1550: Is there a specific reason for this function return type to be int64_t instead of uint64_t?

hazelcast/test/src/compact_test.cpp
```diff
@@ -634,6 +650,351 @@ struct hz_serializer<compact::test::employee_dto> : public compact_serializer
...
+    schema_writer  s_writer { "SomeType" };
+    auto writer = hazelcast::client::serialization::pimpl::create_compact_writer(&s_writer);
+
+    writer.write_int32("id", 0);
+    writer.write_string("name" , std::string {} );
+    writer.write_int8("age" , 0);
+    writer.write_array_of_timestamp("times", boost::none);
+
+    auto schema_id = std::move(s_writer).build().schema_id();
+    ASSERT_EQ(-5445839760245891300, schema_id);
```
akeles85, line 750: what might be the reason to have different assert values from Java API. As I see, the writer is composed from the same parameters but in Java Client side the output is compared with 3662264393229655598L

hazelcast/test/src/compact_test.cpp
```diff
@@ -734,6 +1096,37 @@ TEST_F(CompactSerializationTest, test_rabin_fingerprint_consistent_with_server)
...
+    wrong_field_name_read_obj obj;
+
+    auto data = ss.to_data(obj);
+    ASSERT_THROW(
+        ss.to_object<wrong_field_name_read_obj>(data),
+        exception::hazelcast_serialization
+    );
+}
+
+TEST_F(CompactSerializationTest, test_read_with_type_mismatch)
```
akeles85, line 1115: Should we add test case for two types which has same size. For example, while writing int32 and reading float value, will it give an exception too?

hazelcast/test/src/compact_test.cpp
```diff
@@ -634,6 +650,351 @@ struct hz_serializer<compact::test::employee_dto> : public compact_serializer
...
+        }
+    }
+};
+
+TEST_F(CompactRabinFingerprintTest, test_i8_fingerprint)
+{
+    check_each(
+        std::vector<entry_t<byte>>
+        { // Before               Val   After(Expected)
+            {100                 , -5   , -6165936963810616235},
```
akeles85, line 683: may it be better to write the expected values as uint64_t?

akeles85 (commented):
I have review the changes and write the detail information in review.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: Enum values are incorrect for this PR, I didn't correct those values because I was unaware of it at the time of writing this commit. Anyway, I think it is not in the scope of this PR, just added some unit tests. At replicate schema on write PR, [here](https://github.com/hazelcast/hazelcast-cpp-client/pull/1039/files#diff-90e0938a1ce08badb5a982647a1eee31484c2cec058fb26612cfc1afdefa206bR27) I corrected those enum values.

NOT_AVAILABLE has use when GenericRecord is introduced. Otherwise it is not used in Java which is not implemented yet in C++ too. So I don't think it needs to be added at this point. I plan to add it with GenericRecord implementation.

field_kind is initialized as -1 because any positive value is either vaiid, false-positive or out-of-bound. That's why I chose -1 to express that it is not initialized.

For https://github.com/hazelcast/hazelcast-cpp-client/pull/1039 i can make it 0.

hazelcast/test/src/compact_test.cpp
OzanCansel: I change `uint64_t` to `int64_t`.

Fixed.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
akeles85: As it will be changed in the other PR, I think we can go on with current implementation.

hazelcast/test/src/compact_test.cpp
OzanCansel: Fixed.

hazelcast/test/src/compact_test.cpp
OzanCansel: Enum values are not correct for this PR. It will be fixed in the [PR](https://github.com/hazelcast/hazelcast-cpp-client/pull/1039) when this PR is merged to master.

akeles85 (dismissed):

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1407,6 +1407,13 @@ class HAZELCAST_API default_compact_writer
 
 struct HAZELCAST_API field_descriptor
 {
+    field_descriptor(
```
ihsandemir, line 1410: implementation in header? why?

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1522,6 +1529,29 @@ struct HAZELCAST_API field_operations
     static field_kind_based_operations get(enum field_kind field_kind);
 };
 
+namespace rabin_finger_print
```
ihsandemir, line 1532: Please add class description from Javadoc.

ihsandemir, line 1532: Why don't you do it a class as in Java? We usually used classes instead of namespace at this level.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1522,6 +1529,29 @@ struct HAZELCAST_API field_operations
...
+{
+
+/**
+ * We use uint64_t for computation to match the behaviour of >>> operator
+ * on java. We use >> instead.
+ */
+constexpr uint64_t INIT = 0xc15d213aa4d7a795L;
+
```
ihsandemir, line 1540: Please put the documentation of each method similar to Java side doc.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1522,6 +1529,29 @@ struct HAZELCAST_API field_operations
...
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, byte b);
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, int v);
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, const std::string& value);
```
ihsandemir, line 1548: private at Java.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1522,6 +1529,29 @@ struct HAZELCAST_API field_operations
...
+
+int64_t HAZELCAST_API
+fingerprint64(const std::string& type_name,
```
ihsandemir, line 1551: Why don't we have `Schema` class and use it instead as in Java?

hazelcast/src/hazelcast/client/compact.cpp
```diff
@@ -1495,6 +1497,7 @@ init_fp_table()
     }
     return FP_TABLE;
 }
+
```
ihsandemir, line 1500: is this thread safe?
`static auto FP_TABLE = init_fp_table();`

hazelcast/src/hazelcast/client/compact.cpp
```diff
@@ -1495,6 +1497,7 @@ init_fp_table()
...
 uint64_t
 FP_TABLE_AT(int index)
```
ihsandemir, line 1502: I think that you may not need this extra method, if the fp_table is renamed as `fp_table` and it returns reference to the the static variable which should be initialized in a thread safe way.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
```diff
@@ -1522,6 +1529,29 @@ struct HAZELCAST_API field_operations
...
+ * We use uint64_t for computation to match the behaviour of >>> operator
+ * on java. We use >> instead.
+ */
+constexpr uint64_t INIT = 0xc15d213aa4d7a795L;
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, byte b);
+
+uint64_t HAZELCAST_API
+fingerprint64(uint64_t fp, int v);
```
ihsandemir, line 1545:
```suggestion
fingerprint64(uint64_t fp, int32_t v);
```
Java `int` is `int32_t`.

hazelcast/src/hazelcast/client/compact.cpp
ihsandemir, line 1500: I think that this is also incorrect `return static_cast<int64_t>(fingerPrint);` isn't it? see [this](https://stackoverflow.com/questions/7601731/how-does-one-safely-static-cast-between-unsigned-int-and-int)

hazelcast/test/src/compact_test.cpp
```diff
@@ -46,6 +48,7 @@ struct bits_dto
     bool f = false;
     bool g = false;
     bool h = false;
+    bool i = false;
```
ihsandemir, line 51: can you document why this was needed?

ihsandemir (changes requested):

hazelcast/test/src/compact_test.cpp
```diff
@@ -634,6 +650,351 @@ struct hz_serializer<compact::test::employee_dto> : public compact_serializer
...
+class CompactSchemaTest : public ::testing::Test
+{};
+
+TEST_F(CompactSchemaTest, test_constructor)
+{
+    using hazelcast::client::serialization::pimpl::schema;
+    using hazelcast::client::serialization::pimpl::field_descriptor;
+    using hazelcast::client::serialization::pimpl::field_kind;
+
+    schema sch
```
ihsandemir, line 762:
```suggestion
    schema all_types_schema
```
Please use more descriptor names.

hazelcast/test/src/compact_test.cpp
```diff
@@ -634,6 +650,351 @@ struct hz_serializer<compact::test::employee_dto> : public compact_serializer
...
+    {
+        std::string { "something" } ,
+        std::unordered_map<std::string , field_descriptor>
+        {
+            { std::string { "boolean-0"                 } , field_descriptor { BOOLEAN                          } } ,
```
ihsandemir, line 767: you can simplify the lines with less code similar to  
```suggestion
            { "boolean-0"                 , BOOLEAN                           } ,
```

hazelcast/src/hazelcast/client/compact.cpp
OzanCansel: `init_fp_table()` itself is not thread-safe but `static auto FP_TABLE = init_fp_table()` is thread-safe. But I think `static` keyword usage in `init_fp_table()` is unnecessary.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: Parameters are just defaulted. It is not implementation.

hazelcast/test/src/compact_test.cpp
OzanCansel: Fixed.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: Fixed.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: I dıdn't implement it, it was there I just added declarations to the header to expose outside of the translation unit to use in tests.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: rabin_finger_print should be run on map entries which are sorted but entries are stored in `std::unordered_map` in class `schema`  so entries needs to be sorted first, that's why it doesn't accept `schema` as parameter.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: I exposed it for testing purpose.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: Fixed.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: Fixed.

hazelcast/src/hazelcast/client/compact.cpp
OzanCansel: Fixed.

hazelcast/src/hazelcast/client/compact.cpp
OzanCansel: Fixed.

hazelcast/test/src/compact_test.cpp
OzanCansel: Do you mean code or PR ?

hazelcast/test/src/compact_test.cpp
OzanCansel: Fixed.

hazelcast/test/src/compact_test.cpp
```diff
@@ -634,6 +650,351 @@ struct hz_serializer<compact::test::employee_dto> : public compact_serializer
...
+    schema_writer  s_writer { "SomeType" };
+    auto writer = hazelcast::client::serialization::pimpl::create_compact_writer(&s_writer);
+
+    writer.write_int32("id", 0);
+    writer.write_string("name" , std::string {} );
+    writer.write_int8("age" , 0);
+    writer.write_array_of_timestamp("times", boost::none);
+
+    auto schema_id = std::move(s_writer).build().schema_id();
+    ASSERT_EQ(-5445839760245891300, schema_id);
```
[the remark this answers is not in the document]
OzanCansel: Corresponding test name in Java side added as a comment.

Fixed.

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
ihsandemir: Can we refactor it to a class? Or open an issue to refactor if it will take much time and it is not directly related to your changes?

hazelcast/test/src/compact_test.cpp
ihsandemir: i mean the new variable `bool i`

hazelcast/include/hazelcast/client/serialization/pimpl/compact.h
OzanCansel: Not a big deal. Refactored.

Fixed

hazelcast/test/src/compact_test.cpp
OzanCansel: Fixed.

ihsandemir (approved):

PR #1043, opened 2022-12-09
Deprecated std::random_shuffle usage replaced.

### LINKED ISSUE #1042: Doesn't compile on MSVC with C++17 because of deprecated STL method usage.
C++ compiler version: MSVC v19.32
Hazelcast Cpp client version: v5.0.0
Hazelcast server version: -
Number of the clients: -
Cluster size, i.e. the number of Hazelcast cluster members: -
OS version (Windows/Linux/OSX):
Windows

#### Expected behaviour
Compile successfully.

#### Actual behaviour
Doesn't compile due to [deprecated](https://en.cppreference.com/w/cpp/algorithm/random_shuffle) `std::random_shuffle` method usage at [here](https://github.com/hazelcast/hazelcast-cpp-client/blob/v5.0.0/hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h#L164).

Compile it on MSVC compiler with C++17 flag. Here is a [link](https://godbolt.org/z/P49xb9aM4) to demonstrate.

OzanCansel:
- `std::random_shuffle` replaced with `std::shuffle`

Fixes #1042

### DIFF
--- a/hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
+++ b/hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
@@ -188,7 +188,11 @@ private:
         if (member_addresses.empty()) {
             return;
         }
-        std::random_shuffle(member_addresses.begin(), member_addresses.end());
+
+        static thread_local std::random_device rd;
+        static thread_local std::mt19937 gen(rd());
+
+        std::shuffle(member_addresses.begin(), member_addresses.end(), gen);
     }
 
     void check_client_active();


hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
```diff
@@ -188,7 +188,11 @@ class HAZELCAST_API ClientConnectionManagerImpl
         if (member_addresses.empty()) {
             return;
         }
-        std::random_shuffle(member_addresses.begin(), member_addresses.end());
+
+        std::random_device rd;
+        std::mt19937 gen(rd());
```
ihsandemir, line 193: check [this](https://stackoverflow.com/questions/72577246/why-is-random-device-creation-expensive) link on the cost of these generator constructions. We can cache it to decrease the cost.

ihsandemir (changes requested):

hazelcast/include/hazelcast/client/connection/ClientConnectionManagerImpl.h
OzanCansel: Fixed.

Btw, mean of creating `std::random_device` and `std::mt19937` is about `2us` on my computer but doesn't matter I cached it.

PR #1044, opened 2022-12-09
Discovery token is filtered out in exception message [API-1738].

OzanCansel:
- Discovery token is filtered out at the source of exception.
- `token_should_not_be_leaked` test is added for this case.

### DIFF
--- a/hazelcast/include/hazelcast/util/SyncHttpsClient.h
+++ b/hazelcast/include/hazelcast/util/SyncHttpsClient.h
@@ -37,7 +37,8 @@ class HAZELCAST_API SyncHttpsClient
 public:
     SyncHttpsClient(const std::string& server_ip,
                     const std::string& uri_path,
-                    std::chrono::steady_clock::duration timeout);
+                    std::chrono::steady_clock::duration timeout,
+                    const std::string& secret_removal = std::string{});
 
     std::istream& connect_and_get_response();
 
@@ -45,6 +46,7 @@ private:
     std::string server_;
     std::string uri_path_;
     std::chrono::steady_clock::duration timeout_;
+    std::string secret_removal_;
 
     boost::asio::io_service io_service_;
     boost::asio::ip::tcp::resolver resolver_;
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -3072,7 +3072,7 @@ std::unordered_map<address, address>
 cloud_discovery::get_addresses()
 {
 #ifdef HZ_BUILD_WITH_SSL
-    std::string discovery_token {cloud_config_.discovery_token};
+    std::string discovery_token{ cloud_config_.discovery_token };
     try {
 
         util::SyncHttpsClient httpsConnection(cloud_base_url_,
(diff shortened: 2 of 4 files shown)


hazelcast/test/src/HazelcastTests7.cpp
```diff
@@ -1845,6 +1845,27 @@ TEST_F(cloud_discovery_test, invalid_token)
...
+    cloudConfig.discovery_token = "bwMqx9xp4RBkBWfdpE3WPsuCLd0CXNS";
+
+    auto discovery_token = cloudConfig.discovery_token;
+
+    try {
+        hazelcast::new_client(std::move(config)).get();
+        FAIL();
+    } catch (const exception::illegal_state& e) {
+        std::string message = e.what();
+        ASSERT_EQ(message.find(discovery_token), std::string::npos);
```
ihsandemir, line 1865: you should also check the nested exception message!!!

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -3071,17 +3072,21 @@ std::unordered_map<address, address>
...
                                               std::string(CLOUD_URL_PATH) +
-                                                cloud_config_.discovery_token,
+                                                discovery_token,
                                               timeout_);
         auto& conn_stream = httpsConnection.connect_and_get_response();
         return parse_json_response(conn_stream);
     } catch (std::exception& e) {
+        std::string message{ e.what() };
+        boost::replace_all(message, discovery_token, "<DISCOVERY_TOKEN>");
         std::throw_with_nested(
```
ihsandemir, line 3087: can it leak due to nested exception?

ihsandemir (changes requested):

hazelcast/src/hazelcast/client/spi.cpp
OzanCansel: On application crash it is not printed to `stdout`. But the token is included `nested_exception` that's why I added `filter` handler to `SyncHttpsRequest` earlier.

hazelcast/test/src/HazelcastTests7.cpp
OzanCansel: Nested exception contains it.

OzanCansel: Changed the implementation, `token` is filtered out in the class which is the source of the exception. So it is not leaked by neither inner nor outer exceptions.

There is a problem related with an unit test, I wrote the test to check inner exception message but `std::rethrow_if_nested` doesn't throw the inner exception. Added it to comment.

ihsandemir: Let's öpen an issue for this.

hazelcast/include/hazelcast/util/SyncHttpsClient.h
```diff
@@ -37,14 +37,16 @@ class HAZELCAST_API SyncHttpsClient
 public:
     SyncHttpsClient(const std::string& server_ip,
                     const std::string& uri_path,
-                    std::chrono::steady_clock::duration timeout);
+                    std::chrono::steady_clock::duration timeout,
+                    const std::string& secret_removal = std::string{});
```
ihsandemir, line 41: Reference to a temporary object???

OzanCansel: It is completely safe. [See ref.](https://en.cppreference.com/w/cpp/language/reference_initialization#Lifetime_of_a_temporary)

[An example](https://godbolt.org/z/zdrovvePT)

hazelcast/test/src/HazelcastTests7.cpp
OzanCansel: I will open after merge with the link of the test.

ihsandemir (approved):

PR #1045, opened 2022-12-09
Coordinator url changed to viridan. [API-1765]

OzanCansel:
- Coordinator url changed to `api.viridian.hazelcast.com`


ihsandemir (approved):
Can you also please add the workflow at [here](https://github.com/hazelcast/client-compatibility-suites/tree/master/.github/workflows) for C++ client and let's run the client with that workflow?

PR #1046, opened 2022-12-12
Fix for const qualified warnings

cngzhnp:
When Hazelcast was built up from scratch for our project, we will use our warning flags for this subproject as well.

So, this is also cleanup for Hazelcast library.

Please review if possible, thanks in advance.


ihsandemir:
verify

ihsandemir:
Can you please describe which warnings do these changes fix? There were good reasons for making those as const, hence we really need to understand what the need is here. It also means public API change and we have to be backward compatible.

cngzhnp:
Actually, it is known warning raised by the most of compilers. `const` on primitive types has no direct effect for a return value, because you can't modify an rvalue of a primitive type. If you compile Hazelcast source code with `"-Wextra"` flag, you will get the same warning as we did.

ihsandemir (approved):
Looks good. @OzanCansel 

Can we add the flags to our builds including PR builder? If no objection, should we merge this PR?

OzanCansel:
`-Wextra` causes other types of warnings which would make our builds failed since we treat the warnings as errors. This PR fixes `-Wignored-qualifiers` and `-Wtype-limits` warnings so I can add these specialized warning flags.

Yes, I agree to @cngzhnp, we can merge after all checks have been completed.

g-donev:
Hi @cngzhnp. Thanks a lot for your contribution! 
The Clients team at Hazelcast would be very interested learn more about your project.
Would you like to join us for a short call to say hi and exchange some thoughts?
Regards,
Georgi

OzanCansel:
verify

OzanCansel:
verify

OzanCansel:
verify

ihsandemir:
Thanks @cngzhnp for the contribution.

PR #1059, opened 2022-12-21
Api 1794

OzanCansel:
- `jet` is not included in cluster config if server version is lower than `5.0.0`.
- `cp-member-count` tag is specified for earlier versions server because otherwise it doesn't start the cluster.
- `TestIssue1005` test is refactored to setup 3 members.
- `TestIssue1005` forgotten `.get()` calls to future are added.
- `testListenerSubscriptionOnSingleServerRestart` is improved.
- `cp_test::TearDown` forgotten `.get()` call to future is added.
- `multiplication` class doesn't exist before `4.1.0` version, so `basic_atomic_long_test.test_alter`, `basic_atomic_long_test.test_alter_and_get`, `basic_atomic_long_test.test_get_and_alter`, `basic_atomic_long_test.test_apply` are skipped.
- `append_string` class doesn't exist before `4.1.0` version, so `basic_atomic_ref_test.test_apply` is skipped.
- `basic_latch_test.tets_get_count`, `basic_latch_test.test_wait_until` and `basic_latch_test.tets_get_count.test_wait` are refactored because detached thread wasn't waited so it was causing `hazelcast_client_not_active` on detached thread.
- `member::version cluster_version()` is as free function to `remote_controller_client.h`.
- Mapping queries are not executing before `5.0.0`
- `SqlTest.calling_iterator_next_consecutively` is improved, behaviour is not changed.
- Some SQL tests are skipped due to unsupported SQL clauses.
- `start-rc.sh` is refactored to work with `4.0.x` and `4.1.x` versions. It doesn't download `sql.jar` if the version is lower than `4.2.0` and also it doesn't include `hazelcast-sql-VERSION.jar` in the `CLASSPATH`.

[Compatiblilty tests](https://github.com/hazelcast/client-compatibility-suites/actions/runs/3750112735) are passed. Reliable topic tests are sometimes failed but they are known issue.


hazelcast/test/src/sql_test.cpp
```diff
@@ -752,6 +783,10 @@ TEST_F(SqlTest, simple)
 
 TEST_F(SqlTest, rows_can_be_used_even_after_the_result_is_destroyed)
 {
+    // `VALUES` clause is not supported before 5.0.0
+    if (cluster_version() < member::version{ 5, 0, 0 })
```
ihsandemir, line 787: Why not put these into `SetUp` method? Can we eliminate repetition with this?

scripts/start-rc.bat
```diff
@@ -22,6 +22,7 @@ if not "x%HZ_VERSION:SNAPSHOT=%"=="x%HZ_VERSION%" (
 
 echo Downlading jars for remote controller server
 
+del "hazelcast-remote-controller-%HAZELCAST_RC_VERSION%.jar"
```
ihsandemir, line 25: left over?

ihsandemir (changes requested):

scripts/start-rc.bat
OzanCansel: Fixed.

hazelcast/test/src/sql_test.cpp
OzanCansel: Some specific tests should be disabled, `SetUp()` is not good place for it.

ihsandemir (approved):

PR #1073, opened 2022-12-29
Crash problem is fixed when an adress is not found in private_to_public map.

OzanCansel:
- If network_address_provider cannot translate the adress, translate method throws an exception.
- Unit test is added.

### DIFF
--- a/hazelcast/test/src/HazelcastTests8.cpp
+++ b/hazelcast/test/src/HazelcastTests8.cpp
@@ -37,6 +37,7 @@
 #include <hazelcast/client/connection/ClientConnectionManagerImpl.h>
 #include <hazelcast/client/connection/Connection.h>
 #include <hazelcast/client/connection/AddressProvider.h>
+#include <hazelcast/client/spi/impl/discovery/remote_address_provider.h>
 #include <hazelcast/client/entry_event.h>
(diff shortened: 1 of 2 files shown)


hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -2316,6 +2317,21 @@ TEST_F(connection_manager_translate, default_config_uses_private_addresses)
...
+{
+    auto client = new_client().get();
+    spi::ClientContext ctx(client);
+    connection::ClientConnectionManagerImpl connection_manager(
+      ctx,
+      std::unique_ptr<connection::AddressProvider>(
+        new spi::impl::discovery::remote_address_provider{
+          []() { return std::unordered_map<address, address>{}; }, true }));
+
+    EXPECT_THROW(connection_manager.get_or_connect(hazelcast::client::member{}),
```
ihsandemir, line 2332: can you put a dummy address instead of empty constructor to be more realistic?

OzanCansel: Fixed.

ihsandemir (approved):

ihsandemir (approved):

Issue #1071, opened 2022-12-28
 ReliableTopicTest.testMessageFieldSetCorrectly occasionally crashes on github actions. [API-1809]


OzanCansel:
C++ compiler version: msvc-14.2
Hazelcast Cpp client version: 5.0.0
Hazelcast server version: 5.1.3
Number of the clients: -
Cluster size, i.e. the number of Hazelcast cluster members: -
OS version (Windows/Linux/OSX): Windows

Please attach relevant logs and files for client and server side.

#### Expected behaviour
Pass test

#### Actual behaviour
Crashes on test

#### Steps to reproduce the behaviour
It might be a different test but this is what I suspected.
Here is the [link](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/3784657210/jobs/6434122371).
[ReliableTopicTest_testMessageFieldSetCorrectly.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/10313081/ReliableTopicTest_testMessageFieldSetCorrectly.txt)

srknzl:
This failed again on [Test CPP client v5.1.0 with enterprise server](https://github.com/hazelcast/client-compatibility-suites/actions/runs/6535724054/job/17771144765#logs) in the client compat. tests of 5.3.3 HZ release. https://github.com/hazelcast/client-compatibility-suites/actions/runs/6535724054/job/17771144765#logs


```
[ RUN      ] ReliableTopicTest.testMessageFieldSetCorrectly
08:42:58,116  INFO [Main] - Ping ... Pong ...
Oct 17, 2023 8:42:58 AM com.hazelcast.internal.server.tcp.TcpServerConnection
WARNING: [localhost]:5701 [dev] [5.3.3-SNAPSHOT] Connection[id=1, /127.0.0.1:5701->/127.0.0.1:35360, qualifier=null, endpoint=[127.0.0.1]:35360, remoteUuid=943d04dc-d7d5-4720-a84f-882823ece7cf, alive=false, connectionType=CPP, planeIndex=-1] closed. Reason: Exception in Connection[id=1, /127.0.0.1:5701->/127.0.0.1:35360, qualifier=null, endpoint=[127.0.0.1]:35360, remoteUuid=943d04dc-d7d5-4720-a84f-882823ece7cf, alive=true, connectionType=CPP, planeIndex=-1], thread=hz.nifty_faraday.IO.thread-in-0
java.io.IOException: Connection reset by peer
[...]

PR #1087, opened 2023-02-02
[API-1809]: Executor Lifecycle is extended for reliable_topic async continuation

### LINKED ISSUE #1079: ReliableTopicTest.testConfig occassionally crashes on github actions. [API-1816]
C++ compiler version: vc-14.2
Hazelcast server version: 5.1.3
Number of the clients:
OS version (Windows/Linux/OSX):
Windows

#### Expected behaviour
Pass test

#### Actual behaviour
Crashes on test

Here is the [link](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/3814810754/jobs/6489384465)
[ReliableTopicTest_testConfig.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/10330385/ReliableTopicTest_testConfig.txt)

akeles85:
For reliable_topic, in MessageRunner class, the next function is using boost async continuation with executor. When the client is destroyed, the executor object is destroyed too. If an async call is done and in the meanwhile the client is destroyed, an invalid executor is being used by boost and segfault is occured. To solve this problem, the lifecycle of executor is extended with passing the execution service as a parameter of async call.

To test the fix, first the test case is changed. testConfig test case is done for multiple times, as it destroys the client. When we add 10 test cases with testConfig, the error started to occur nearly every run. After the fix, we did not see the problem for more than 10 test cases. There is an example of test results in the attached file.

Fixes #1079, #1071 

[reliable_topic_test.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/10566242/reliable_topic_test.txt)

### DIFF
--- a/hazelcast/include/hazelcast/client/reliable_topic.h
+++ b/hazelcast/include/hazelcast/client/reliable_topic.h
@@ -159,7 +159,7 @@ private:
           , cancelled_(false)
           , logger_(lg)
           , name_(topic_name)
-          , execution_service_(execution_service)
+          , execution_service_(std::move(execution_service))
           , executor_(executor)
           , serialization_service_(service)
           , batch_size_(batch_size)
--- a/hazelcast/include/hazelcast/client/reliable_topic.h
+++ b/hazelcast/include/hazelcast/client/reliable_topic.h
@@ -104,6 +104,7 @@ public:
                                       get_serialization_service(),
                                       batch_size_,
                                       logger_,
+                                      execution_service_,
                                       executor_,
                                       runners_map_));
         runners_map_.put(id, runner);
@@ -147,6 +148,8 @@ private:
                       serialization::pimpl::SerializationService& service,
                       int batch_size,
                       logger& lg,
+                      std::shared_ptr<spi::impl::ClientExecutionServiceImpl>
+                        execution_service,                      
                       util::hz_thread_pool& executor,
                       util::SynchronizedMap<int, util::concurrent::Cancellable>&
                         runners_map)
@@ -156,6 +159,7 @@ private:
           , cancelled_(false)
           , logger_(lg)
           , name_(topic_name)
+          , execution_service_(execution_service)
           , executor_(executor)
           , serialization_service_(service)
           , batch_size_(batch_size)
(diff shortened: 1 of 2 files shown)


hazelcast/include/hazelcast/client/reliable_topic.h
```diff
@@ -156,6 +159,7 @@ class HAZELCAST_API reliable_topic : public proxy::ProxyImpl
           , cancelled_(false)
           , logger_(lg)
           , name_(topic_name)
+          , execution_service_(execution_service)
```
OzanCansel, line 162: `execution_service_` instance will be copy-constructed rather than move constructed. So I suggest to change it as :

```suggestion
          , execution_service_(move(execution_service))
```

hazelcast/src/hazelcast/client/proxy.cpp
```diff
@@ -39,6 +39,8 @@ const std::chrono::milliseconds imap::UNSET{ -1 };
 reliable_topic::reliable_topic(const std::string& instance_name,
                                spi::ClientContext* context)
   : proxy::ProxyImpl(reliable_topic::SERVICE_NAME, instance_name, context)
+  , execution_service_(
+      context->get_client_execution_service().shared_from_this())  
   , executor_(context->get_client_execution_service().get_user_executor())
```
OzanCansel, line 44: `execution_service_` is usable at the construction of `executor_` so I suggest to change it as :

```suggestion
  , executor_(execution_service->get_user_executor())
```
****
It is not crucial but it makes less verbose.

OzanCansel (changes requested):

hazelcast/src/hazelcast/client/proxy.cpp
akeles85: Fixed

hazelcast/include/hazelcast/client/reliable_topic.h
akeles85: Fixed

OzanCansel (approved):

ihsandemir (approved):

PR #1089, opened 2023-02-08
Executor pool size

akeles85:
set_executor_pool_size has no effect in the code. It justs sets a variable and this variable is never used anywhere. In this PR, user thread pool size can be set with set_executor_pool_size function. The PR is tested with written test cases.

### DIFF
--- a/hazelcast/test/src/HazelcastTests8.cpp
+++ b/hazelcast/test/src/HazelcastTests8.cpp
@@ -25,6 +25,7 @@
 #include <vector>
 
 #include <boost/asio.hpp>
+#include <boost/thread/barrier.hpp>
 
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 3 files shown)


hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -2344,6 +2344,87 @@ TEST_F(
...
+
+    spi::ClientContext ctx(*client);
+    auto state = std::make_shared<ThreadState>(num_of_jobs);
+    std::mutex mutex_for_thread_id;
+
+    ASSERT_EQ(0, state->thread_ids.size());
+    for (int i = 0; i < num_of_jobs; i++) {
+        ctx.get_client_execution_service().get_user_executor().submit(
+          [state, &mutex_for_thread_id]() {
+              std::this_thread::sleep_for(std::chrono::seconds(1));
```
ihsandemir, line 2407: why sleep?

akeles85: to be sure that the job takes some time. It may go to the same thread if the job instantly returns.

ihsandemir: This is still time sensitive and may fail on GA runs depending on machine speed. Can we make it more time insensitive? E.g. inside the thread we can have all threads wait on a latch and sync each other and then continue.

akeles85: I removed the sleep method and used a barrier to sync the threads.

ihsandemir (approved):

PR #1095, opened 2023-02-09
Compact test failures

### LINKED ISSUE #1090: Compact Test Case Failures in Github Actions [API-1871]
C++ compiler version: clang 14
Hazelcast server version: 5.1.3
Number of the clients:
OS version (Windows/Linux/OSX): All

#### Expected behaviour
Pass test

#### Actual behaviour
Failed tests

Here is the [link](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/4119357506/jobs/7112925109).
[compact_failed_test.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/10685192/compact_failed_test.txt)

akeles85:
Fixes #1090 
As I observed, the problem occurs in static library. It does not occur in dynamic library.
The problem is occured because of the static member initialization order. As there are template schemas, the build_schema function is called at startup. As finger print table is also initialized at startup, the schemas are built with empty table.

### DIFF
--- a/hazelcast/src/hazelcast/client/compact.cpp
+++ b/hazelcast/src/hazelcast/client/compact.cpp
@@ -1489,6 +1489,8 @@ field_descriptor::field_descriptor(enum field_kind f,
 std::array<uint64_t, 256>
 init_fp_table()
 {
+    std::mutex tmp_mutex;
+    std::lock_guard<std::mutex> lg(tmp_mutex);
     static std::array<uint64_t, 256> FP_TABLE;
     for (int i = 0; i < 256; ++i) {
         uint64_t fp = i;
@@ -1500,12 +1502,12 @@ init_fp_table()
     return FP_TABLE;
 }
 
-static std::array<uint64_t, 256> FP_TABLE = init_fp_table();
 constexpr uint64_t rabin_finger_print::INIT;
 
 uint64_t
 rabin_finger_print::fingerprint64(uint64_t fp, byte b)
-{
+{    
+    static std::array<uint64_t, 256> FP_TABLE = init_fp_table();
     return (fp >> 8) ^ FP_TABLE[(int)(fp ^ b) & 0xff];
 }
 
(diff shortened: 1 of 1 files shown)


hazelcast/src/hazelcast/client/compact.cpp
```diff
@@ -1489,6 +1489,8 @@ field_descriptor::field_descriptor(enum field_kind f,
 std::array<uint64_t, 256>
 init_fp_table()
 {
+    std::mutex tmp_mutex;
+    std::lock_guard<std::mutex> lg(tmp_mutex);
```
OzanCansel, line 1493: Lock is not required, static initialization is already thread safe,

```suggestion
```

OzanCansel (changes requested):

hazelcast/src/hazelcast/client/compact.cpp
akeles85: I put it because of the static FP_TABLE initialization in init_fp_table. But maybe it will be better to remove both mutex and static keyword in this function.

OzanCansel (approved):

hazelcast/src/hazelcast/client/compact.cpp
```diff
@@ -1500,12 +1500,12 @@ init_fp_table()
     return FP_TABLE;
 }
 
-static std::array<uint64_t, 256> FP_TABLE = init_fp_table();
 constexpr uint64_t rabin_finger_print::INIT;
 
 uint64_t
 rabin_finger_print::fingerprint64(uint64_t fp, byte b)
 {
+    static std::array<uint64_t, 256> FP_TABLE = init_fp_table();
```
ihsandemir, line 1508: will it create a table for each call?

akeles85: It should be called just once as it is a static variable. So one table is created.

ihsandemir (approved):

PR #1104, opened 2023-02-22
BaseSocket async_write buffer lifecycle [API-1814] [API-1808]

akeles85:
It seems that there is a problem with the lifecycle of invocations. When async_write method of the boost library is called, user should guarantee the validity of the buffer until the write operation complete. The invocations are added to an unordered_map with add_invocation_to_map method and erased when the response of the message is received or connection is closed. The problem occurs when the invocation is deleted as the response is received, and after that the write handler completion occurs. As the buffer is already deleted, boost async_write completion runs with deleted buffers and error occurs. We think that this case occurs when an async_write occurs while there is already a message in outbox.


hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -336,7 +338,12 @@ class BaseSocket : public hazelcast::client::socket
     boost::asio::ip::tcp::resolver& resolver_;
     T socket_;
     int32_t call_id_counter_{ 0 };
-    typedef std::deque<std::vector<boost::asio::const_buffer>> Outbox;
+    struct outbox_entry_t{
```
OzanCansel, line 341: `_t` suffix appended is generally appended if it is alias of a type. So I would recommend to delete it. For sake of simplicity, `entry` is pretty understandable.

```suggestion
    struct entry
```

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -263,12 +265,12 @@ class BaseSocket : public hazelcast::client::socket
...
 
                   if (!this->outbox_.empty()) {
-                      do_write(connection, invocation);
+                      do_write(connection, outbox_[0].invocation);
                   }
               }
           };
-
-        const auto& message = outbox_[0];
+
```
OzanCansel, line 272:
```suggestion

```

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -125,12 +125,14 @@ class BaseSocket : public hazelcast::client::socket
             add_invocation_to_map(connection, invocation, message);
 
             auto& datas = message->get_buffer();
-            std::vector<boost::asio::const_buffer> buffers;
-            buffers.reserve(datas.size());
+
+            struct outbox_entry_t outbox_entry;
```
OzanCansel, line 129: `struct` keyword is not needed here.

```suggestion
            outbox_entry_t outbox_entry;
```

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -125,12 +125,14 @@ class BaseSocket : public hazelcast::client::socket
...
+            outbox_entry.buffers.reserve(datas.size());
             for (const auto& data : datas) {
-                buffers.emplace_back(boost::asio::buffer(data));
+                outbox_entry.buffers.emplace_back(boost::asio::buffer(data));
             }
-            this->outbox_.push_back(buffers);
+            outbox_entry.invocation = invocation;
+            this->outbox_.push_back( outbox_entry );
```
OzanCansel, line 135: `outbox_entry` should be moved to eliminate copy cost.

```suggestion
            this->outbox_.push_back(std::move(outbox_entry));
```

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -263,12 +265,12 @@ class BaseSocket : public hazelcast::client::socket
                   connection->last_write_time(std::chrono::steady_clock::now());
 
                   if (!this->outbox_.empty()) {
-                      do_write(connection, invocation);
+                      do_write(connection, outbox_[0].invocation);
```
OzanCansel, line 268: I think it is no longer needed to pass `invocation` to the `do_write` method since its lifetime is already extended by appending `outbox_` queue. So I think it is better to remove `invocation` parameter from `do_write` function.

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -125,12 +125,14 @@ class BaseSocket : public hazelcast::client::socket
...
-            buffers.reserve(datas.size());
+
+            struct outbox_entry_t outbox_entry;
+            outbox_entry.buffers.reserve(datas.size());
             for (const auto& data : datas) {
-                buffers.emplace_back(boost::asio::buffer(data));
+                outbox_entry.buffers.emplace_back(boost::asio::buffer(data));
             }
-            this->outbox_.push_back(buffers);
+            outbox_entry.invocation = invocation;
```
OzanCansel, line 134: If `invocation` parameter is removed from the `do_write` method, `invocation` becomes unnecessary after this line. So I suggest it to be moved. Btw, lambda object should be marked as `mutable` to make `move` operation valid.

```suggestion
            outbox_entry.invocation = std::move(invocation);
```

OzanCansel (changes requested):
As a general comment, could you run `git-clang-format master` to format ?

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
akeles85: fixed

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
akeles85: fixed

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
akeles85: fixed

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
akeles85: fixed

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
akeles85: fixed

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
akeles85: fixed

OzanCansel (dismissed):

OzanCansel (approved):

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -336,7 +338,14 @@ class BaseSocket : public hazelcast::client::socket
     boost::asio::ip::tcp::resolver& resolver_;
     T socket_;
     int32_t call_id_counter_{ 0 };
-    typedef std::deque<std::vector<boost::asio::const_buffer>> Outbox;
+    struct entry
+    {
+        std::vector<boost::asio::const_buffer> buffers;
+        std::shared_ptr<spi::impl::ClientInvocation> invocation;
+        std::shared_ptr<protocol::ClientMessage> message;
```
ihsandemir, line 345: The `message` should be accessible via invocation, right? why keep it separate?

akeles85: because of the retry mechanism, the message might be deleted from invocation. That is why message is hold in this structure.

PR #1118, opened 2023-02-27
[API-1511] -DHZ_BUILD_WITH_SSL requirement is stated.

OzanCansel:
`HZ_BUILD_WITH_SSL` symbolic constant has been added.

### DIFF
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -320,6 +320,14 @@ For example, if you want to build the static library with SSL support, you can u
 cmake .. -DWITH_OPENSSL=ON -DBUILD_SHARED_LIBS=OFF
 ```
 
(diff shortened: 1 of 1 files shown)


Reference_Manual.md
```diff
@@ -320,6 +320,14 @@ For example, if you want to build the static library with SSL support, you can u
 cmake .. -DWITH_OPENSSL=ON -DBUILD_SHARED_LIBS=OFF
 ```
 
+Note that, if you want to use the `hazelcast-cpp-client` library which is compiled with `-DWITH_OPENSSL=ON` option without `find_package()` then you need to define `HZ_BUILD_WITH_SSL` symbolic constant before including any `hazelcast-cpp-client` header.
+It can be either passed via compiler flags.
+
+For example
```
akeles85, line 326:
```suggestion
For example:
```

akeles85 (approved):

akeles85 (approved):

PR #1122, opened 2023-02-27
1114 reliabletopic terminate test failure [AP-1893]

### LINKED ISSUE #1114: ReliableTopic Terminate Test Failure [API-1893]
C++ compiler version: gcc 7
Hazelcast server version: 5.2.0
Number of the clients: 1
OS version (Windows/Linux/OSX): Linux (Ubuntu i386)

#### Expected behaviour
Terminate test case for reliable topic should pass.

#### Actual behaviour
Terminate test case for reliable topic causes a segmantation fault.

It just occurs for Ubuntu i386 randomly. Here is the log of the github actions.
[reliable_topic_terminate.txt](https://github.com/hazelcast/hazelcast-cpp-client/files/10837302/reliable_topic_terminate.txt)
Link https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/4277129180/jobs/7445727007

akeles85:
Fixes #1114 

MessageRunner is created with add_message_listener and it uses runners_map_ of its parent listener topic. When an exception occurs on user code, message runner calls the terminate function and then calls the cancel method which deletes the runner from runners_map_. Everything is ok up to now. But if this case occurs while the reliable topic is deleted, MessageRunner calls the cancel method and access the deleted runners_map_, it gives a seg fault. This situation may also occur, when a message is received, while just deleting a topic. So the lifetime of owner of runners_map_ should be extended with MessageRunner continuation.

After repeating the terminate test case, the test failure becomes every run. After fixing, test always becomes success.

This error started to occur randomly at github actions after the addition of new test cases.

### DIFF
--- a/hazelcast/include/hazelcast/client/reliable_topic.h
+++ b/hazelcast/include/hazelcast/client/reliable_topic.h
@@ -107,8 +107,7 @@ public:
                                       batch_size_,
                                       logger_,
                                       execution_service_,
-                                      executor_,
-                                      runners_map_,
+                                      executor_,                                      
                                       shared_from_this()));
         runners_map_.put(id, runner);
         runner->next();
@@ -154,9 +153,7 @@ private:
                       std::shared_ptr<spi::impl::ClientExecutionServiceImpl>
                         execution_service,
                       util::hz_thread_pool& executor,
-                      util::SynchronizedMap<int, util::concurrent::Cancellable>&
-                        runners_map,
-                      std::shared_ptr<reliable_topic> topic)
+                      std::weak_ptr<reliable_topic> topic)
           : listener_(listener)
           , id_(id)
           , ringbuffer_(rb)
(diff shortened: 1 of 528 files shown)


hazelcast/include/hazelcast/client/reliable_topic.h
```diff
@@ -415,6 +420,7 @@ class HAZELCAST_API reliable_topic : public proxy::ProxyImpl
         serialization::pimpl::SerializationService& serialization_service_;
         int batch_size_;
         util::SynchronizedMap<int, util::concurrent::Cancellable>& runners_map_;
+        std::shared_ptr<reliable_topic> topic_;
```
OzanCansel, line 423: Could you make it `weak_ptr` since it is not required to be alive to continue work.

OzanCansel (changes requested):

hazelcast/include/hazelcast/client/reliable_topic.h
akeles85: Shared_ptr is changed with weak_ptr and runners_map_ is removed from MessageRunner. It is access via topic_ pointer.

OzanCansel (approved):

srknzl:
@ihsandemir I think we should backport this if we haven't already. It failed on server backward compat test of 5.3.2 https://github.com/hazelcast/client-compatibility-suites/actions/runs/5876916830/job/15936429468

PR #1127, opened 2023-03-01
local_date parse bug is fixed. Tests are added.

### LINKED ISSUE #1125: Wrong SQL TIMESTAMP column value [API-1914]
@OzanCansel 's Using ```inline std::ostream& operator<<(std::ostream& os, const hazelcast::client::local_date_time& dt)
{
    return os << int(dt.date.year)
              << "-"
              << int(dt.date.month)
              << "-"
              << int(dt.date.day_of_month)
              << "T"
              << int(dt.time.hours)
              << ":"
              << int(dt.time.minutes)
              << ":"
              << int(dt.time.seconds)
			  ;
}```

prints 
```
231-2-28T13:19:41
```
CCYY is wrong, should be 2023.

Sql column value is read wrong.

OzanCansel:
`local_date` is parsed wrongly in `ClientMessage` it is fixed.

Relevant tests are added to `sql_test.cpp` as well.

Closes #1125


hazelcast/test/src/ClientTest.h
```diff
@@ -88,6 +91,50 @@ class ClientTest : public ::testing::Test
...
+      const std::string& map_name,
+      const std::string& cluster_id,
+      int n_entries = 10,
+      ValueGenerator&& value_gen = ValueGenerator{},
+      KeyGenerator&& key_gen = KeyGenerator{})
+    {
+        std::unordered_map<Key, Value> entries;
+        entries.reserve(n_entries);
+
+        generate_n(
```
akeles85, line 109: std namespace can be used before generate_n and inserter methods.

akeles85 (approved):

hazelcast/test/src/ClientTest.h
OzanCansel: I intentionally don't use those qualifiers, if they are unnecessary

akeles85 (approved):

PR #1130, opened 2023-03-02
context is not protected, then is made sync

akeles85:
In github actions, cp semaphore tests are failed sometimes. In windows dump file, we saw that the seg fault is occured at then continuation. As we review the code, context lifecycle is not protected. As there is no user code, instead of extending lifecycle, it is also made sync. So that the client will not destroyed until this continuation.


OzanCansel (changes requested):
Could you run `git-clang-format` for this change ?

akeles85:
> Could you run `git-clang-format` for this change ?

done

akeles85:
This PR will be closed, solution to the problem will be changed.

PR #1132, opened 2023-03-03
Listener unhandled exception

akeles85:
When an exception occurs in invoke_urgent().then, it is not handled with futures's get method. Because of this, assert method may be called in invoke_on_selection and gives error in debug releases.

### DIFF
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -1572,7 +1572,7 @@ ClientInvocation::invoke_on_selection()
         }
     } catch (exception::iexception&) {
         notify_exception(std::current_exception());
-    } catch (std::exception&) {
+    } catch (std::exception& exp) {
         HZ_LOG(
           logger_,
           finest,
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -1573,11 +1573,10 @@ ClientInvocation::invoke_on_selection()
     } catch (exception::iexception&) {
         notify_exception(std::current_exception());
     } catch (std::exception&) {
-        HZ_LOG(logger_,
-                   finest,
-                   boost::str(boost::format(
-                                "Unexpected exception. %1%") %
-                              exp.what() ));        
+        HZ_LOG(
+          logger_,
+          finest,
+          boost::str(boost::format("Unexpected exception. %1%") % exp.what()));
         assert(false);
     }
 }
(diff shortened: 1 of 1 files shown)


hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1572,7 +1572,11 @@ ClientInvocation::invoke_on_selection()
         }
     } catch (exception::iexception&) {
         notify_exception(std::current_exception());
-    } catch (std::exception&) {
+    } catch (std::exception& exp) {
+        HZ_LOG(
+          logger_,
+          finest,
```
OzanCansel, line 1578: This line is executed just before termination, so `finest` is not appropriate log type. Also, I don't think these lines are related with this bug. It can even be removed (if it is OK for you). But at least, it needs to be `severe` log type.

```suggestion
          severe,
```

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -3006,6 +3010,13 @@ cluster_view_listener::try_register(
               return;
           }
 
+          try {
+              f.get();
+          } catch (exception::hazelcast_client_not_active& e) {
+              /*If client is shutdown, we should not retry for another
+               * connection*/
```
OzanCansel, line 3017:
```suggestion
              /**
               * If client is shutdown, we should not retry for another
               * connection
               */
```

OzanCansel (changes requested):
Empty comment

hazelcast/src/hazelcast/client/spi.cpp
akeles85: log is removed

hazelcast/src/hazelcast/client/spi.cpp
akeles85: fixed

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1572,7 +1572,7 @@ ClientInvocation::invoke_on_selection()
         }
     } catch (exception::iexception&) {
         notify_exception(std::current_exception());
-    } catch (std::exception&) {
+    } catch (std::exception& exp) {
```
OzanCansel, line 1575: `exp` is unused.

```suggestion
    } catch (std::exception&) {
```

OzanCansel (changes requested):

hazelcast/src/hazelcast/client/spi.cpp
akeles85: fixed

OzanCansel (approved):

PR #1134, opened 2023-03-04
Fix pedantic warnings for cleaning up

cngzhnp:
When building up Hazelcast with -Wpedantic flag, there are some places which needs to be fixed. 

This MR contains fix for only this flag, it would be extended for other flags like -Wunused or -Wreorder etc.

Please review my changes,

Thanks.


akeles85:
Thanks for your contribution. As it can be seen in your PR, there are some unnecessary semicolons. The library can be compiled with -Wall flag which contains -Wunused and -Wreorder flags.

OzanCansel:
verify

cngzhnp:
> Thanks for your contribution. As it can be seen in your PR, there are some unnecessary semicolons. We can remove them. On the other hand, the library can be compiled with -Wall flag which contains -Wunused and -Wreorder flags.

Thanks, I also try to compile with Wall and Wextra flags. However, there are a lot of warnings which could be hard to review at once. So, I would like to introduce first only pedantic warnings. As a next step, I would like to do in that way.

akeles85:
> compile with Wall and Wextra flags. However, there are a lot of warnings which could be

with -Wall there seems no warning at least for clang 14 and gcc 11. As you mentioned, there are some warnings with -Wextra, I checked them and lots of them are because of unused-parameter warning type. We would be glad to see your PR about these warnings.

akeles85 (approved):

cngzhnp:
> > compile with Wall and Wextra flags. However, there are a lot of warnings which could be
> 
> with -Wall there seems no warning at least for clang 14 and gcc 11. As you mentioned, there are some warnings with -Wextra, I checked them and lots of them are because of unused-parameter warning type. We would be glad to see your PR about these warnings.

Created PR #1138 for this, could you please review & verify it, @akeles85 ?

OzanCansel:
verify

PR #1135, opened 2023-03-06
Include iosfwd instead of xstream system header file

cngzhnp:
In header file, if all types are incomplete, actually it is not recommended that `<iostream>`, `<sstream>`, `<fstream>`, `<istream>` or `<sstream>` is used. 

So, in this MR, I'd like to change these header files to `iosfwd` whose size is less than others and used for forward declaration.


OzanCansel:
verify

OzanCansel:
verify

OzanCansel (approved):

PR #1138, opened 2023-03-06
Fix unused parameter warnings for all components

### LINKED ISSUE #1137: NearCacheConfig constructor does not take the correct parameter
C++ compiler version: GNU 12.2.0
Hazelcast server version: -
Number of the clients: -

#### Expected behaviour
One of the constructor of near_cache_config class has two parameters. The first parameter was used to assign the cache name into class member _name field via base constructor.

#### Actual behaviour
However, this constructor takes default class member name instead of parameter. It should be fixed for this class. Solution must be done like below, IMHO.

```cpp
near_cache_config::near_cache_config(const std::string& cache_name,
                                     in_memory_format memory_format)
  : near_cache_config(cache_name)
{
    this->in_memory_format_ = memory_format;
}
```

cngzhnp:
Fix for unused parameter warnings that mentioned in previous PR. This patch also fixes #1137 as well.


hazelcast/include/hazelcast/client/connection/Connection.h
```diff
@@ -70,7 +69,6 @@ class HAZELCAST_API Connection
                spi::ClientContext& client_context,
                int32_t connection_id,
                internal::socket::SocketFactory& socket_factory,
-               ClientConnectionManagerImpl& client_connection_manager,
```
akeles85, line 73: If a public API needs to be changed, there should be a real reason. As Connection is public method, we should not remove the parameter to remove the warning. We might something like this (void)client_connection_manager; to suppress the warning.

hazelcast/include/hazelcast/client/internal/eviction/EvictableStore.h
```diff
@@ -64,8 +64,8 @@ class EvictableStore
      */
     virtual int evict(
       std::vector<std::shared_ptr<EvictionCandidate<MAPKEY, MAPVALUE, A, E>>>*
-        eviction_candidates,
```
akeles85, line 67: public API, we should not change the decleration.

hazelcast/include/hazelcast/client/spi/impl/ClientExecutionServiceImpl.h
```diff
@@ -40,8 +40,7 @@ class HAZELCAST_API ClientExecutionServiceImpl
   : public std::enable_shared_from_this<ClientExecutionServiceImpl>
 {
 public:
-    ClientExecutionServiceImpl(const std::string& name,
-                               const client_properties& properties,
+    ClientExecutionServiceImpl(const client_properties& properties,
```
akeles85, line 43: As this is a public API, we should not change the signature. Could you revert this change?

hazelcast/include/hazelcast/client/internal/nearcache/impl/store/BaseHeapNearCacheRecordStore.h
```diff
@@ -49,7 +49,6 @@ class BaseHeapNearCacheRecordStore
       ANCRS;
 
     BaseHeapNearCacheRecordStore(
-      const std::string& name,
```
akeles85, line 52: Could we revert this change? As it is public method, we should not change the signature.

hazelcast/include/hazelcast/client/internal/nearcache/impl/store/NearCacheDataRecordStore.h
```diff
@@ -49,14 +49,12 @@ class NearCacheDataRecordStore
     static const int64_t REFERENCE_SIZE =
       sizeof(std::shared_ptr<serialization::pimpl::data>);
 
-    NearCacheDataRecordStore(const std::string& name,
```
akeles85, line 52: Could we revert this change? As it is public method, we should not change the signature. We might use another way to suppress the warning.

hazelcast/include/hazelcast/client/internal/nearcache/impl/store/NearCacheDataRecordStore.h
```diff
@@ -49,14 +49,12 @@ class NearCacheDataRecordStore
...
-                             const client::config::near_cache_config& config,
+    NearCacheDataRecordStore(const client::config::near_cache_config& config,
                              serialization::pimpl::SerializationService& ss)
       : BaseHeapNearCacheRecordStore<K,
                                      V,
                                      serialization::pimpl::data,
-                                     record::NearCacheDataRecord>(name,
```
akeles85, line 58: Could we revert this change? As it is public method, we should not change the signature. We might use another way to suppress the warning.

hazelcast/include/hazelcast/client/internal/nearcache/impl/store/NearCacheObjectRecordStore.h
```diff
@@ -46,27 +46,25 @@ class NearCacheObjectRecordStore
       HeapNearCacheRecordMap<K, V, KS, record::NearCacheObjectRecord<V>>>
       ANCRS;
 
-    NearCacheObjectRecordStore(const std::string& name,
```
akeles85, line 49: Could we revert this change? As it is public method, we should not change the signature. We might use another way to suppress the warning.

hazelcast/include/hazelcast/client/internal/nearcache/impl/store/NearCacheObjectRecordStore.h
```diff
@@ -46,27 +46,25 @@ class NearCacheObjectRecordStore
...
-                               const client::config::near_cache_config& config,
+    NearCacheObjectRecordStore(const client::config::near_cache_config& config,
                                serialization::pimpl::SerializationService& ss)
       : BaseHeapNearCacheRecordStore<K,
                                      V,
                                      serialization::pimpl::data,
-                                     record::NearCacheObjectRecord<V>>(name,
```
akeles85, line 55: We should also revert here.

hazelcast/src/hazelcast/client/client_impl.cpp
```diff
@@ -334,7 +334,6 @@ std::shared_ptr<spi::impl::ClientExecutionServiceImpl>
 hazelcast_client_instance_impl::init_execution_service()
 {
     return std::make_shared<spi::impl::ClientExecutionServiceImpl>(
-      instance_name_,
```
akeles85, line 337: We should also revert here

hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -902,7 +902,7 @@ near_cache_config::near_cache_config(const std::string& cache_name)
 
 near_cache_config::near_cache_config(const std::string& cache_name,
                                      in_memory_format memory_format)
-  : near_cache_config(name_)
+  : near_cache_config(cache_name)
```
akeles85, line 905: It seems there was a bug here and with this change it is fixed :)

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -1063,7 +1063,6 @@ ClientConnectionManagerImpl::connect(const address& addr)
                                                    client_,
                                                    ++connection_id_gen_,
                                                    *socket_factory_,
-                                                   *this,
```
akeles85, line 1066: We should revert this too.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -1199,7 +1198,6 @@ Connection::Connection(
   spi::ClientContext& client_context,
   int connection_id, // NOLINT(cppcoreguidelines-pro-type-member-init)
   internal::socket::SocketFactory& socket_factory,
-  ClientConnectionManagerImpl& client_connection_manager,
```
akeles85, line 1202: We should revert this too.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1334,7 +1334,6 @@ ClientInvocationServiceImpl::fail_on_indeterminate_state() const
 }
 
 ClientExecutionServiceImpl::ClientExecutionServiceImpl(
-  const std::string& name,
```
akeles85, line 1337: We should revert this too.

akeles85 (changes requested):
As a general rule, if a method is public API, we should have a strong reason to change it. If we change it, there might be compatibility issues with client code. In the light of this, I gave my comments for this PR. Could you please review my comments and do the appropriate changes?

akeles85:
verify

akeles85:
By the way, could you please give the details about compiler options and compiler version in the description of PR.

akeles85:
verify

akeles85 (approved):

akeles85:
verify

PR #1142, opened 2023-03-07
Ambiguity fixed. [API-1511]

OzanCansel:
Documentation is improved.

### DIFF
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -320,7 +320,7 @@ For example, if you want to build the static library with SSL support, you can u
 cmake .. -DWITH_OPENSSL=ON -DBUILD_SHARED_LIBS=OFF
 ```
 
-Note that, if you want to use the `hazelcast-cpp-client` library with `-DWITH_OPENSSL=ON` option without `find_package()` then you need to define `HZ_BUILD_WITH_SSL` symbolic constant before including any `hazelcast-cpp-client` header. Those symbolic constants can be defined via compiler options or can be passed directly through cmake as `-DVARIABLE=VALUE` pairs.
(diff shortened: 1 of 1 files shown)


Reference_Manual.md
```diff
@@ -320,8 +320,7 @@ For example, if you want to build the static library with SSL support, you can u
 cmake .. -DWITH_OPENSSL=ON -DBUILD_SHARED_LIBS=OFF
 ```
 
-Note that, if you want to use the `hazelcast-cpp-client` library which is compiled with `-DWITH_OPENSSL=ON` option without `find_package()` then you need to define `HZ_BUILD_WITH_SSL` symbolic constant before including any `hazelcast-cpp-client` header.
-It can be either passed via compiler flags.
+Note that, if you want to use the `hazelcast-cpp-client` library with `-DWITH_OPENSSL=ON` option without `find_package()` then you need to define `HZ_BUILD_WITH_SSL` symbolic constant before including any `hazelcast-cpp-client` header. Those symbolic constants can be defined via compiler options or can be passed directly through cmake as `-DVARIABLE=VALUE` pairs.
```
akeles85, line 323:
```suggestion
Note that, if you want to use the `hazelcast-cpp-client` library with `-DWITH_OPENSSL=ON` option without `find_package()` then you need to define `HZ_BUILD_WITH_SSL` symbolic constant before including any `hazelcast-cpp-client` header. This symbolic constant can be defined via compiler options or can be passed directly through cmake command as `-D_VARIABLE_=_VALUE_` pairs.
```

OzanCansel: Fixed.

akeles85 (approved):

PR #1154, opened 2023-03-13
version is updated to 5.2.0

akeles85:
Version is updated to 5.2.0


OzanCansel (changes requested):
Could you replace `5.1.0` on `Reference_Manual.md` file ?

akeles85:
> Could you replace `5.1.0` on `Reference_Manual.md` file ?

Do you mean the log outputs? I did not change them consciously. I think they are temporary outputs, it would be better not to change. What do you say?

akeles85:
> Could you replace `5.1.0` on `Reference_Manual.md` file ?

5.1.0 in logs is updated as 5.x.x

OzanCansel (approved):

PR #1161, opened 2023-03-15
row_iterator_sync have been implemented. [API-1762]

OzanCansel:
- `row_iterator_sync` is added.


Reference_Manual.md
```diff
@@ -3741,6 +3745,36 @@ std::vector<std::shared_ptr<sql_page>> pages;
 copy(result->pbegin(), result->pend(), back_inserter(pages));
 ```
 
+#### 7.11.8.3 row_iterator_sync
+`row_iterator_sync` is the third iterator which wraps `page_iterator_sync` and allows user to iterate over rows rather than pages.
+It is similar to native C++ iterators. So it can be used with `algorithm` header.
+It supports `operator*()`, `operator->()`, `operator++()` operators. Post increment operator is marked as `delete`.
+`timeout` can be set, it is used when fetching new page. It throws `exception::no_such_element` exception if the fetch operation is timedout.
```
akeles85, line 3752: timed out

examples/sql/sql_row_iterator_sync.cpp
```diff
@@ -0,0 +1,217 @@
...
+    // `generate_stream(1)` generates a row per seconds, so it will guaranteed
+    // that it will timeout
+    auto result =
+      client.get_sql().execute("SELECT * FROM TABLE(generate_stream(1))").get();
+
+    auto it = result->begin(std::chrono::milliseconds{ 1 });
+
+    try {
+        ++it;
+        ++it;
```
akeles85, line 208: wouldn't it be enough just two? Maybe we can add comments near them, in which it will throw exception according to tested scenario.

akeles85 (dismissed):

Reference_Manual.md
OzanCansel: Fixed.

akeles85 (approved):

akeles85 (approved):

PR #1167, opened 2023-03-20
Sql partition aware routing[API-1862]

akeles85:
This PR includes partition aware routing for SQL.

A LRU cache will be added to the SQL service. This cache will be responsible for holding key argument index of the most recently used SQL statements. If the entry is not cached, the partition id will be queried via the current connection. When the sql query result is received, the entry will be added to the cache and next pages will be queried to the connection represented by the partition id.


hazelcast/include/hazelcast/client/sql/impl/read_optimized_lru_cache.h
```diff
@@ -0,0 +1,233 @@
...
+            int64_t remove_threshold = oldest_timestamps.top();
+            oldest_timestamps.pop();
+
+            // 2nd pass
+            cache_.remove_values_if(
+              [remove_threshold](const value_and_timestamp<V>& v) -> bool {
+                  return (v.timestamp_ <= remove_threshold);
+              });
+
+            cleanup_lock_.release();
```
OzanCansel, line 219: `finally` class in `util` can be used here to execute `release` method at the end of scope.

hazelcast/include/hazelcast/client/sql/impl/read_optimized_lru_cache.h
```diff
@@ -0,0 +1,233 @@
...
+            return lock_.compare_exchange_strong(expected, true);
+        }
+
+        void release() { lock_.store(false); }
+
+        ~custom_atomic_lock() { release(); }
+
+    private:
+        std::atomic<bool> lock_;
+    };
```
OzanCansel, line 151: I don't think this class is a must. Raw `std::atomic<bool>` can be used instead.

hazelcast/src/hazelcast/client/sql.cpp
```diff
@@ -193,6 +252,21 @@ sql_service::handle_execute_response(
...
+        if (is_smart_routing_ && response.partition_argument_index !=
+                                   original_partition_argument_index) {
+            if (response.partition_argument_index != -1) {
+                partition_argument_index_cache_->put(
+                  sql_query,
+                  std::make_shared<int32_t>(response.partition_argument_index));
+                auto temp_shared_ptr = statement_par_arg_index_ptr.lock();
+                if (temp_shared_ptr) {
+                    temp_shared_ptr->store(response.partition_argument_index);
+                }
```
OzanCansel, line 265:
```suggestion
                if (auto argument_index = statement_par_arg_index_ptr.lock()) {
                    argument_index->store(response.partition_argument_index);
                }
```

OzanCansel (changes requested):

hazelcast/include/hazelcast/client/sql/impl/read_optimized_lru_cache.h
akeles85: Fixed

hazelcast/include/hazelcast/client/sql/impl/read_optimized_lru_cache.h
akeles85: Fixed

hazelcast/src/hazelcast/client/sql.cpp
OzanCansel: Can you rename the variable as well because `temp_shared_ptr` doesn't express much ?

akeles85: renamed

OzanCansel (approved):

OzanCansel (approved):

hazelcast/include/hazelcast/client/sql/impl/read_optimized_lru_cache.h
```diff
@@ -0,0 +1,203 @@
...
+protected:
+    /**
+     * Helper class to hold the value with timestamp.
+     */
+    template<typename T>
+    class value_and_timestamp
+    {
+    public:
+        const T value_;
+        int64_t timestamp_;
```
ihsandemir, line 133: i think that this timestamp should be atomic. It is atomically updated at Java.

hazelcast/include/hazelcast/client/sql/impl/read_optimized_lru_cache.h
```diff
@@ -0,0 +1,203 @@
...
+        return (existing_value != nullptr) ? existing_value : default_value;
+    }
+
+    /**
+     * @param key the key of the cache entry
+     * Returns the value to which the specified key is cached,
+     * or {@code null} if this cache contains no mapping for the key.
+     * @returns Returns the value to which the specified key is cached
+     */
+    std::shared_ptr<V> get(const K& key)
```
ihsandemir, line 89: why not return `optional` which makes small value optimizations for you?

hazelcast/src/hazelcast/client/sql.cpp
```diff
@@ -238,6 +314,55 @@ sql_service::query_connection()
...
+        return boost::none;
+    }
+
+    if (arg_index >=
+          static_cast<int32_t>(statement.serialized_parameters_.size()) ||
+        arg_index < 0) {
+        return boost::none;
+    }
+
+    const auto key = statement.serialized_parameters_[arg_index];
```
ihsandemir, line 361: value copy for the whole data binary? why not 
     const auto &key = statement.serialized_parameters_[arg_index];

hazelcast/src/hazelcast/client/sql.cpp
```diff
@@ -58,7 +70,21 @@ sql_service::execute(const sql_statement& statement)
...
+                                        : -1;
+
+    auto arg_index = statement_par_arg_index != -1
+                       ? statement_par_arg_index
+                       : *(partition_argument_index_cache_->get_or_default(
+                           statement.sql(), std::make_shared<int32_t>(-1)));
+
+    auto partition_id = extract_partition_id(statement, arg_index);
+    std::shared_ptr<connection::Connection> query_conn =
+      partition_id != boost::none ? query_connection(partition_id.value())
```
ihsandemir, line 85: partition_id ? query_connection(partition_id.value())

hazelcast/src/hazelcast/client/sql.cpp
akeles85: Fixed. As this PR is merged, the changes are made in another branch. https://github.com/hazelcast/hazelcast-cpp-client/pull/1188

hazelcast/src/hazelcast/client/sql.cpp
akeles85: Fixed. As this PR is merged, the changes are made in another branch. https://github.com/hazelcast/hazelcast-cpp-client/pull/1188

hazelcast/include/hazelcast/client/sql/impl/read_optimized_lru_cache.h
akeles85: I made it to be consistent with SynchronizedMap, so I prefer to leave it as it is.

akeles85: Data is stored in synchronized map, and we just increment the usage count of the shared_ptr with get (which is done via atomic instruction). If we use optional, wouldn't it be expensive, because this time we will copy the data for storing it in optional?

ihsandemir: For this specific case, the value is an (int32 + int64_t) struct (`std::shared_ptr<impl::read_optimized_lru_cache<std::string, int32_t>>
      partition_argument_index_cache_;`) hence, copy would not be a big problem for this specific case. We have no other use of read_optimized_lru_cache structure any other places and have no such concern at the moment. Hence, I thought that it would have been better than using shared_ptr.

PR #1172, opened 2023-03-22
Windows GitHub cache [API-1890]

akeles85:
For windows github actions, boost and thrift directories are cached. They are used from cache after first run. We had 5-6 minutes advantage after doing this.

### DIFF
--- a/.github/workflows/nightly-windows.yml
+++ b/.github/workflows/nightly-windows.yml
@@ -140,7 +140,7 @@ jobs:
           path: C:\Thrift
           key: ${{ runner.os }}-${{ matrix.arch.address_model }}-build-thrift-0.13-${{ matrix.build_type }}
           restore-keys: |
-            ${{ runner.os }}-${{ matrix.arch.address_model }}-build-thrift-0.13-${{ matrix.build_type }}-
+            ${{ runner.os }}-${{ matrix.arch.address_model }}-build-thrift-0.13-${{ matrix.build_type }}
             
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-windows.yml
```diff
@@ -119,7 +131,19 @@ jobs:
           cd ..
           Remove-Item ${{ matrix.vc_boost.boost_folder_name }} -Recurse -Force
 
-      - name: Install Thrift
+      - name: Cache Thrift Version
+        id: cache-thrift
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-thrift-version
```
OzanCansel, line 138: Are these lines required ?

.github/workflows/nightly-windows.yml
```diff
@@ -105,7 +105,19 @@ jobs:
                 "https://curl.se/ca/cacert.pem" `
                 -OutFile "C:\cacert.pem"
 
-      - name: Install Boost
+      - name: Cache Boost Version
+        id: cache-boost
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-boost-version
```
OzanCansel, line 112: Are these lines required ?

.github/workflows/nightly-windows.yml
akeles85: Not a must but it may be used

.github/workflows/nightly-windows.yml
akeles85: Not a must but it may stay

OzanCansel (approved):

PR #1176, opened 2023-03-24
TPC feature is added. [API-1901]

OzanCansel:
- `ClientInvocation` doesn't set `partition_id` if it is already specified with `ClientMessage`.
- Adapter class is not implemented in C++. `partition_id` is set to redirect to specific channel.

This [PR](https://github.com/hazelcast/hazelcast/pull/23774) reflects the `codec` changes.
This [PR](https://github.com/hazelcast/hazelcast/pull/23771) reflects the `HeartbeatManager` changes.


hazelcast/include/hazelcast/client/config/client_tpc_config.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+#pragma once
+
+#include "hazelcast/util/export.h"
+
+namespace hazelcast {
+namespace client {
+namespace config {
+
+class HAZELCAST_API client_tpc_config
```
akeles85, line 24: We may add the below comment.
/**
 * Contains client configurations for TPC.
 * <p>
 * TPC is the next generation Hazelcast that uses thread-per-core model.
 * <p>
 * TPC-aware clients will maintain connections to all cores of all cluster
 * members. The client will route partition-specific invocations to the
 * correct core of the correct member in the best effort basis.
 *
 * @since 5.3
 */

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -1305,11 +1525,48 @@ Connection::close(const std::string& reason, std::exception_ptr cause)
     });
 }
 
+void
+Connection::set_tpc_channels(std::vector<std::unique_ptr<socket>> tpc_channels)
```
akeles85, line 1529: It is called with an rvalue "connection_->set_tpc_channels(move(tpc_channels_));". As it is just used in this place, it might be better to make the parameter as an rvalue to decrease copy.

hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -1367,6 +1367,52 @@ client_config::backup_acks_enabled(bool enabled)
...
 }
 
+config::client_tpc_config&
+client_config::get_tpc_config()
+{
+    return tpc_config_;
+}
+
+config::client_tpc_config::client_tpc_config()
+    :   enabled_ {false}
```
akeles85, line 1377: Can be formetted.

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -40,14 +40,16 @@ class BaseSocket : public hazelcast::client::socket
                const address& addr,
                client::config::socket_options& socket_options,
                boost::asio::io_context& io,
-               std::chrono::milliseconds& connect_timeout_in_millis)
-      : socket_options_(socket_options)
+               std::chrono::milliseconds& connect_timeout_in_millis,
+               connection::Connection& connection)
+      : is_closed_{ true }
```
akeles85, line 45: Should we set is_closed_ to false in "void close() override" method as it is called from other classes.

hazelcast/test/src/tpc_test.cpp
```diff
@@ -0,0 +1,327 @@
...
+                                            seconds);
+    }
+
+    static void ensure_that_there_are_no_tpc_channels(hazelcast_client client,
+                                                      int seconds = 2)
+    {
+        ensure_that_there_are_n_tpc_channels(client, 0, seconds);
+    }
+
+    static void ensure_that_there_are_n_tpc_channels(hazelcast_client client,
```
akeles85, line 69:
```suggestion
    static void ensure_that_there_are_n_tpc_channels_for_each_conn(hazelcast_client client,
```

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -111,37 +118,51 @@ class BaseSocket : public hazelcast::client::socket
         do_read(std::move(connection));
     }
 
+    bool write(std::vector<byte> buffer) override
```
akeles85, line 121:
```suggestion
    bool write(std::vector<byte> &&buffer) override
```

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -1450,18 +1707,26 @@ Connection::get_local_socket_address() const
...
 Connection::last_read_time() const
 {
-    return read_handler.get_last_read_time();
+    return socket_->last_read_time();
 }
 
 void
 Connection::inner_close()
 {
+    auto thisConnection = shared_from_this();
```
akeles85, line 1716:
```suggestion
    auto this_connection = shared_from_this();
```

akeles85 (changes requested):
I have wrote my comments that need some changes.

hazelcast/include/hazelcast/client/config/client_tpc_config.h
OzanCansel: Fixed.

hazelcast/src/hazelcast/client/network.cpp
OzanCansel: I don't want to make it more complex. There will be probably `16` or `32` cores approximately. So trying to prevent cost of one extra `move` operation will probably not affect the performance, probably it will be eliminated in compile time. Also taking parameter by value is far simpler than taking any type of reference. So I generally prefer to stay in safe side unless there is a practical reason not to.

hazelcast/src/hazelcast/client/config.cpp
OzanCansel: Fixed.

hazelcast/test/src/tpc_test.cpp
OzanCansel: Fixed.

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
OzanCansel: As general, usage of pointer, reference or smart pointer mostly cause some obscure bugs. So unless there is a strong reason, I use value semantic. For this case, there is no difference. `buffer` will not be reused. So there is no copy. Btw, this buffer approximately `20bytes` so it can be ignored easily.

akeles85 (approved):

ihsandemir:
It was decided that we will not have this BETA feature for C++ at the moment. Closing this PR.

PR #1177, opened 2023-03-24
Shutdown fix for test cases [API-1956]

akeles85:
We have found that if we trust the destructor of the client for shutdown, it may not do it because of the use count of shared ptr. So user should explicitly call the shutdown method to ensure that client is shutdown. This PR changes the test cases to make them shutdown correctly.

### DIFF
--- a/hazelcast/test/src/remote_controller_client.cpp
+++ b/hazelcast/test/src/remote_controller_client.cpp
@@ -79,7 +79,12 @@ cluster_version()
 
     version.major = std::stoi(major_minor_patch.at(0));
     version.minor = std::stoi(major_minor_patch.at(1));
-    version.patch = std::stoi(major_minor_patch.at(2));
+
+    if (major_minor_patch.size() > 2) {
+      version.patch = std::stoi(major_minor_patch.at(2));
+    } else {
+      version.patch = 0;
+    }
 
     return version;
 }
--- a/hazelcast/test/src/sql_test.cpp
+++ b/hazelcast/test/src/sql_test.cpp
@@ -1553,6 +1553,10 @@ TEST_F(SqlTest, find_with_page_sync_iterator)
 
 TEST_F(SqlTest, timeout_for_page_iterator_sync)
 {
+    // `TABLE` clause is not supported before 5.0.0
+    if (cluster_version() < member::version{ 5, 0, 0 })
+        GTEST_SKIP();
+
     // `generate_stream(1)` generates a row per seconds, so it will guaranteed
     // that it will timeout
     auto result =


hazelcast/test/src/HazelcastTests3.cpp
```diff
@@ -990,6 +1008,9 @@ class BasicClientReplicatedMapNearCacheTest
 
     void create_no_near_cache_context()
     {
+        if (client_ != nullptr) {
+            client_->shutdown().get();
+        }
```
OzanCansel, line 1013: These lines are OK but latest `client_` instance will be left as is. So `destructor` is needed.

hazelcast/test/src/HazelcastTests3.cpp
```diff
@@ -999,6 +1020,9 @@ class BasicClientReplicatedMapNearCacheTest
     {
         client_config nearCachedClientConfig = get_config();
         nearCachedClientConfig.add_near_cache_config(near_cache_config_);
+        if (near_cached_client_ != nullptr) {
+            near_cached_client_->shutdown().get();
+        }
```
OzanCansel, line 1025: These lines are OK but latest `near_cached_client_` instance will be left as is. So destructor is needed.

hazelcast/test/src/compact/compact_test_base.h
```diff
@@ -44,6 +44,10 @@ class compact_test_base : public testing::Test
         factory_.reset(
           new HazelcastServerFactory{ "hazelcast/test/resources/compact.xml" });
         member_.reset(new HazelcastServer{ *factory_ });
+
+        if (client != nullptr) {
+            client->shutdown().get();
+        }
```
OzanCansel, line 50: Latest instance will be left, so it is better to add this to destructor or `TearDown`.

hazelcast/test/src/cp_test.cpp
```diff
@@ -44,6 +44,9 @@ class cp_test : public ClientTest
 
     virtual void SetUp()
     {
+        if (client_) {
+            client_->shutdown().get();
+        }
```
OzanCansel, line 49: Latest instance will be left. This needs to be moved to `TearDown` method.

OzanCansel (changes requested):

hazelcast/test/src/HazelcastTests1.cpp
```diff
@@ -2904,6 +2973,7 @@ ClientTxnTest::ClientTxnTest()
         }
         return boost::make_optional<member>(std::move(members[0]));
     }));
+
```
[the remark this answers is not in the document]
akeles85: Fixed

hazelcast/test/src/HazelcastTests1.cpp
```diff
@@ -3127,6 +3200,10 @@ TEST_F(ClientTxnTest, testTxnInitAndNextMethodRValue)
             }
             return boost::make_optional<member>(std::move(members[0]));
         }));
+
+    if (client_ != nullptr) {
+        client_->shutdown().get();
+    }
```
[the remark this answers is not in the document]
akeles85: Fixed

hazelcast/test/src/HazelcastTests2.cpp
```diff
@@ -1965,6 +1971,9 @@ class serialization_with_server
                                   : "dev");
         config.get_serialization_config().set_byte_order(GetParam());
 
+        if (client_) {
+            client_->shutdown().get();
+        }
```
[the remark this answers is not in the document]
akeles85: Fixed

hazelcast/test/src/compact/compact_test_base.h
akeles85: Fixed

hazelcast/test/src/cp_test.cpp
akeles85: Fixed

OzanCansel (approved):

PR #1194, opened 2023-04-18
Replace NULL keyword to nullptr to modernize as C++ style

cngzhnp:


OzanCansel:
@cngzhnp  Could you remove changes at `RemoteController.cpp` and ``remote_controller_types.cpp` since they are generated code, so we don't modify those classes.

OzanCansel (changes requested):
@cngzhnp Could you remove changes at `RemoteController.cpp` and `remote_controller_types.cpp` since they are generated code, so we don't modify those classes.

OzanCansel:
verify

cngzhnp:
@OzanCansel Thanks for the review, I just reverted back changes on generated source files.

PR #1208, opened 2023-08-14
[Windows] The choco package manager no more supports 32-bit OpenSSL version

ihsandemir:
The choco package manager no more supports 32-bit OpenSSL version. Hence, we have to use OpenSSL 1.1.1.2100 version and force 32-bit usage.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -343,7 +343,7 @@ jobs:
       matrix:
         arch:
           - cmake: Win32
-            choco_options: '--x86'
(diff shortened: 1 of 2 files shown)


ihsandemir:
Verified that this solves the installation of the OpenSSL for windows. We can not test it int he PR itself since it is a github actions workflow change. I verified it at my fork [here](https://github.com/ihsandemir/hazelcast-cpp-client/actions/runs/5858727841/job/15883244344) that the build succeeds, hence I am going to force merge this PR.

PR #1216, opened 2023-12-15
Do not use the deprecated `repoUrl` mvn option when downloading the test jars [API-2183]

ihsandemir:
Do not use the deprecated `repoUrl` mvn option but use the `remoteRepositories` option instead.

Also adde `mvn depedency:copy`


emreyigit (commented):
Approved.

PR #1217, opened 2023-12-25
Update the documentation for the reliable_topic listener callback on_cancel [API-2183]

ihsandemir:
Updated the documentation for the reliable_topic listener callback function usages including the new callback on_cancel.

### DIFF
--- a/Reference_Manual.md
+++ b/Reference_Manual.md
@@ -1910,15 +1910,26 @@ A reliable_topic usage example is shown below.
 hazelcast::client::topic::reliable_listener make_listener(std::atomic<int> &n_received_messages, int64_t sequence_id = -1) {
   using namespace hazelcast::client::topic;
   
-  return reliable_listener(false, sequence_id).on_received([&n_received_messages](message &&message){
-      ++n_received_messages;
-      auto object = message.get_message_object().get<std::string>();
-      if (object) {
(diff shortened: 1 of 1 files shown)


Reference_Manual.md
```diff
@@ -1972,6 +1983,7 @@ void listen_with_config() {
   client.shutdown().get();
 }
 ```
+The above examples composes a reliable topic listener which listens to new messages and increment received message count atomically. The listener also prints a debug message for each callback function available in the listener. `on_store_sequence_id` callback function is used to store the last received message sequence id. `terminate_on_exception` gives the user a chance to evaluate the exception and continue or terminate the listener. `on_cancel` callback function is a way to notify the user before the listener is cancelled for any reason so that user can take necessary action. The first example uses a default config for the reliable topic and the second one uses a custom configuration.
```
Serdaro, line 1986:
```suggestion
The above examples compose a reliable topic listener that listens to the new messages and increments the received message count atomically. The listener also prints a debug message for each callback function available in the listener. The `on_store_sequence_id` callback function is used to store the last received message's sequence ID. The `terminate_on_exception` function allows you to evaluate the exception and continue or terminate the listener. The `on_cancel` callback function notifies you before the listener is cancelled for any reason so that you can take the necessary action. The first example uses a default configuration for the reliable topic, and the second one uses a custom configuration.
```

Serdaro (approved):

Issue #1232, opened 2024-08-09
Make Dependabot flag outdated GitHub actions


JackPGreen:
any chance to upgrade this as well?

_Originally posted by @nishaatr in https://github.com/hazelcast/hazelcast-cpp-client/pull/1231#discussion_r1711193094_

PR #1234, opened 2024-08-22
Enable Dependabot

JackPGreen:
Fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/1232

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
ihsandemir, line 3: does this mean dependabot will only be triggered if change in github-actions?

JackPGreen: Yes, it will only raise PRs if there are updateable GitHub Actions, checked weekly.

ihsandemir: What do you mean by `updateable GitHub Actions`? I thought that dependabot would check the code base on periodic basis regardless of github actions.

JackPGreen: > What do you mean by `updateable GitHub Actions`? I thought that dependabot would check the code base on periodic basis regardless of github actions.

It means the GitHub Actions dependencies declared in the codebase - e.g.:
https://github.com/hazelcast/hazelcast-cpp-client/blob/b2d7e9424dddb6f6cd55512167ba29eb2fe905cc/.github/workflows/nightly-windows.yml#L78

ihsandemir (approved):

PR #1248, opened 2024-11-12
Remove `test-membership` duplication

JackPGreen:
The PR builder uses a local copy of an action which has [since been centralised](https://github.com/hazelcast/hazelcast-tpm/blob/main/membership/action.yaml).

This was originally copied from the pattern in `hazelcast-nodejs-client` - but this has [since been migrated](https://github.com/hazelcast/hazelcast-nodejs-client/blob/master/.github/workflows/coverage_runner.yml#L16-L29).

Updated to use centralise version - allowing a central place to make improvements.


.github/workflows/build-pr.yml
```diff
@@ -39,21 +39,12 @@ jobs:
     name: Test Hazelcast Membership
     runs-on: ubuntu-latest
     outputs:
-      is-hazelcast: ${{ steps.test-membership.outputs.is-member }}
+      is-hazelcast: ${{ steps.test-membership.outputs.check-result }}
 
     steps:
-      # checkout the hazelcast/hazelcast-cpp-client repository
-      # bare minimum - just to use actions
-      - name: Checkout code
```
ihsandemir, line 47: why deleted `Checkout code` ?

JackPGreen: > why deleted `Checkout code` ?

Only for the `test-membership` step - it does a `checkout` to get a copy of the `test-membership` action.

By referencing the (remote) `hazelcast/hazelcast-tpm/membership` action, that's done implicitly.

nishaatr (approved):

ihsandemir (approved):

JackPGreen:
@ihsandemir it looks like the PR builders gone back to being flakey - but I don't think this change is related. Would you be able to force merge, please...?

ihsandemir:
@JackPGreen We need to disable some tests. I submitted a [PR](https://github.com/hazelcast/hazelcast-cpp-client/pull/1251) for it. We need to disable these tests since we shutdown the test Aws instance permanently now due to costs. These tests need to change. Opened an [issue](https://github.com/hazelcast/hazelcast-cpp-client/issues/1250) for future reference.

PR #1281, opened 2025-05-17
Centralise duplicated matrix configurations

JackPGreen:
GitHub Actions use matrixes' to execute builds with variations - e.g. different compiler configurations.
These configurations are duplicated for each platform - and instead should be centralised.


JackPGreen:
[Re-raising as a non-fork to facilitate testing](https://github.com/hazelcast/hazelcast-cpp-client/pull/1284).

PR #1290, opened 2025-05-23
Increment project version to `5.4`

JackPGreen:
The existence of a `5.3` branch implies `master` is `5.4`.

Updated version and externalised/centralised.


CMakeLists.txt
```diff
@@ -16,8 +16,12 @@
 
 cmake_minimum_required(VERSION 3.10)
 
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/.github/config.env" ENV_FILE_CONTENTS)
+string(REGEX REPLACE ".*PROJECT_VERSION=([^\n\r]*).*" "\\1" PROJECT_VERSION "${ENV_FILE_CONTENTS}")
+message(STATUS "PROJECT_VERSION = ${PROJECT_VERSION}")
```
ihsandemir, line 21: I did not like this reading parsing approach. It is error prone. Our users use it from source code, hence it is important that they can compile locally and in most possible environments. Hence, I do not favor making the version yet another github environment variable. Some of our users may not use github but download source zip and build as well. 

We may have hazelcast-cpp-client-examples cmake use it from parent cmake or a common `CMakelists.in` which can be included in both files.

Another alternative would be passing the version from cmake  build command but I would rather not leave the version that flexible and leave it to the users and that is yet an extra parameter.

PR #1295, opened 2025-05-28
Remove unused docker images

JackPGreen:
The images are not being used, and a duplicate exists in https://github.com/hazelcast/client-compatibility-suites/blob/master/KubernetesInternalClients/cpp/Dockerfile which _is_ being used.

Moreover, hazelcast-fedora22-x86_64.dockerfile is using an outdated version of Fedora that is no longer able to download boost which means it _cannot_ work, which was already fixed in the compatibility suites in https://github.com/hazelcast/client-compatibility-suites/pull/164

The GitHub actions are using Ubuntu images and there are no Jenkins C++ builds that I can find.


ihsandemir (approved):
OK, they were used in deleted Jenkins jobs.

PR #1297, opened 2025-05-30
Update usage of deprecated `windows-2019` GitHub runner

JackPGreen:
[`windows-2019` is deprecated](https://github.com/actions/runner-images/issues/12045) and will soon be removed.

Changes:
- increment all Windows builds to the next Windows runner
- remove Boost `1.73` test [as doesn't work on Windows 2022](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/15351436507/job/43200388192)
     - `"Unknown toolset: vcunk"`
- improve logging on failure

[Example execution](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/15351519557) - flakey tests but all passed on 2nd run.

### DIFF
--- a/.github/workflows/nightly-windows.yml
+++ b/.github/workflows/nightly-windows.yml
@@ -16,12 +16,6 @@ jobs:
       fail-fast: false
       matrix:
         vc_boost:
-          - name: msvc-2022_boost_1730
-            image: 'windows-2022'
-            boost_url: 'https://archives.boost.io/release/1.73.0/source/boost_1_73_0.tar.gz'
-            boost_archive_name: 'boost_1_73_0.tar.gz'
-            boost_folder_name: 'boost_1_73_0'
-            boost_include_folder: 'C:\Boost\include\boost-1_73'
           - name: msvc-2022_boost_1800
             image: 'windows-2022'
             boost_url: 'https://archives.boost.io/release/1.80.0/source/boost_1_80_0.tar.gz'


.github/workflows/nightly-windows.yml
```diff
@@ -16,26 +16,20 @@ jobs:
...
         vc_boost:
-          - name: msvc-2019_boost_1730
-            image: 'windows-2019'
-            boost_url: 'https://archives.boost.io/release/1.73.0/source/boost_1_73_0.tar.gz'
-            boost_archive_name: 'boost_1_73_0.tar.gz'
-            boost_folder_name: 'boost_1_73_0'
-            boost_include_folder: 'C:\Boost\include\boost-1_73'
-          - name: msvc-2019_boost_1800
-            image: 'windows-2019'
+          - name: msvc-2022_boost_1800
```
ihsandemir, line 19: Is boost 1.78 the minimum version of boost supported on at msvc-2025?

We need a decision on minimum supported boost version. We need to re-evaluate our original min version support decision. We also need to evaluate the min language level to support, C++ 14? Based on these decisions, yes, we should et up to data the versions.

ihsandemir (changes requested):
We need to decide on the min boost version support.

.github/workflows/nightly-windows.yml
ihsandemir: i create the linked issue to study this and update the min supported windows and VC compiler versions properly. We need to be sure we do not break any existing users.

PR #1313, opened 2025-06-24
Remove top-level `env` from GitHub composite actions

JackPGreen:
tl;dr - it doesn't work

In composite actions, a `shell` _must_ be specified. To avoid duplication, this was centralised to an `env` variable.

However, GitHub [doesn't support this](https://github.com/orgs/community/discussions/51280) - for example, in [this job](https://github.com/hazelcast/hazelcast-go-client/actions/runs/15848149662/job/44674983175) the output of `${{ env.shell }}` is blank.

It _appeared_ to work because if the specified `shell` is empty, it uses the default shell of the executing runner - and all of our executions were setting `bash` on Linux/Mac or `pwsh` on Windows anyway so the net effect is the same.

This bug appeared when a `bash` composite action failed to resolve `bash` commands on a Windows runner - because it was actually running in an (implicit) `pwsh` shell.

### DIFF
--- a/.github/actions/build-test/macos-x86_64/action.yml
+++ b/.github/actions/build-test/macos-x86_64/action.yml
@@ -22,20 +22,15 @@ inputs:
   HZ_TEST_AWS_INSTANCE_PRIVATE_IP:
     required: true
 
-env:
-  # Not possible to set this as a default
-  # https://github.com/orgs/community/discussions/46670
-  shell: bash
-
 runs:
   using: composite
   steps:
     - name: Read Config
-      shell: ${{ env.shell }}
+      shell: bash
       run: cat .github/config.env >> $GITHUB_ENV
 
     - name: Install Dependencies
-      shell: ${{ env.shell }}
+      shell: bash
       run: |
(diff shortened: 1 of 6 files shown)


nishaatr (approved):
GH should really fix this. specifying  shell for every step is cumbersome

Just mentioning below alternatives as FYI (credit to Google). 

1. set shell as [input](https://github.com/hazelcast/hazelcast-mono/pulls?q=is%3Apr+milestone%3A5.4.z+is%3Aopen) with `bash` as default. For Windows can be hardcoded. But this ADR is in accepted state so not sure it works
2. Use job [defaults](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun). Not sure this works or not for composite
3. change the default shell directly in the runner itself but this is likely going to be problematic
4. specify `shebang` instead but this is no different to hardcoding. Also, default shell will be first called then `bash` so pointless

This [table](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrunshell) shows what `shell` is used based on platform

PR #1315, opened 2025-06-30
Update `apache.org` download links to use `closer.lua`

JackPGreen:
[We have seen instability downloading from `archive.apache.org`](https://github.com/hazelcast/client-compatibility-suites/actions/runs/15973338298/job/45050422979).

Update links to use `closer.lua` instead, as mentioned in [Apache's documentation](https://infra.apache.org/release-download-pages.html).

### DIFF
--- a/.github/actions/build-test/windows/action.yml
+++ b/.github/actions/build-test/windows/action.yml
@@ -100,7 +100,7 @@ runs:
       shell: pwsh
       run: |
         Invoke-WebRequest `
-            "https://archive.apache.org/dist/thrift/${{ inputs.THRIFT_VERSION }}/thrift-${{ inputs.THRIFT_VERSION }}.tar.gz" `
+            "https://www.apache.org/dyn/closer.lua/thrift/${{ inputs.THRIFT_VERSION }}/thrift-${{ inputs.THRIFT_VERSION }}.tar.gz?action=download" `
             -OutFile "thrift-${{ inputs.THRIFT_VERSION }}.tar.gz"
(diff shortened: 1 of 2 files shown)


.github/actions/build-test/windows/action.yml
```diff
@@ -100,7 +100,7 @@ runs:
       shell: pwsh
       run: |
         Invoke-WebRequest `
-            "https://archive.apache.org/dist/thrift/${{ inputs.THRIFT_VERSION }}/thrift-${{ inputs.THRIFT_VERSION }}.tar.gz" `
+            "https://www.apache.org/dyn/closer.lua/thrift/${{ inputs.THRIFT_VERSION }}/thrift-${{ inputs.THRIFT_VERSION }}.tar.gz?action=download" `
```
ihsandemir, line 103: For thrift installation why don't we put a script like `install-thrift.bat` similar to `install-thrift.sh` and simplify this step as we do for linux/macos?

JackPGreen: I don't think having it externalised helps as it's not very long and it's not duplicated - but it's a fair point that it's done inconsistently between platforms.

Will address seperately

https://github.com/hazelcast/hazelcast-cpp-client/issues/1316

nishaatr: Also, may be time to upgrade Thrift?
note: `install-thrift.sh` is using `https://archive.apache.org`

ihsandemir (approved):

Issue #1317, opened 2025-07-01
Upgrade Thrift version? [API-2344]


ihsandemir:
Should we upgrade Thrift version to use latest one?
note: `install-thrift.sh` is using `https://archive.apache.org`

_Originally posted by @nishaatr in https://github.com/hazelcast/hazelcast-cpp-client/pull/1315#discussion_r2176860581_

PR #1318, opened 2025-07-02
Fix `install-thrift` download errors on GitHub Runners

JackPGreen:
Despite the changes in https://github.com/hazelcast/hazelcast-cpp-client/pull/1315, the downloads are [still failing](https://github.com/hazelcast/client-compatibility-suites/actions/runs/16020766406/job/45197809787) on GitHub Runners.

Specifically, the [mirror service doesn't _have_ a mirror of the specific version of `thrift` we want](https://www.apache.org/dyn/closer.cgi?path=/thrift/0.13.0/thrift-0.13.0.tar.gz):
> The requested file or directory is not on the mirrors.
The object is in our archive : https://archive.apache.org/dist/thrift/0.13.0/thrift-0.13.0.tar.gz

And it just redirects to `archive.apache.org` - which is what it was using originally.

Updated to use an alternative mirror (Debian) and checked the checksums match (`38a27d391a2b03214b444cb13d5664f1`).

### DIFF
--- a/.github/actions/build-test/windows/action.yml
+++ b/.github/actions/build-test/windows/action.yml
@@ -99,8 +99,11 @@ runs:
       name: Install Thrift
       shell: pwsh
       run: |
+        # Ideally we would download directly from Apache:
+        # https://www.apache.org/dyn/closer.lua/thrift/$1/thrift-$1.tar.gz?action=download
+        # But this has intermittent failures in GitHub Actions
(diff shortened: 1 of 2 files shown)


nishaatr (approved):
we have https://github.com/hazelcast/hazelcast-cpp-client/issues/1317 so might need to revisit this

PR #1319, opened 2025-07-02
DOC-527: remove duplicate content, add pointers to docs site instead

amandalindsay:
Removed the duplicate getting started & install content from the C++ Reference manual and Readme which has been copied to the main docs site, and also added a pointer to the Docs site for this information.


amandalindsay:
Reopened

PR #1322, opened 2025-07-16
Fixed build problems with latest Boost library versions [DEX-431]

### LINKED ISSUE #1253: support boost 1.87
hi, asio removed some previously deprecated facilities, see https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/history.html

I found this when I update vcpkg's boost to 1.87, if you want update you can use my patch, https://github.com/microsoft/vcpkg/pull/42678/files#diff-8c050938d3e295d388fcb05fd405b5edf99a1935e1d13390e0c39c3962eddc37

ihsandemir:
Fixed build problems with latest Boost library versions.
Updated the project to use the minimum C++ standard to be the same as the one required by the Boost library. This means if a user is using an older Boost version we use C++11 but if newer Boost version is used we change level to C++ 14 or C++ 17. the algorithm is like this:
```
    # Check Boost version and set C++ standard accordingly
    # Our library can not use a lower C++ standard than the one required by Boost.
    if (Boost_VERSION VERSION_LESS 1.70)
        set(CMAKE_CXX_STANDARD 11)
    elseif (Boost_VERSION VERSION_LESS 1.77)
        set(CMAKE_CXX_STANDARD 14)
    else()
        set(CMAKE_CXX_STANDARD 17)  # Boost.JSON, Boost.Process, etc.
    endif()
```

fixes #1253 
fixes https://hazelcast.atlassian.net/browse/DEX-431

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,8 +21,6 @@ project(hazelcast-cpp-client
         DESCRIPTION "Hazelcast C++ Client"
         LANGUAGES CXX)
 
-# Set the C++ standard
-set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # Generate the Reference_Manual.md file using the latest project version.
@@ -89,15 +87,27 @@ find_package(Threads REQUIRED)
 
 # find Boost
 if (WIN32)
-	find_package(Boost 1.73 REQUIRED COMPONENTS thread chrono)
+	find_package(Boost 1.73 CONFIG REQUIRED COMPONENTS thread chrono)
 else()
-	find_package(Boost 1.71 REQUIRED COMPONENTS thread chrono)
+	find_package(Boost 1.71 CONFIG REQUIRED COMPONENTS thread chrono)
 endif()
 
 if (Boost_FOUND)
     message(STATUS "Boost version: ${Boost_VERSION}")
     message(STATUS "Boost include directory: ${Boost_INCLUDE_DIRS}")
     message(STATUS "Boost libraries: ${Boost_LIBRARIES}")
+
+    # Check Boost version and set C++ standard accordingly
+    # Our library can not use a lower C++ standard than the one required by Boost.
+    if (Boost_VERSION VERSION_LESS 1.70)
+        set(CMAKE_CXX_STANDARD 11)
+    elseif (Boost_VERSION VERSION_LESS 1.77)
+        set(CMAKE_CXX_STANDARD 14)
+    else()
+        set(CMAKE_CXX_STANDARD 17)  # Boost.JSON, Boost.Process, etc.
+    endif()
+
+    message(STATUS "Required C++ standard: ${CMAKE_CXX_STANDARD}")
 else()
     message(FATAL_ERROR "Boost not found!")
 endif()
@@ -263,8 +273,12 @@ install(
 )
 
 if (BUILD_TESTS)
+    message(STATUS "BUILD_TESTS is enabled. Ensure the 'build-tests' feature is enabled in vcpkg.json.")
+    find_package(GTest REQUIRED)
     add_subdirectory(hazelcast/test)
-endif ()
+else()
+    message(STATUS "BUILD_TESTS is disabled.")
+endif()
 
 if (BUILD_EXAMPLES)
     add_subdirectory(examples)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,6 +94,14 @@ else()
 	find_package(Boost 1.71 REQUIRED COMPONENTS thread chrono)
 endif()
 
+if (Boost_FOUND)
+    message(STATUS "Boost version: ${Boost_VERSION}")
+    message(STATUS "Boost include directory: ${Boost_INCLUDE_DIRS}")
+    message(STATUS "Boost libraries: ${Boost_LIBRARIES}")
+else()
+    message(FATAL_ERROR "Boost not found!")
+endif()
+
 # find OpenSSL if building WITH_OPENSSL
 if (WITH_OPENSSL)
     if (APPLE)
@@ -104,6 +112,14 @@ if (WITH_OPENSSL)
     endif ()
 
     find_package(OpenSSL REQUIRED)
+
+    if (OpenSSL_FOUND)
+        message(STATUS "OpenSSL version: ${OpenSSL_VERSION}")
+        message(STATUS "OpenSSL include directory: ${OPENSSL_INCLUDE_DIR}")
+        message(STATUS "OpenSSL libraries: ${OPENSSL_LIBRARIES}")
+    else()
+        message(FATAL_ERROR "OpenSSL not found!")
+    endif()
 endif ()
 
 # add the library target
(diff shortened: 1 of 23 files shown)


Reference_Manual.md
JackPGreen, line 1: Isnt this a file that shouldn't be edited?

hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -2360,11 +2360,11 @@ TEST(ClientMessageTest, test_encode_sql_query_id)
 
     protocol::ClientMessage msg;
 
-    boost::uuids::uuid server_uuid{ 1, 2,  3,  4,  5,  6,  7,  8,
-                                    9, 10, 11, 12, 13, 14, 15, 16 };
-    boost::uuids::uuid client_uuid{ 21, 22, 23, 24, 25, 26, 27, 28,
-                                    29, 30, 31, 32, 33, 34, 35, 36 };
-    msg.set(sql::impl::query_id{ server_uuid, client_uuid });
+boost::uuids::uuid server_uuid = boost::uuids::uuid{{ 1, 2,  3,  4,  5,  6,  7,  8,
```
JackPGreen, line 2363: Formatting looks a bit wierd.

vcpkg.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json",
+  "name": "hazelcast-cpp-client",
+  "version": "5.5.0",
+  "dependencies": [
```
JackPGreen, line 5: Same comments from here - https://github.com/hazelcast/hazelcast-cpp-client/pull/1320/files#r2204924261 - applies.

hazelcast/test/src/HazelcastTests8.cpp
ihsandemir: fixed at df7ff8f2dc2a8a62a1ffa7a3c8fb71ae1f9d452f

vcpkg.json
ihsandemir: fixed at 9fdfb56c65586350d6dc8e22354efcc96b141ce7

Reference_Manual.md
ihsandemir: It is auto generated during built but we still commit it after generation.

JackPGreen (approved):

ihsandemir:
> @ihsandemir Could you add some reasoning about why language level is set to C++14 for Boost `[ 1.70, 1.77)` ? According to release notes for Boost 1.82, it is compatible with C++11: https://www.boost.org/releases/1.82.0/
> 
> > The new baseline requirement for the above libraries, and for most of Boost, will be C++11.

I will re-look at these ranges. I believe the AI replied these ranges since some modules started to support later C++ support with those releases. I also observed our PR builder failed with Boost 1.78.0 wince it tried to use C++ 17 as the compiler. I will double check and fix these versions.

.github/actions/build-test/windows/action.yml
```diff
@@ -91,7 +91,7 @@ runs:
         rm ${{ inputs.BOOST_ARCHIVE_NAME }}
         cd ${{ inputs.BOOST_FOLDER_NAME }}
         .\bootstrap.bat
-        .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-atomic --with-thread --with-chrono install
+        .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-atomic --with-uuid --with-thread --with-chrono install
```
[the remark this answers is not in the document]
ihsandemir: Yes, I agree. I also tried to find one online, could not find except [this](index) page with only one option used. When you have b2, you can list options by `b2 --help`, that is probably the best way for now.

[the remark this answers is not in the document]
JackPGreen: Ok, I've raised https://github.com/hazelcast/hazelcast-cpp-client/pull/1332

_if_ you approve, we can merge this on top of that and then add the new library. But at least I understand it's not an option, it's adding [a new library](https://www.boost.org/libs/uuid/).

CMakeLists.txt
```diff
@@ -104,6 +118,14 @@ if (WITH_OPENSSL)
...
 
     find_package(OpenSSL REQUIRED)
+
+    if (OpenSSL_FOUND)
+        message(STATUS "OpenSSL version: ${OpenSSL_VERSION}")
+        message(STATUS "OpenSSL include directory: ${OPENSSL_INCLUDE_DIR}")
+        message(STATUS "OpenSSL libraries: ${OPENSSL_LIBRARIES}")
+    else()
+        message(FATAL_ERROR "OpenSSL not found!")
+    endif()
```
JackPGreen: Based on the [docs for `find_package`](https://cmake.org/cmake/help/latest/command/find_package.html), shouldn't it already fail due to `REQUIRED` rendering the `else` pointless?

JackPGreen (changes requested):
I'm a bit confused about the Boost versions in use:
- `vcpkg.json` - `1.88`
- PR builder - `1.89`
- Coverage report - `1.78`
- Nightly builds
   - `1.71` <-- this is an unsupported version I think?
   - `1.73`
   - `1.76`
   - `1.78`
   - `1.80` 

I think that we should either:
- test with every version 
_or_
- test with at most two versions - newest and oldest

.github/actions/build-test/windows/action.yml
```diff
@@ -94,7 +94,7 @@ runs:
         # Build the libs for:
         # - Atomic - https://www.boost.org/libs/atomic/
         # - Thread - https://www.boost.org/libs/thread/
-        # - Chrono - https://www.boost.org/libs/chrono/
+        # - Chrono - https://www.boost.org/libs/chrono/        
         .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-atomic --with-thread --with-chrono install
```
JackPGreen: Don't we need `--with-uuid` (+ docs)? Guessing got lost in the merge.

Reference_Manual.md
```diff
@@ -199,15 +199,15 @@ Please see [getting started](https://github.com/microsoft/vcpkg#getting-started)
 If you use Linux or Mac:
 
 ```sh
-git clone https://github.com/microsoft/vcpkg --branch 2025.02.14
+git clone https://github.com/microsoft/vcpkg
```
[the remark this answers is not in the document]
ihsandemir: fixed at 2a38a0140a524bf258fefb5034ab7b356f75f971

ihsandemir:
> > > I'm a bit confused about the Boost versions in use:
> 
> > vcpkg: We can not control their version.
> 
> But we are, in `vcpkg.json`? My understand was that file was a hint to a suggested version for use in an IDE?
> 
> > After I merge this PR, then I will update nightlies and coverage {...} will be separate PR.
> 
> Why not do it at the same time?

I mean vcpkg current latest is 1.88.0. They did not update there yet to 1.89.0. Does this make sense?

ihsandemir:
> > > I'm a bit confused about the Boost versions in use:
> 
> > vcpkg: We can not control their version.
> 
> But we are, in `vcpkg.json`? My understand was that file was a hint to a suggested version for use in an IDE?
> 
> > After I merge this PR, then I will update nightlies and coverage {...} will be separate PR.
> 
> Why not do it at the same time?

Ok, changed out nightlies at 66a1b0d083c600f50f5b00edfd6d3a2e2b7df8c7

JackPGreen (approved):

PR #1333, opened 2025-10-15
Remove `atomic` library from Windows tests

JackPGreen:
The Linux tests don't use it, consistency.

### DIFF
--- a/.github/actions/build-test/windows/action.yml
+++ b/.github/actions/build-test/windows/action.yml
@@ -91,6 +91,9 @@ runs:
         rm ${{ inputs.BOOST_ARCHIVE_NAME }}
         cd ${{ inputs.BOOST_FOLDER_NAME }}
         .\bootstrap.bat
+        # Build the libs for:
+        # - Thread - https://www.boost.org/libs/thread/
+        # - Chrono - https://www.boost.org/libs/chrono/
         .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-thread --with-chrono install
         cd ..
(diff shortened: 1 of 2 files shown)


.github/actions/build-test/windows/action.yml
```diff
@@ -92,10 +92,9 @@ runs:
         cd ${{ inputs.BOOST_FOLDER_NAME }}
         .\bootstrap.bat
         # Build the libs for:
-        # - Atomic - https://www.boost.org/libs/atomic/
         # - Thread - https://www.boost.org/libs/thread/
         # - Chrono - https://www.boost.org/libs/chrono/
-        .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-atomic --with-thread --with-chrono install
+        .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-thread --with-chrono install
```
ihsandemir: According to [this](https://www.boost.org/doc/user-guide/header-organization-compilation.html) page, atomic is one of non-header libs, hence needs build. 

Q: Why linux one work? [Boost.Thread depends on Boost.Atomic](https://www.boost.org/doc/libs/1_89_0/doc/html/thread/build.html#thread.build.configuration) hence, atomic is built as well. 

Well, we can omit or not, the result is the same. less code the better. In that sense we can even remove, Chrono since Thread depends on Chrono :)

JackPGreen: > In that sense we can even remove, Chrono since Thread depends on Chrono :)

https://github.com/hazelcast/hazelcast-cpp-client/pull/1335

PR #1334, opened 2025-10-15
Fix build problems with latest Boost library versions  [DEX-431] [CTT-734][CTT-716][CTT-726]

ihsandemir:
Fixed build problems with latest Boost library versions.

related to https://github.com/hazelcast/hazelcast-cpp-client/pull/1322

Had to bump up google test version due to this problem at Mac os build:
```
CMake Error at build/hazelcast/test/googletest-src/CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

Problem with MacOs builds is reported at [this](https://github.com/boostorg/thread/issues/402) issue at Boost: 
```
In file included from /Users/runner/work/hazelcast-cpp-client/hazelcast-cpp-client/hazelcast/include/hazelcast/client/serialization/pimpl/compact/compact.h:22:
/usr/local/include/boost/thread/future.hpp:4672:19: error: no member named 'that' in 'run_it<FutureExecutorContinuationSharedState>'; did you mean 'that_'?
 4672 |           that_=x.that;
      |                   ^~~~
      |                   that_
/usr/local/include/boost/thread/future.hpp:4650:55: note: 'that_' declared here
 4650 |     shared_ptr<FutureExecutorContinuationSharedState> that_;
      |                                                       ^
```
This problem is only fixed at boost 1.88.0 and later. See [this](https://github.com/boostorg/thread/commit/49ccf9c30a0ca556873dbf64b12b0d741d1b3e66) commit.


hazelcast/test/src/ClientTest.cpp
JackPGreen: Is this related to boost, or a port of https://github.com/hazelcast/hazelcast/pull/16375 ?

ihsandemir: It is test enhancement which can be applied without the port PR you mentioned.

hazelcast/src/hazelcast/client/protocol.cpp
```diff
@@ -376,7 +376,8 @@ operator<<(std::ostream& os, const ClientMessage& msg)
 void
 ClientMessage::set(unsigned char* /* memory */, boost::uuids::uuid uuid)
 {
-    std::memcpy(wr_ptr(uuid.size()), uuid.data, uuid.size());
+    assert(uuid.size() == util::Bits::UUID_SIZE_IN_BYTES);
```
yuce: `static_assert` may be better here.

yuce (changes requested):
Here are some ideas...

hazelcast/include/hazelcast/client/protocol/ClientMessage.h
```diff
@@ -758,7 +764,7 @@ class HAZELCAST_API ClientMessage
     {
         if (get<bool>()) {
             // skip the next 16 bytes
-            rd_ptr(sizeof(boost::uuids::uuid));
+            rd_ptr(util::Bits::UUID_SIZE_IN_BYTES);
```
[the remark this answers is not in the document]
ihsandemir: I removed intentionally all usages of sizeof(uuid) in the code base and replaced with this static defined constant, in case they change uuid impl which may result in a different number than 16. In terms of the number of bytes the uuid is represented, it has to be 16 by definition of uuid, hence, it will never change, hence, our usage should be fine.

I have the assert check at [this](https://github.com/hazelcast/hazelcast-cpp-client/blob/5f8570bf3786f4934ff85408dfd0312587f205d5/hazelcast/src/hazelcast/client/protocol.cpp#L379) line, is that OK?

ihsandemir:
Closing since I can not split the original PR. MacOS build does not work.

PR #1336, opened 2025-10-20
Updated the codecs used with latest changes [CTT-713]

ihsandemir:
Updated the codecs used with latest changes. Also, updated the code to use these codecs properly with the new introduced parameters.

Associated protocol PR is https://github.com/hazelcast/hazelcast-client-protocol/pull/557

### DIFF
--- a/.github/actions/build-test/windows/action.yml
+++ b/.github/actions/build-test/windows/action.yml
@@ -93,8 +93,7 @@ runs:
         .\bootstrap.bat
         # Build the libs for:
         # - Thread - https://www.boost.org/libs/thread/
-        # - Chrono - https://www.boost.org/libs/chrono/
-        .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-thread --with-chrono install
+        .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} --with-thread install
         cd ..
         Remove-Item ${{ inputs.BOOST_FOLDER_NAME }} -Recurse -Force
 
--- a/.github/workflows/nightly-macos-x86_64.yml
+++ b/.github/workflows/nightly-macos-x86_64.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         boost:
-            - version: 1.73.0
+            - version: 1.88.0 # MacOS issue (https://github.com/boostorg/thread/issues/402) fixed at 1.88.0
             - version: 1.89.0
         build_type: ${{ fromJSON(needs.shared-matrix.outputs.build-type) }}
         shared_libs: ${{ fromJSON(needs.shared-matrix.outputs.shared-libs) }}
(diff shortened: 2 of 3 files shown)


hazelcast/include/hazelcast/client/internal/version.h
```diff
@@ -0,0 +1,37 @@
...
+ * @since 5.5
+ */
+struct version
+{
+    byte major;
+    byte minor;
+};
+} // namespace impl
+} // namespace client
+} // namespace hazelcast
```
JackPGreen:
```suggestion
} // namespace hazelcast

```

JackPGreen (commented):
> I generate and format the code. Is the problem about formatting(clang format) {...}?

I think this makes it very challenging to review - if it's generated code, as long as I'm happy with the generator configuration (i.e. https://github.com/hazelcast/hazelcast-client-protocol/pull/557) and I can validate the code is generated as it should be, I can simply rubber-stamp it.

In this scenario, I can't validate the code is generated as it should be, so instead I need to review the generator configuration _and_ review the generated code as well.

I tried to update the `client-protocol` template formatting in https://github.com/JackPGreen/hazelcast-client-protocol/tree/clang-formatting to see if I could produce the output heree "out-of-the-box" but it's not (easily) possible as the formatter is line breaking in different positions due to the line length (i.e. a parameter may be declared on the same line as a function _or_ a newline depending on the name of the method - which isn't easy to implement in a template).

I think it would be better simply to leave the formatting of the generated code as-is.

> do you see other differences?

I don't _think_ so.

JackPGreen:
> We can integrate `clang-format` into the generator. the commands is similar to `clang-format -i --verbose hazelcast/generated-sources/src/hazelcast/client/protocol/codec/codecs.* `

🤦

Good point - it's much simpler to _actually use_ the formatter than to try to approximate it.

hazelcast/generated-sources/src/hazelcast/client/protocol/codec/codecs.h
```diff
@@ -83,13 +103,21 @@ struct HAZELCAST_API client_addclusterviewlistener_handler
...
+     * based on RoutingStrategy.
+     */
+    virtual void handle_membergroupsview(
+      int32_t version,
+      std::vector<std::vector<boost::uuids::uuid>> const& member_groups) = 0;
+
+    /**
+     * @param version The cluster version.
+     */
+    virtual void handle_clusterversion(internal::version const& version) = 0;
```
JackPGreen:
```suggestion
    virtual void handle_clusterversion(version const& version) = 0;
```

When I regenerated the `client-protocol`, I got a (_very slightly_) different output.

I'm not _really_ saying we should change it. Just trying to understand if it's outdated, or our tooling differs.

ihsandemir: I think you did not update the client protocol repo to latest. See https://github.com/hazelcast/hazelcast-client-protocol/blob/41a96a9bb2798bfff174c52140781d7eb36d2bae/cpp/__init__.py#L127 line.

ihsandemir: Ok, there was a duplication which I fixed now at [this](https://github.com/hazelcast/hazelcast-client-protocol/pull/558) PR. Should be fine after this PR.

PR #1338, opened 2025-10-22
Added the sanitizer to the verify script [HZ-5191]

ihsandemir:
Added the sanitizer to the verify script. The nightly tests were failing with:

```
Undefined symbols for architecture arm64:
  "___asan_after_dynamic_init", referenced from:
```

Green build [here](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/18887761793).

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -210,7 +210,6 @@ jobs:
     name: windows-${{ matrix.options.address_model }}-(${{ matrix.build_type }}, ${{ matrix.shared_libs.name }}, ${{ matrix.with_openssl.name }})
     steps:
       - uses: actions/checkout@v5
-        if: ${{ matrix.with_openssl.toggle != 'OFF' }}
         with:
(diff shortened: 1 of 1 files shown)


scripts/verify-installation-unix.sh
```diff
@@ -19,6 +19,14 @@ if [ -n "$BIT_VERSION" ]; then
   CXXFLAGS="$CXXFLAGS -m$BIT_VERSION"
 fi
 
+# enable all compiler warnings
+CXXFLAGS="$CXXFLAGS -Wall"
+
+# treat compiler warnings as errors when the build type is Debug
```
JackPGreen: We aren't actually _checking_ `BUILD_TYPE` but we should be.

(it's not passed into this script either, even though we run the nightly builds with both...)

ihsandemir: Fixed at ffaee6f6a773d85850a30d81f14b6aea4044b4bc

JackPGreen (dismissed):

PR #1340, opened 2025-10-23
Apply _existing_ code format to codebase

JackPGreen:
In https://github.com/hazelcast/hazelcast-cpp-client/pull/946 the codebase was reformatted for consistency, but as this was not enforced the codebase has drifted since then.

Reformatted with `scripts/format-all.sh`.

Post-merge actions:
- [ ] proceed with https://github.com/hazelcast/hazelcast-cpp-client/pull/1342


hazelcast/test/src/sql_test.cpp
```diff
@@ -934,12 +934,13 @@ TEST_F(SqlTest, calling_iterator_next_consecutively)
...
+    auto handler = [this]() {
         sql::sql_statement statement{ client,
-                        (boost::format(
-                                R"(
+                                      (boost::format(
+                                         R"(
                 SELECT * FROM %1%
-            )") % map_name).str() };
+            )") % map_name)
+                                        .str() };
```
ihsandemir: ?

hazelcast/test/src/sql_test.cpp
```diff
@@ -2309,26 +2307,171 @@ TEST_F(sql_encode_test, execute)
 
     std::vector<unsigned char> expected_bytes = {
         // first frame
-        36,  0,   0,   0,   0,   192, 0,
-        4,   33,  0,   0,   0,   0,   0,
-        0,   0,   0,   0,   255, 255, 255, 255, 42,  0,   0,   0,   0,   0,
-        0,   0,   210, 4,   0,   0,   5,   1,
+        36,
```
ihsandemir: doesnt look good.

ihsandemir (changes requested):

hazelcast/test/src/sql_test.cpp
```diff
@@ -426,14 +425,15 @@ class SqlTest
...
     {
         sql::sql_statement statement{ client,
-                        (boost::format(
-                                R"(
+                                      (boost::format(
+                                         R"(
                 SELECT * FROM %1%
-            )") % map_name).str() };
+            )") % map_name)
+                                        .str() };
```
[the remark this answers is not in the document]
ihsandemir: > @ihsandemir are we able to move anywhere on this?

No, I do not like this formatting. We need to fix it. Otherwise the code looks badly formatted. We need to understand why it formats that way, and then we can fix those places.

hazelcast/test/src/sql_test.cpp
JackPGreen: As suggested, I've let the formatter have another try on these sections after removing any pre-exisiting formatting and it's done a much better job.

fe0b9ffd2b56fc7a0ef0e7f34b71efcf2c600636

hazelcast/test/src/sql_test.cpp
JackPGreen: Ok, this change is bad I agree - it was deliberately formatted - fixed in 8f7d201c96e328e4ddba03fd393ceb42fd6b6fa0

PR #1347, opened 2025-10-31
Update the test server version to 5.6.0 [HZ-5191]

ihsandemir:
Updated the test server version to the new released latest server version 5.6.0.


ihsandemir:
> What about
> 
> https://github.com/hazelcast/hazelcast-cpp-client/blob/387f3f878b8a4644dc8071624f96a30db41a097a/.github/config.env#L4

Good catch. fixed e4cd8ecb3661dff3d7dac4b9e521ed94ff8409e6

JackPGreen (approved):

PR #1348, opened 2025-10-31
Add the pthread linkage while verifying the build [HZ-5191]

ihsandemir:
Added the pthread linkage while verifying the build. 32-bit linux builds were failing due to this. 

Example green build at here:

- [32 bit linux](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/18982348106)
- [nightly-Windows](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/18983809078)
- [nightly-Ubuntu-x86_64](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/18983801725)
- [nightly-macOS-x86_64](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/18983793895)

### DIFF
--- a/.github/actions/build-test/windows/action.yml
+++ b/.github/actions/build-test/windows/action.yml
@@ -98,7 +98,8 @@ runs:
         # - Thread - https://www.boost.org/libs/thread/
         # - Chrono - https://www.boost.org/libs/chrono/
         # - Atomic - https://www.boost.org/libs/atomic/
-        .\b2 address-model=${{ inputs.ARCH_ADDRESS_MODEL }} toolset=msvc --with-thread --with-chrono --with-atomic install        
+        # d0 turns off info logging but only prints warnings and errors.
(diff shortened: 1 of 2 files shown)


JackPGreen:
> > > Also added workflow input parameter commit ref.
> > 
> > 
> > Why did we need this?
> 
> I needed to be able to run the workflow on my fıx branch and hence I needed that we checkout a certaın branch for the testing of the PR change.

You should be able to do that already? If you run the workflow _from your branch_ it should implicitly use the branch as source.

ihsandemir:
> > > > Also added workflow input parameter commit ref.
> > > 
> > > 
> > > Why did we need this?
> > 
> > 
> > I needed to be able to run the workflow on my fıx branch and hence I needed that we checkout a certaın branch for the testing of the PR change.
> 
> You should be able to do that already? If you run the workflow _from your branch_ it should implicitly use the branch as source.

You are right. I reverted back. See 5565ab5902fd47b47dfe025401d2c9882727db37

JackPGreen (approved):

PR #1349, opened 2025-10-31
Suppress the info messages while installing Boost

ihsandemir:
Suppressed the info messages e.g. file copy info logs, while installing Boost. Default prints too many logs.

./b2 --help-options
...
  * -dn; Enables output of diagnostic messages. The debug level 'n' and all below it are enabled by this option.
  * -d+n; Enables output of diagnostic messages. Only the output for debug level 'n' is enabled.

Debug Levels:

   Each debug level shows a different set of information. Usually with higher
  levels producing more verbose information. The following levels are supported:

  * 0; Turn off all diagnostic output. Only errors are reported.
  * 1; Show the actions taken for building targets, as they are executed.
  * 2; Show quiet actions and display all action text, as they are executed.
  * 3; Show dependency analysis, and target/source timestamps/paths.
  * 4; Show arguments of shell invocations.
  * 5; Show rule invocations and variable expansions.
  * 6; Show directory/header file/archive scans, and attempts at binding to targets.
  * 7; Show variable settings.
  * 8; Show variable fetches, variable expansions, and evaluation of 'if' expressions.
  * 9; Show variable manipulation, scanner tokens, and memory usage.
  * 10; Show execution times for rules.
  * 11; Show parsing progress of Jamfiles.
  * 12; Show graph for target dependencies.
  * 13; Show changes in target status (fate).


JackPGreen (dismissed):
nit - would be nice to add this as a comment to the `-d0` arg usage as it's not exactly descriptive.

ihsandemir:
> nit - would be nice to add this as a comment to the `-d0` arg usage as it's not exactly descriptive.

fixed at 2a6a59d1723305262b54fdc73aafed293cd89429

JackPGreen (approved):

PR #1339, opened 2025-10-23
DOC-527-2: delete duplicate content

amandalindsay:
Remove duplicate content from C++ Ref Manual & Readme

Following on from DOC-470 which moved install and getting started info into the main docs site, this will remove the duplicate information from the C++ Ref Manual & Readme files, and point to the docs for this information

PR #1367, opened 2025-11-07
Update to new Release cycle 5.6.0

dmitriyrazboev:

PR #1368, opened 2025-11-07
Remove the ref manual generation since it is not needed after #1339

ihsandemir:
Remove the ref manual generation since it is not needed after #1339.


ihsandemir:
> We also need to revert the PR builder assertions added in #1351

Well i see the changes already incorporated into https://github.com/hazelcast/hazelcast-cpp-client/pull/1367 I may close this PR if that is merged.

JackPGreen:
> > We also need to revert the PR builder assertions added in #1351
> 
> Well i see the changes already incorporated into #1367 I may close this PR if that is merged.



> > We also need to revert the PR builder assertions added in #1351
> 
> Well i see the changes already incorporated into #1367 I may close this PR if that is merged.

Ok - id rather we fixed this issue separately to bumping versions to have a clearer history of what changed.

EDIT - as in, let’s do it here, merge and then just update the versions in the other PR.

PR #1404, opened 2026-01-20
Fix indentation

dmitriyrazboev:
Indentation fix for future doxygen release automation


dmitriyrazboev:
> is there a way to test without releasing?

I had tested reqexp, but no, only next release will show properly, anyhow changes is quite small just

ihsandemir (approved):

PR #1410, opened 2026-03-05
feat: add IO_THREAD_COUNT client property [HZ-5386]

ihsandemir:
Instead of using a single IO thread, now we use similar to Java client, 3 threads. This was causing throughput to suffer where IO thread would be a bottleneck and concurrent max thread count to increase the throughput would saturate at 160 threads.

Each connection still work on one of these 3 threads. 

feat: add IO_THREAD_COUNT client property
Add two new client property constant to control threading:
- IO_THREAD_COUNT (default 3): number of IO threads for networking

Includes static string definitions, constructor initializers, and
accessor methods following the existing client_properties pattern.

Design doc at https://github.com/hazelcast/hazelcast-cpp-client/blob/master/docs/design/pr-1410-2026-03-05-cpp-io-thread-count-design.md


hazelcast/src/hazelcast/client/client_impl.cpp
```diff
@@ -989,6 +989,10 @@ const std::string client_properties::INTERNAL_EXECUTOR_POOL_SIZE =
   "hazelcast.client.internal.executor.pool.size";
 const std::string client_properties::INTERNAL_EXECUTOR_POOL_SIZE_DEFAULT = "3";
 
+const std::string client_properties::IO_THREAD_COUNT =
+  "hazelcast.client.io.thread.count";
+const std::string client_properties::IO_THREAD_COUNT_DEFAULT = "3";
```
JackPGreen: It would be _nice_ if there was documentation to explain why `3` was chosen. Even if it just links to the config in the Java client or something.

hazelcast/src/hazelcast/client/network.cpp
```diff
@@ -114,24 +114,31 @@ ClientConnectionManagerImpl::start()
...
-      new boost::asio::ip::tcp::resolver(io_context_->get_executor()));
-    socket_factory_.reset(new internal::socket::SocketFactory(
-      client_, *io_context_, *io_resolver_));
-    auto guard = boost::asio::make_work_guard(*io_context_);
-    io_guard_ = std::unique_ptr<
-      boost::asio::executor_work_guard<boost::asio::io_context::executor_type>>(
-      new boost::asio::executor_work_guard<
-        boost::asio::io_context::executor_type>(std::move(guard)));
+    auto& props = client_.get_client_properties();
+    int io_thread_count = props.get_integer(props.get_io_thread_count());
```
JackPGreen: With a known thread count, is it worth informing the various vectors they need to be of size `io_thread_count`? (Google-fu says `reserve`).

JackPGreen (changes requested):

hazelcast/src/hazelcast/client/client_impl.cpp
ihsandemir: That is internal.

JackPGreen (dismissed):

JackPGreen (approved):

PR #1412, opened 2026-03-12
Change client IO pipeline to be inline with Java client design: IO thread pool, response threads, lock-free writes [HZ-5387]

ihsandemir:
## Summary

We like to improve the throughput and performance of the client following the IO thread PR #1410 . We hope to gain some improvements in throughput and latency by making the IO pipeline design similar to Java current design as possible.

- **Dedicated response thread pool** — offloads invocation completion from IO threads to dedicated `ClientResponseHandler` threads (`RESPONSE_THREAD_COUNT` property, default 2), preventing IO thread stalls
- **Lock-free write queue with batch flushing** — replaces per-write dispatch with `boost::lockfree::queue` and scatter-gather `async_write`, coalescing multiple outbound messages into single syscalls
- **Global concurrent invocation registry** — uses `boost::concurrent_flat_map` for O(1) correlation ID lookup, replacing per-connection maps
- **CallIdSequence for correlation IDs** — mirrors Java client pattern with optional backpressure support
- **Fix race condition in backup ack handling** — adds double-check in `ClientInvocation::notify()` after storing `pending_response_` to prevent invocations from getting stuck when backup acks arrive concurrently with primary responses (mirrors Java `BaseInvocation.notifyResponse` pattern)

Design doc at https://github.com/hazelcast/hazelcast-cpp-client/blob/master/docs/design/pr-1412-2026-03-05-cpp-io-pipeline-sync-to-java-design.md

## Test plan

- [x] Run `ClientMapTest.testGet` with 1000 entries — previously hung due to backup ack race condition
- [x] Run unit tests against 2-node cluster to validate backup-aware operations
- [x] Verify no regression in existing tests (all `HazelcastTests1-8`)
- [x] Benchmark sequential put/get throughput to confirm pipeline improvements


docs/plans/2026-03-05-cpp-pipeline-rearchitecture-design.md
```diff
@@ -0,0 +1,290 @@
+# C++ Client Pipeline Rearchitecture Design
+
+**Date:** 2026-03-05
+**Status:** Approved
+**Base branch:** `fixIoThreadCount` (PR #1410 on hazelcast-cpp-client)
+**Goal:** Close the 4.3x throughput gap (103K vs 446K ops/sec PUT 4B) by aligning C++ client networking architecture
```
JackPGreen:
```suggestion
**Date:** 2026-03-05 \
**Status:** Approved \
**Base branch:** `fixIoThreadCount` ([PR](https://github.com/hazelcast/hazelcast-cpp-client/pull/144)) \
**Goal:** Close the 4.3x throughput gap (103K vs 446K ops/sec PUT 4B) by aligning C++ client networking architecture
```

Difficult to read otherwise

Image

JackPGreen (changes requested):
I've done a very limited review to the best of my abilities and added some comments.

hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -48,6 +57,7 @@ class BaseSocket : public hazelcast::client::socket
       , connect_timeout_(connect_timeout_in_millis)
       , resolver_(io_resolver)
       , socket_(io)
+      , write_queue_(128)
```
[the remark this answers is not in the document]
ihsandemir: fixed at 3ec75e4babb1c2ad30cb5cd0b1b4db7da9bf87e0

JackPGreen (changes requested):

JackPGreen (dismissed):

hazelcast/include/hazelcast/client/connection/Connection.h
```diff
@@ -122,18 +122,16 @@ class HAZELCAST_API Connection
...
 
     std::chrono::steady_clock::time_point last_write_time() const;
 
+    spi::impl::ClientResponseHandler& get_response_handler() const;
+
     friend std::ostream& operator<<(std::ostream& os,
                                     const Connection& connection);
 
     ReadHandler read_handler;
-    std::unordered_map<int64_t, std::shared_ptr<spi::impl::ClientInvocation>>
```
emreyigit, line 135: What will happen to listeners registered with `nullptr` on connection close? I think there is leakage on listener invocations.

hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
```diff
@@ -176,26 +179,28 @@ class HAZELCAST_API ClientInvocation
...
+    std::atomic<int8_t> backup_acks_expected_{ -1 };
 
     /**
      * Contains the pending response from the primary. It is pending because it
      * could be that backups need to complete. Note that we do not need thread
      * safety since these are only read/write from the same io thread for the
      * connection.
      */
-    std::shared_ptr<protocol::ClientMessage> pending_response_;
+    boost::atomic_shared_ptr<std::shared_ptr<protocol::ClientMessage>>
```
emreyigit, line 193: What's the point of indirection here?

emreyigit (changes requested):

hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
ihsandemir: because at language level below C++20 (we allow 14 and 17), there is no support for atomic std::shared_ptr, see [this](https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2.html). Hence, ı need to use `boost::atomic_shared_ptr` but `boost::atomic_shared_ptr` only works with the `boost::shared_ptr`. Well, at all other places I need to use std::shared_ptr for the same object, hence, in order to support this I had to wrap the std::shared_ptr in boost::shared_ptr to be used with boost::atomic_shared_ptr. Yes, I also do not like it but until we switch to C++20, we need to live with it. The cost these wrappers are very minor to the performance, the pointer implementations are pretty efficient.

hazelcast/include/hazelcast/client/connection/Connection.h
ihsandemir: I am not sure if I understood your question but let me explain the path of delivering the event when received:
`Connection::handle_client_message` --> `listener_service_impl::handle_client_message` --> `listener_service_impl::handle_client_message` --> `listener_service_impl::process_event_message` (null checks inside this method exist, if no invocation waiting just drop the event)

Does this answer your question? I do not see the leakage path. can you explain more?

emreyigit: As I see, event invocations are registered with `nullptr` at `get_permission_to_notify` which was fine before since when connection is closed, the invocations in it were freeing apart from its connection value. Now, `notify_connection_closed` will fail to `deregister_invocation` since listener invocations are registered with nullptr.

hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
emreyigit: I see. My concern is mostly about possible issues on dereferencing. If you think this is fine, I have no problem.

hazelcast/include/hazelcast/client/connection/Connection.h
ihsandemir: `get_permission_to_notify` is only called from `ClientInvocation::notify` or `ClientInvocation::notify_exception` , these two paths are only if the received client message is NOT and event, but for regular request responses. For events, the delivery of the received message (`Connection::handle_client_message`) is routed to the listener service and in that path there will be no call to the `ClientInvocation::notify`. Also, when processing the original response to the listener registration request message, we pass the `erase` flag as `false` hence the invocation stays at the invocations map. 
```
ClientInvocation::notify(const std::shared_ptr<protocol::ClientMessage>& msg,
                         bool erase)
```
I hope this makes sense. There is an additional improvement we miss which is different from Java client is this: We still keep event registration invocation in the map for the delivery of events to the event handler. In Java, it is not kept but they keep a map of handlers in the connection to find which handler to deliver the event to. It may be the next PR for improvement. But the current event delivery also works, we just keep a fatter invocation object instead of the event handler itself in the map. By the way, this invocation for the event which stays in the map is removed if the listener is de-registered (`listener_service_impl::remove_event_handler`) hence it will not leak.

ihsandemir: fixed at 6ee9f980727c8c1fb6a1bbd14470e26e35f720f9

emreyigit (approved):

PR #1413, opened 2026-03-16
fix: make RC logs visible in Windows GitHub Actions [HZ-5386]

ihsandemir:
## Summary

- Redirect Windows remote controller stdout/stderr to `rc_stdout.log` (RC runs in a separate minimized console window via `start /MIN`, so output is lost in CI)
- Print RC log contents after tests complete in both success and failure paths of `test-windows.bat`

## Test plan

- [ ] Verify Windows CI workflow runs successfully
- [ ] Confirm RC logs appear in GitHub Actions output after test completion
- [ ] Verify RC startup failure path also prints logs

### DIFF
--- a/scripts/start-rc.bat
+++ b/scripts/start-rc.bat
@@ -82,4 +82,4 @@ set CLASSPATH="hazelcast-remote-controller-%HAZELCAST_RC_VERSION%.jar;hazelcast-
 echo "Starting Remote Controller ... enterprise ...Using classpath: %CLASSPATH%"
 
 echo "Starting hazelcast-remote-controller"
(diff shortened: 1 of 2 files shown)


JackPGreen (changes requested):
The change is fine in isolation but a bit confused about how it fits in with the other platforms.

Specifically, the Windows script starts RC and echos to a file, but the Unix one just starts and prints to console:
https://github.com/hazelcast/hazelcast-cpp-client/blob/6a064f94bcf28592abb4f139764f42335bafa27a/scripts/start-rc.sh#L127-L132

PR #1428, opened 2026-03-30
Upgrade CI clang-format from v18 to v22 [HZ-5387]

ihsandemir:
## Summary
- The `enforce-code-formatting` CI step installs the default `clang-format` package on Ubuntu 24.04, which is version 18
- I locally use clang-format 22, which produces different formatting for the same `.clang-format` config
- This causes false failures on PRs (e.g., #1412) where code is correctly formatted per v22 but differs under v18
- Fix: install `clang-format-22` from the official LLVM apt repository to match the developer version

Sample run passes [here](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/23738265756/job/69148434329).

## Test plan
- [x] Verify the `enforce-code-formatting` job passes on this PR
- [ ] Re-run the `enforce-code-formatting` job on PR #1412 to confirm it passes

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -46,9 +46,8 @@ jobs:
         id: test-membership
         uses: hazelcast/hazelcast-tpm/membership@main
         with:
-          organization-name: hazelcast
           member-name: ${{ github.actor }}
-          token: ${{ secrets.GH_TOKEN }}
+          token: ${{ github.token }}
           
   # ensure PR is trusted
   ensure-membership:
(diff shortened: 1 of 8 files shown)


.github/workflows/build-pr.yml
```diff
@@ -248,8 +248,11 @@ jobs:
     steps:
       - name: Install Necessary Packages
         run: |
+          wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
```
JackPGreen: `Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).`

https://github.com/llvm/llvm-project/issues/55784

JackPGreen (approved):

PR #1429, opened 2026-03-30
Remove redundant `actions/checkout` token

JackPGreen:
We don't need to supply a token to `actions/checkout`, the implicit `${{ github.token }}` is fine and preferable to a long-lived PAT.


ihsandemir (commented):
does this change contradict with https://github.com/hazelcast/hazelcast-cpp-client/pull/1431?

JackPGreen:
> does this change contradict with #1431?

It does not.

ihsandemir (approved):

PR #1433, opened 2026-04-07
Fix Miscasting [API-2378] [HZ-5417]

### LINKED ISSUE #1432: Infinite loop in `twos_complement` [API-2378] [HZ-5417]
While doing PR review, realized that there is an infinite loop in `twos_complement` at `client_impl.cpp`. The auto value is unsigned which never reaches negative to break the loop.

emreyigit:
Refactored the casting. 

closes #1432


ihsandemir (commented):
can you add a test to verify it was actually a problem?

hazelcast/test/src/HazelcastTests2.cpp
```diff
@@ -140,6 +140,24 @@ TEST_F(DecimalTest, cascading_carry_bit_test)
...
+    // carry propagates out of the only byte; the signed loop index must exit
+    // cleanly
+    std::vector<int8_t> v = { 0 };
+    pimpl::twos_complement(v);
+    ASSERT_EQ((std::vector<int8_t>{ 0 }), v);
+}
+
+TEST_F(DecimalTest, twos_complement_multi_byte_full_carry)
+{
+    // carry propagates out of every byte; a size_t loop index would wrap to
```
ihsandemir, line 154: what does this comment mean? you changed size_t usage to int int he loop at this PR.

ihsandemir (commented):
Can you also add a non-trivial array which causes infinite loop without this fix but works after this fix? This will validate the fix.

hazelcast/test/src/HazelcastTests2.cpp
emreyigit: Yes,  only way to test it is calling the function and verifying its result. I just explained the situation what was it before.  Also, I couldn't find any test on `twos_complement`, so I added single and multiple vector size test which verifies both loop and the function.

ihsandemir: Actually, there were existing tests for the complement (see `DecimalTest::....` test cases above your added test. e.g. `TEST_F(DecimalTest, carry_bit_test)`) but not for this exact scenario. you need a specific value to actually see the index goes all the way to 0. Your added tests are OK but they do not catch the infinite loop problem, right? Does you test catch the failure without the fix?

emreyigit: The test should verify the loop as well since it was discarding the carry bit before due to unsigned int casting.

ihsandemir (approved):

PR #1434, opened 2026-04-10
clang-format fix

ihsandemir:
clang-format fix. Ran scripts/format-all.sh and opened the PR.


JackPGreen (approved):
I have no strong opinions on if we should or should not make such changes.

Issue #1409, opened 2026-02-09
[HZ-5351] C++ Client doesn't sort field names during compact schema creation


yuce:
The compact `schema` class doesn't sort all of its fields by name before creating the Rabin fingerprint.
Its constructor receives an unordered map of fields, so it is not possible to pass a sorted list of fields to it.
Its best to sort the fields in the constructor, like the Java Client.

The expected behavior is: `schema` objects with the same fields must have the same Rabin fingerprint, regardless of their order in the schema.

See https://github.com/hazelcast/hazelcast-nodejs-client/issues/1576 for the Node.js Client issue.

ihsandemir:
I analyzed the issue. We followed the exact same implementation in C++ client as in Java for schema. 

We actually sort the fields here at line [3426](https://github.com/hazelcast/hazelcast-cpp-client/blob/bf5b1d89068fc2c66fd7c4eefceabe0e86c6e740/hazelcast/src/hazelcast/client/compact.cpp#L3426). This is a sorted map, see [here](https://en.cppreference.com/w/cpp/container/map.html).

I do not see any problem in C++ implementation. @yuce  Can you share with me any reproducer if you still think there is a problem? Otherwise, I like to close the issue as not doing.

yuce:
@ihsandemir The field definition is passed as an `unordered_map` to the method: https://github.com/hazelcast/hazelcast-cpp-client/blob/bf5b1d89068fc2c66fd7c4eefceabe0e86c6e740/hazelcast/src/hazelcast/client/compact.cpp#L3418

The doc for `std::unordered_map`: https://en.cppreference.com/w/cpp/container/unordered_map.html
> Internally, the elements are not sorted in any particular order,

The line you refer to creates a new map by iterating the unordered map, so the order of field definitions is undefined.

ihsandemir:
> [@ihsandemir](https://github.com/ihsandemir) The field definition is passed as an `unordered_map` to the method:
> 
> [hazelcast-cpp-client/hazelcast/src/hazelcast/client/compact.cpp](https://github.com/hazelcast/hazelcast-cpp-client/blob/bf5b1d89068fc2c66fd7c4eefceabe0e86c6e740/hazelcast/src/hazelcast/client/compact.cpp#L3418)
> 
> Line 3418 in [bf5b1d8](/hazelcast/hazelcast-cpp-client/commit/bf5b1d89068fc2c66fd7c4eefceabe0e86c6e740)
> 
[...]

PR #1440, opened 2026-04-16
Use stable sort for compact schema fixed-size fields [HZ-5351] 

ihsandemir:
## Summary

Fixes the issue raised in #1409 (comment [4241592093](https://github.com/hazelcast/hazelcast-cpp-client/issues/1409#issuecomment-4241592093)).

`schema::schema` sorted fixed-size fields by size with `std::sort`, which is
not stable. Fields of the same kind could end up in implementation-defined
order, so the C++ client sometimes may produce different field offsets than the
Java reference client for the same `schema_id`. This would break the cross-language
deserialization of compact values with multiple same-size fixed fields.

Switch the `std::sort` to `std::stable_sort` so the alphabetical order seeded by the
preceding `std::map<std::string, field_descriptor>` traversal is preserved,
matching `com.hazelcast.internal.serialization.impl.compact.Schema`. 
`std::stable_sort` has been in `<algorithm>` since C++98
so it works under both the C++11 minimum and C++14-with-Boost>=1.82 build
modes.

## Tests added

- `CompactSchemaStableSortTest.fixed_size_fields_with_same_kind_keep_alphabetical_order`
  — unit test verifying that the size and kind matching fields are ordered alphabetically.
- `CompactSchemaStableSortIntegrationTest.cpp_schema_layout_matches_java_for_same_size_fields`
  — replicates the schema and compares (kind, offset, index, bit_offset)
  against what the Java server computes.
- `CompactSchemaStableSortIntegrationTest.cpp_write_java_read_round_trip_with_same_size_fields`
  — C++ put → Java GenericRecord read round-trip with distinguishable values and values are as expected.

## Test plan

- [x] `./build/hazelcast/test/src/client_test --gtest_filter="CompactSchemaStableSort*"` passes.

### DIFF
--- a/hazelcast/src/hazelcast/client/compact.cpp
+++ b/hazelcast/src/hazelcast/client/compact.cpp
@@ -3438,7 +3438,12 @@ schema::schema(
         }
     }
 
-    std::sort(
+    // Use stable_sort so that fields of the same fixed-size kind keep the
+    // alphabetical order seeded by the preceding std::map traversal. This
+    // matches the Java client's stable sort in
+    // com.hazelcast.internal.serialization.impl.compact.Schema and keeps
+    // field offsets identical across clients for the same schema_id.
+    std::stable_sort(
       fixed_size_fields.begin(), fixed_size_fields.end(), kind_size_comparator);
 
     int offset = 0;
--- a/hazelcast/src/hazelcast/client/compact.cpp
+++ b/hazelcast/src/hazelcast/client/compact.cpp
@@ -3438,11 +3438,11 @@ schema::schema(
         }
     }
 
-    // Use stable_sort so that fields of the same fixed-size kind keep the
-    // alphabetical order seeded by the preceding std::map traversal. This
-    // matches the Java client's stable sort in
-    // com.hazelcast.internal.serialization.impl.compact.Schema and keeps
-    // field offsets identical across clients for the same schema_id.
+    // Fixed size fields should be in descending order of size in bytes.
+    // For ties, the alphabetical order(ascending) of the field name will
+    // be used. Since, `fields` is sorted at this point, and the `sort`
+    // method is stable, only sorting by the size in bytes is enough for
+    // this invariant to hold.
     std::stable_sort(
       fixed_size_fields.begin(), fixed_size_fields.end(), kind_size_comparator);
 
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -46,7 +46,7 @@ jobs:
         id: test-membership
         uses: hazelcast/hazelcast-tpm/membership@main
         with:
-          member-name: ${{ github.actor }}
+          member-name: ${{ github.triggering_actor }}
           token: ${{ secrets.GH_TOKEN }}
           
   # ensure PR is trusted
@@ -61,7 +61,7 @@ jobs:
     steps:
       - name: Report
         shell: bash
-        run: echo "User ${{ github.actor }} is trusted for test execution"          
+        run: echo "User ${{ github.triggering_actor }} is trusted for test execution"          
 
 
   # get
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -1189,7 +1189,7 @@ ClientClusterServiceImpl::get_member(boost::uuids::uuid uuid) const
     if (it == members_view_ptr->members.end()) {
         return boost::none;
     }
-    return { it->second };
+    return boost::optional<member>(it->second);
 }
 
 std::vector<member>
--- a/hazelcast/test/src/compact/compact_schema_stable_sort_test.h
+++ b/hazelcast/test/src/compact/compact_schema_stable_sort_test.h
@@ -1,0 +1,342 @@
+/*
+ * Copyright (c) 2008-2026, Hazelcast, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <sstream>
+#include <vector>
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include <boost/format.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+#include "hazelcast/client/serialization/serialization.h"
+#include "hazelcast/client/serialization/pimpl/compact/schema.h"
+#include "hazelcast/client/spi/ClientContext.h"
+
+#include "compact_test_base.h"
+#include "../TestHelperFunctions.h"
+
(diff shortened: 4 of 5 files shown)


JackPGreen:
Rather than changing the sort algorithm, couldn't we ensure the items are comparable in such a way that they will always sort in the right order?

ihsandemir:
> Rather than changing the sort algorithm, couldn't we ensure the items are comparable in such a way that they will always sort in the right order?

The problem is when items are equal using the sorting function. When equal how would you sort them? The `std::sort` does implementation dependent. But `std::stable_sort` does it deterministically. Does it make sense?

JackPGreen:
> > Rather than changing the sort algorithm, couldn't we ensure the items are comparable in such a way that they will always sort in the right order?
> 
> The problem is when items are equal using the sorting function. When equal how would you sort them? The `std::sort` does implementation dependent. But `std::stable_sort` does it deterministically. Does it make sense?

I'm not sure what the items are to answer properly, but my point was that if they are comparable-but-equals, but actually they _should_ be ordered in a certain way (hence stable sort), the method of comparison is the issue.

ihsandemir:
> > > Rather than changing the sort algorithm, couldn't we ensure the items are comparable in such a way that they will always sort in the right order?
> > 
> > 
> > The problem is when items are equal using the sorting function. When equal how would you sort them? The `std::sort` does implementation dependent. But `std::stable_sort` does it deterministically. Does it make sense?
> 
> I'm not sure what the items are to answer properly, but my point was that if they are comparable-but-equals, but actually they _should_ be ordered in a certain way (hence stable sort), the method of comparison is the issue.

We pass the comparator `kind_size_comparator` to `std::sort` but this is not enough for the `stable_sort` capability. The comparator just will say if 2 items are equal greater or less. When equal which item will be the first is implementation dependent. If lucky, you will get the same as java. hence, we avoid this possibility and make it more deterministic this way. This is the same as in Java client.

JackPGreen:
> > > > Rather than changing the sort algorithm, couldn't we ensure the items are comparable in such a way that they will always sort in the right order?
> > > 
> > > 
> > > The problem is when items are equal using the sorting function. When equal how would you sort them? The `std::sort` does implementation dependent. But `std::stable_sort` does it deterministically. Does it make sense?
> > 
> > 
> > I'm not sure what the items are to answer properly, but my point was that if they are comparable-but-equals, but actually they _should_ be ordered in a certain way (hence stable sort), the method of comparison is the issue.
> 
> We pass the comparator `kind_size_comparator` to `std::sort` but this is not enough for the `stable_sort` capability. The comparator just will say if 2 items are equal greater or less. When equal which item will be the first is implementation dependent. If lucky, you will get the same as java. hence, we avoid this possibility and make it more deterministic this way. This is the same as in Java client.

My suggestion is to improve the `kind_size_comparator` instead.
[The comment on the Java client](https://github.com/hazelcast/hazelcast/blob/11036e3f88aef3db0ee45efb4b9c30c548c5ba3b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/compact/Schema.java#L91-L95) seems to suggest how they should be sorted (by size and then by name) - it seems safer (and easier to reason) to sort explicitly based on what you want, than to rely on the list already being sorted and being "careful" with a subsequent sort?

E.G. could `kind_size_comparator` sort by both size _and then_ name?

yuce:
> Rather than changing the sort algorithm, couldn't we ensure the items are comparable in such a way that they will always sort in the right order?

The stable sort is used for that purpose.
And all other clients use the counterparts of `std::stable_sort` for this code.
IMO no need to complicate it.

hazelcast/src/hazelcast/client/compact.cpp
```diff
@@ -3438,7 +3438,12 @@ schema::schema(
         }
     }
 
-    std::sort(
+    // Use stable_sort so that fields of the same fixed-size kind keep the
+    // alphabetical order seeded by the preceding std::map traversal. This
+    // matches the Java client's stable sort in
+    // com.hazelcast.internal.serialization.impl.compact.Schema and keeps
+    // field offsets identical across clients for the same schema_id.
```
yuce, line 3445: Could you copy the comment from the Java client instead of using this one? I think it gives better context.

yuce (changes requested):

ihsandemir:
> > > > > Rather than changing the sort algorithm, couldn't we ensure the items are comparable in such a way that they will always sort in the right order?
> > > > 
> > > > 
> > > > The problem is when items are equal using the sorting function. When equal how would you sort them? The `std::sort` does implementation dependent. But `std::stable_sort` does it deterministically. Does it make sense?
> > > 
> > > 
> > > I'm not sure what the items are to answer properly, but my point was that if they are comparable-but-equals, but actually they _should_ be ordered in a certain way (hence stable sort), the method of comparison is the issue.
> > 
> > 
> > We pass the comparator `kind_size_comparator` to `std::sort` but this is not enough for the `stable_sort` capability. The comparator just will say if 2 items are equal greater or less. When equal which item will be the first is implementation dependent. If lucky, you will get the same as java. hence, we avoid this possibility and make it more deterministic this way. This is the same as in Java client.
> 
> My suggestion is to improve the `kind_size_comparator` instead. [The comment on the Java client](https://github.com/hazelcast/hazelcast/blob/11036e3f88aef3db0ee45efb4b9c30c548c5ba3b/hazelcast/src/main/java/com/hazelcast/internal/serialization/impl/compact/Schema.java#L91-L95) seems to suggest how they should be sorted (by size and then by name) - it seems safer (and easier to reason) to sort explicitly based on what you want, than to rely on the list already being sorted and being "careful" with a subsequent sort?
> 
> E.G. could `kind_size_comparator` sort by both size _and then_ name?

@JackPGreen Yes, that would be possible but then we need to also add a new field in the `field_descriptor` to carry the field `name`, which we are not keeping at the moment. And in the comparison method we can compare the field names in addition to the kind size like this:
```
bool
kind_size_comparator(const field_descriptor* i, const field_descriptor* j)
{
    auto i_kind_size = field_operations::get(i->kind).kind_size_in_byte_func();
    auto j_kind_size = field_operations::get(j->kind).kind_size_in_byte_func();
    if (i_kind_size == j_kind_size) {
        return i->name < j->name;
    }
    return i_kind_size > j_kind_size;
}
```

If you think about it, this can also be done for Java as well but it is not done currently:
```
        fixedSizeFields.sort(
                Comparator.comparingInt(
                        d -> fieldOperations(((FieldDescriptor) d).getKind()).kindSizeInBytes()
                ).reversed()
        );
```
We can easily change the Java client like this as well:
```
        fixedSizeFields.sort(
                (o1, o2) -> {
                    int f1Size = fieldOperations(o1.getKind()).kindSizeInBytes();
                    int f2Size = fieldOperations(o2.getKind()).kindSizeInBytes();
                    if (f1Size == f2Size) {
                        return o1.getFieldName().compareTo(o2.getFieldName());
                    }
                    return Integer.compare(f1Size, f2Size);
                }
        );
```

I would like to be in sync in terms of implementation. Both solves the problem. But if we like to change lets change both Java and C++ at the same time. I am neutral on this. I do not think it is a big deal as long as we have proper documentation for future implementors. what do you think? Are you strong on changing it?

hazelcast/src/hazelcast/client/compact.cpp
ihsandemir: fixed at 2e699dd39cb9078707be99cd78980972900117d6

JackPGreen:
> I would like to be in sync in terms of implementation. Both solves the problem. But if we like to change lets change both Java and C++ at the same time. I am neutral on this. I do not think it is a big deal as long as we have proper documentation for future implementors. what do you think? Are you strong on changing it?

My _preference_ is for approach I suggested, but as it's consistent with the implementations in other clients I think it's acceptable.

JackPGreen (approved):

yuce (approved):

PR #1444, opened 2026-05-04
Open "ClientMapTest::test*Ttl* tests" [HZ-5311]

### LINKED ISSUE #1390: Enable the ttl based tests for ClientMapTest [API-2375]
#1388 disabled the ttl tests until the Java server side fix is in place. We need to re-enable these tests and remove the test skipping. once the server side is fixed.

ihsandemir:
refactor:
- Opened "ClientMapTest::test*Ttl* tests" [HZ-5311]

- Enhanced ClientMapTest to use structured client configuration parameters which is better printed in the test names.

fixes https://github.com/hazelcast/hazelcast-cpp-client/issues/1390

[HZ-5311]: https://hazelcast.atlassian.net/browse/HZ-5311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/start-rc.bat
JackPGreen, line 1: There is no publish `5.7.0-SNAPSHOT` artifact. Is it a good idea for the client to require this? It might make community contributions more difficult.

ihsandemir: you can start the rc with environment `HZ_VERSION`, hence, any community can use any version. This is the only version that can work with the latest PR changes, hence, we need it.

JackPGreen: Shouldn't we also update (e.g.!) https://github.com/hazelcast/hazelcast-cpp-client/blob/0a527c28b2bc6b9b3ec797ec2cb8821b57f8e3a9/.github/config.env#L4

ihsandemir: good catch. Fixed at fcbe3295da223c4a3a8f72c5114eebe7e710a696

JackPGreen (approved):

JackPGreen (changes requested):
The change is fine, but the PR builder is failing - I don't think (private) OS `SNAPSHOT`s are supported.
Supporting this is non-trivial, and I have concerns about how a non-Hazelcast user would test this locally. Can we wait for `5.7.0` release before this PR?

ihsandemir:
> The change is fine, but the PR builder is failing - I don't think (private) OS `SNAPSHOT`s are supported. Supporting this is non-trivial, and I have concerns about how a non-Hazelcast user would test this locally. Can we wait for `5.7.0` release before this PR?

Master is updated to use 5.7.0 now.

.github/config.env
```diff
@@ -1,4 +1,4 @@
 JAVA_VERSION=17
 JAVA_DISTRIBUTION=temurin
 MAVEN_VERSION=3.9.11
-HZ_VERSION=5.6.0
+HZ_VERSION=5.7.0-SNAPSHOT
```
JackPGreen: This should be reverted.

JackPGreen (changes requested):

.github/config.env
ihsandemir: fixed.

JackPGreen (approved):

PR #1448, opened 2026-05-15
Update test Hazelcast server version to 5.7.0 in start-rc scripts

ihsandemir:
Updated test Hazelcast server version to 5.7.0 in start-rc scripts.


JackPGreen (changes requested):
What about `config.env`?

ihsandemir:
> What about `config.env`?

good catch. fixed at 427217a7402305503a14f347e3bed6cf59ee3b57

JackPGreen (approved):

PR #1454, opened 2026-05-18
Fix flake ID generator new_id() to match Java AutoBatcher semantics (#1452)[HZ-5440]

### LINKED ISSUE #1452: flake_id_generator_impl::new_id() diverges from Java AutoBatcher: many parallel batch fetches, INT64_MIN leak, wasted batches, and validity<=0 bug [HZ-5440]
## Summary

`flake_id_generator_impl::new_id()` is intended to mirror the Hazelcast Java
client's batching logic (`com.hazelcast.flakeidgen.impl.AutoBatcher#newId()`,
as used by `ClientFlakeIdGeneratorProxy`). Because the C++ client exposes an
**async** API (`new_id()` returns `boost::future<int64_t>`) while the Java
`AutoBatcher.newId()` is **synchronous** (it holds `synchronized(this)` across
a blocking `newIdBatch(...).joinInternal()` and loops in `for(;;)` until it has
an ID), the current C++ implementation took shortcuts that diverge from the
reference and cause real defects under concurrency, plus a hard bug for a
valid configuration.

## Affected code

* `hazelcast/src/hazelcast/client/proxy.cpp`
  * `flake_id_generator_impl::new_id()` — lines 1156–1172
  * `flake_id_generator_impl::new_id_batch()` — lines 1174–1189
  * `flake_id_generator_impl::new_id_internal()` — lines 1142–1154
  * `flake_id_generator_impl::Block::Block()` — lines 1010–1016
  * `flake_id_generator_impl::Block::next()` — lines 1018–1033
* `hazelcast/include/hazelcast/client/proxy/flake_id_generator_impl.h`
  * unused `std::mutex lock_;` — line 146
  * `IdBatch` / `Block` are private nested classes — lines 68–137

Reference: `com.hazelcast.flakeidgen.impl.AutoBatcher#newId()`
(`AutoBatcher.java:46-62`) and `ClientFlakeIdGeneratorProxy`
(`ClientFlakeIdGeneratorProxy.java:46-56`) in the Hazelcast Java repository.

### 1. Thundering herd — no single-flight on batch fetch

There is no equivalent of Java's `synchronized(this)`. When `N` threads call
`new_id()` concurrently and the current batch is exhausted/expired, all `N`
threads independently call `new_id_batch()`, producing `N` network requests and
`N` server-reserved ID batches, of which `N-1` are discarded. This defeats the
purpose of prefetching and wastes server-side ID space. The
`std::mutex lock_;` member declared in the header (line 146) is **never used**,
indicating the synchronization was intended but not implemented.

### 2. No retry loop — `INT64_MIN` can be returned to the caller

The `.then()` continuation calls `newBlock->next()` exactly once and returns
its result (`proxy.cpp:1167-1170`). Java instead loops (`for(;;)`) until it
obtains a real ID. If the freshly fetched block yields nothing (batch size 0,
or already expired by the time the continuation runs), C++ returns `INT64_MIN`
to the user as if it were a valid ID.

### 3. Ignored `compare_exchange_strong` result — wasted/leaked batches

```cpp
auto value = newBlock->next();          // ID #0 already consumed from newBlock
auto b = block_.load();
block_.compare_exchange_strong(b, newBlock);   // result ignored
return value;
```

If the CAS fails (another thread installed a different block), `newBlock` is
silently dropped while `value` was already taken from it — the rest of that
server-reserved batch is leaked. If the CAS succeeds, it may overwrite a
*still-valid* fresh block another thread just installed, leaking that batch
instead. Java's double-checked `if (block != this.block) continue;` prevents
both.

### 4. `validity <= 0` semantics missing (hard bug)

Java `Block`:

```java
this.invalidSince = validity > 0 ? Clock.currentTimeMillis() + validity : Long.MAX_VALUE;
```

i.e. a non-positive validity means "never expire". The C++ `Block`
constructor (`proxy.cpp:1013`) does:

```cpp
invalid_since_(std::chrono::steady_clock::now() + validity)
```

unconditionally. With a Java-valid configuration of `prefetch_validity = 0`,
every block is born already expired, so `Block::next()` always returns
`INT64_MIN` and **every** `new_id()` call fails.

### 5. Testability gap

`new_id_batch()` is welded to the network (`invoke(...)`), and `IdBatch` /
`Block` are private nested classes. The Java unit tests
(`com.hazelcast.flakeidgen.impl.AutoBatcherTest`,
`FlakeIdGenerator_IdBatchTest`) are deterministic because `AutoBatcher` takes
an injectable `IdBatchSupplier` ("a separate class due to testability", per its
own Javadoc). There is no equivalent seam in C++, so that coverage cannot be
ported.

## Why a literal Java port is not possible

`new_id()` must keep returning `boost::future<int64_t>` and its continuation
runs on a Boost.Asio IO thread. It therefore can neither block the caller nor
hold a `std::mutex` across the batch network round-trip the way Java does.

## Proposed fix

Extract a testable `auto_batcher` (the C++ analogue of Java's `AutoBatcher`)
and fix `new_id()` with a **single-flight batch fetch** whose winner is
offloaded to the **user executor** — the same pattern `reliable_topic` already
uses (`get_client_execution_service().get_user_executor()` +
`boost::async(executor, ...)`). This also aligns the C++ structure with Java,
where `ClientFlakeIdGeneratorProxy` holds an `AutoBatcher`.

**New units (proxy namespace):**

* `id_batch` — promote `IdBatch`/iterator to a standalone, unit-testable class.
* `auto_batcher` — owns the batching/coalescing state and logic:
  * `boost::atomic_shared_ptr<block> block_;` (lock-free fast path)
  * `std::mutex mutex_;` — guards winner election + in-flight handle only,
    **never** held across the network call
  * `boost::shared_future<std::shared_ptr<block>> fetch_in_progress_;` +
    `uint64_t fetch_generation_` to reset the handle after completion
  * `util::hz_thread_pool& executor_;` (user executor)
  * `std::function<boost::future<id_batch>(int32_t)> batch_supplier_;` — the
    injectable seam (Java's `IdBatchSupplier`); real proxy passes the
    `invoke(...)`-based lambda, unit tests pass a deterministic fake
  * `block` keeps the existing lock-free CAS `next()` plus a new
    non-consuming `bool usable() const`
* `flake_id_generator_impl` becomes thin: owns an `auto_batcher` built with the
  real-fetch lambda + config + user executor; `new_id()` delegates.

**`auto_batcher::new_id()` algorithm:**

1. Lock-free fast path: `block_->next()`; if valid → `make_ready_future`.
2. Under `mutex_`: double-check the block changed (Java's
   `if (block != this.block) continue`); if no fetch in progress, become the
   winner, bump `fetch_generation_`, launch the winner on `executor_`, and
   publish `fetch_in_progress_` as a `shared_future`; copy the handle and
   release the mutex.
3. Winner (on a user-executor thread, blocking is acceptable there):
   `for(;;) { batch = batch_supplier_(batch_size_).get(); block_ = make_shared<block>(...); if (usable) { install; return; } }`.
   Exceptions propagate through the shared_future to all waiters.
4. Every caller (winner-side and losers) attaches a `.then(launch::sync, ...)`
   continuation (never blocking the caller): clear the in-flight handle if
   still the same generation; `f.get()`; `block_->next()` for a distinct ID via
   the existing CAS; if still `INT64_MIN` (more concurrent waiters than
   `prefetch_count`), recurse into `new_id()` (`.unwrap()` keeps the
   `boost::future<int64_t>` signature) — the async analogue of Java's outer
   loop.

**Orthogonal bug fixes (independent of the above):**

* `block` ctor: `invalid_since_ = validity > milliseconds::zero()
  ? now() + validity : steady_clock::time_point::max();` (Java `<= 0` = never
  expire).
* Add non-consuming `block::usable()`:
  `invalid_since_ > now() && num_returned_ < batch_size`.

This yields Java behavioral parity: one outstanding batch request per
generator, no thundering herd, no wasted/leaked batches, never returns
`INT64_MIN`, and the public API stays fully async/non-blocking.

## Testing plan

Add tests next to `FlakeIdGeneratorApiTest` in `HazelcastTests1.cpp`. The
`auto_batcher` seam makes the deterministic ones run with **no server**
(`util::hz_thread_pool` is standalone-constructible).

Ported from `com.hazelcast.flakeidgen.impl`:

| Java test | C++ port |
|---|---|
| `FlakeIdConcurrencyTestUtil.concurrentlyGenerateIds` | helper `concurrently_generate_ids(gen, 4, 100000)`; assert `size == threads*ids` |
| `AutoBatcherTest.when_validButUsedAll_then_fetchNew` | `auto_batcher{3, 10000ms, fake}`, fake returns `id_batch(base,1,size); base+=size`; expect 0,1,2,3 |
| `AutoBatcherTest.when_notValid_then_fetchNew` | same fake, validity 100ms; `new_id()`→0; sleep 150ms; `new_id()`→3 |
| `AutoBatcherTest.concurrencySmokeTest` | helper over `auto_batcher` w/ fake; assert full coverage `[0, threads*ids)` **and** fake call-count == expected #batches (proves single-flight) |
| `FlakeIdGenerator_IdBatchTest.testIterator` / `_hasNextNotCalled` / `_emptyIdBatch` | `id_batch` unit tests: `id_batch(1,10,2)` yields 1,11 then `== end()`; `id_batch(1,10,0)` `iterator()==end()` |
| `FlakeIdGenerator_IdBatchTest.testIterator_hasNextMultipleCalls` | adapt/skip — C++ `IdIterator` is single-pass; Java `hasNext()` idempotency N/A |
| `FlakeIdGeneratorProxyTest`, `*BackpressureTest`, `*MemberIntegrationTest`, `*NodeIdOverflowIntegrationTest` | not ported — member/server-side, outside the client `new_id()` path |

Plus a new **`validity == 0`** regression test (no Java equivalent):
`auto_batcher` with `validity = 0ms`; assert repeated `new_id().get()` over
time never returns `INT64_MIN` and never throws.

Existing real-server tests (`FlakeIdGeneratorApiTest.testStartingValue`,
`testSmoke`, `testAddGetFlakeIdGeneratorIntegrity`) must stay green. Build
Debug (`-fsanitize=address`, `-Werror`) and run the `auto_batcher`/`id_batch`
unit tests under ASan to verify no data races on `block_` /
`fetch_in_progress_` / `mutex_`.

ihsandemir:
Fixes #1452.

## Problem

`flake_id_generator_impl::new_id()` was meant to mirror the Java client's
`com.hazelcast.flakeidgen.impl.AutoBatcher`, but because the C++ client exposes
an async API (`boost::future<int64_t>`) while Java's `AutoBatcher.newId()` is
synchronous, the implementation diverged and had real defects:

1. **Thundering herd** — no equivalent of Java's `synchronized`. N concurrent
   callers hitting an exhausted/expired batch each issued their own
   `newIdBatch` request; N-1 server-reserved batches were discarded.
2. **`INT64_MIN` could reach the caller** — the `.then()` continuation took a
   single ID from the fresh block and returned it with no retry, unlike Java's
   `for(;;)`.
3. **Ignored `compare_exchange_strong` result** — a freshly fetched batch was
   silently dropped (or a still-valid block overwritten), leaking
   server-reserved ID ranges.
4. **`validity <= 0` semantics missing** — Java treats `validity <= 0` as
   "never expire"; C++ computed `now() + validity` unconditionally, so with a
   valid config of `prefetch_validity = 0` every block was born expired and
   every `new_id()` returned `INT64_MIN`.

Secondary: the batching logic was welded to the network and `IdBatch`/`Block`
were private nested classes, so the deterministic Java unit tests could not be
ported.

## Change

- Extract **`auto_batcher`** (C++ analogue of Java `AutoBatcher`) and promote
  **`id_batch`** to its own class. `auto_batcher` takes an injectable
  `id_batch` supplier, mirroring Java's `IdBatchSupplier` ("separate class due
  to testability"); `flake_id_generator_impl` is now thin and delegates.
- `new_id()` keeps a **lock-free fast path**. On the slow path it elects a
  **single fetcher**: concurrent callers coalesce onto one
  `boost::shared_future`; the winning fetch and its `for(;;)` retry loop run on
  the **user executor** (`reliable_topic`'s pattern), so the caller thread and
  IO threads are never blocked; the block is **double-checked** before fetching
  (Java's `if (block != this.block)`); and a caller that still cannot get an ID
  **retries** instead of returning `INT64_MIN` (async analogue of Java's outer
  loop). The single-flight handle is reset via a generation counter so the next
  exhaustion re-elects; supplier/server errors propagate to every coalesced
  caller.
- `block` now treats `validity <= 0` as "never expire" and gains a
  non-consuming `usable()` check used by the fetch loop.
- Fixed the `id_batch` iterator off-by-one (it yielded `batch_size + 1`
  elements; now yields exactly `batch_size`).

## Tests

Server-free unit tests added next to `FlakeIdGeneratorApiTest` in
`HazelcastTests1.cpp`, ported from `com.hazelcast.flakeidgen.impl`:

| Java test | C++ |
|---|---|
| `AutoBatcherTest.when_validButUsedAll_then_fetchNew` | `AutoBatcherTest.when_validButUsedAll_then_fetchNew` |
| `AutoBatcherTest.when_notValid_then_fetchNew` | `AutoBatcherTest.when_notValid_then_fetchNew` |
| `AutoBatcherTest.concurrencySmokeTest` | `AutoBatcherTest.concurrencySmokeTest` (also asserts the supplier is hit ~total/batch_size times, i.e. no thundering herd) |
| `FlakeIdGenerator_IdBatchTest` (iterator / empty) | `FlakeIdBatchTest.testIterator`, `testIterator_emptyIdBatch`, `getters` |
| `FlakeIdConcurrencyTestUtil.concurrentlyGenerateIds` | local `concurrently_generate_ids` helper |

Plus `AutoBatcherTest.validityZeroNeverExpires` (no Java equivalent) covering
defect 4.

Built `Debug` (`-fsanitize=address -Werror`). The new unit tests were written
first and observed to fail (RED) against deliberately-wrong skeletons, then
pass (GREEN) — all 7 green under ASan, including the 4-thread × 100k-ID
concurrency test.


hazelcast/include/hazelcast/client/impl/auto_batcher.h
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * This is the C++ analogue of com.hazelcast.flakeidgen.impl.AutoBatcher. The
+ * Java client's {@code AutoBatcher.newId()} is synchronous: it holds a lock
+ * across a blocking batch fetch and loops until it has an ID. The C++ client
+ * exposes an async API, so the equivalent guarantees are achieved without
+ * blocking the caller or an IO thread:
+ *
+ *  - lock-free fast path while the current batch has IDs;
+ *  - single-flight batch fetch: concurrent callers that hit an exhausted
+ *    batch coalesce onto one outstanding request (no thundering herd);
```
JackPGreen, line 50: Does the Java client have such a mechanism?

hazelcast/include/hazelcast/client/impl/auto_batcher.h
```diff
@@ -0,0 +1,117 @@
...
+ *  - the winning fetch and its retry loop run on the user executor, so the
+ *    caller thread and IO threads are never blocked;
```
JackPGreen: Can we define `winning` or use an alternative word?

JackPGreen: How can the thread be blocked if it's async + lock free?

hazelcast/include/hazelcast/client/impl/auto_batcher.h
```diff
@@ -0,0 +1,117 @@
...
+ *  - never resolves to INT64_MIN: a caller that cannot get an ID from the
+ *    fresh batch transparently retries (async analogue of Java's for(;;)).
```
JackPGreen: I don't think these two concepts are connected (what is considered a valid response + the retry mechanism).

hazelcast/src/hazelcast/client/proxy.cpp
JackPGreen, line 1: If the batcher is split for testability, does it make sense for it's implementation to be coalesced?

hazelcast/src/hazelcast/client/proxy.cpp
```diff
@@ -1007,171 +1008,261 @@ IListImpl::ListListenerMessageCodec::encode_remove_request(
...
+    } else {
+        base2_ += increment_;
+    }
+    return *this;
+}
+
+auto_batcher::block::block(id_batch batch, std::chrono::milliseconds validity)
+  : id_batch_(batch)
+  // Mirror Java AutoBatcher.Block: validity <= 0 means "never expire".
+  , invalid_since_(validity > std::chrono::milliseconds::zero()
```
JackPGreen: The comment and the implementation are equivalent but backwards, makes it difficult to read.

hazelcast/src/hazelcast/client/proxy.cpp
```diff
@@ -1007,171 +1008,261 @@ IListImpl::ListListenerMessageCodec::encode_remove_request(
...
-        if (res != INT64_MIN) {
-            return res;
+        int64_t v = b->next();
+        if (v != INT64_MIN) {
+            return boost::make_ready_future(v);
+        }
+    }
+
+    // Slow path: elect a single fetcher; concurrent callers coalesce onto
+    // one outstanding batch request (single-flight, no thundering herd).
```
JackPGreen, line 1167: This seems like a positive, but not-directly related improvement. The PR could've been easier to review with it as a separate, more focussed PR.

JackPGreen (changes requested):

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: yes, Java uses blocking API and there is a lock all threads wait on [here](https://github.com/hazelcast/hazelcast-mono/blob/d7689cc5caee077b606c70bb6bd6217d3e0b13d0/hazelcast/hazelcast/src/main/java/com/hazelcast/flakeidgen/impl/AutoBatcher.java#L54) and the [next](https://github.com/hazelcast/hazelcast-mono/blob/d7689cc5caee077b606c70bb6bd6217d3e0b13d0/hazelcast/hazelcast/src/main/java/com/hazelcast/flakeidgen/impl/AutoBatcher.java#L55) line guarantees that only one thread does the update. Does this make sense to you?

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: fixed at ef800ff4125f39aba8506d59fb0f3475abb8e9ee how does it look?

ihsandemir: proxy.cpp line 1171 uses a [lock](https://github.com/hazelcast/hazelcast-cpp-client/pull/1454/changes#diff-e22ba4a63fce690f9f9fe7efa8fd11261b84f31f799b2fcce4f7155ab3004afaR1171). The work inside the lock is not a blocking call since we offload the actual fetch of the batch range (the remote call) see line 1194. Hence, only one thread will do the fetch and the others will use the returned shared_future from that thread's async call. Hence, all threads return a future based on this shared_future. Hence, this is non-blocking in that sense. Does this make sense?

JackPGreen: I still don't think it's clear - is that winning, as in the coalesced request?

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: why? see line prox.cpp line [1232](https://github.com/hazelcast/hazelcast-cpp-client/pull/1454/changes#diff-e22ba4a63fce690f9f9fe7efa8fd11261b84f31f799b2fcce4f7155ab3004afaR1232) . The new batch is retrieved but if there are no ids (ids are consumed by waiting calls), then a new batch needs to be retrieved. In Java for loop, it is similar such that a batch is retrieved and it tries to [get a new id from the block ](https://github.com/hazelcast/hazelcast-mono/blob/4e659148e98bad071beeae92abec3347f593f3bb/hazelcast/hazelcast/src/main/java/com/hazelcast/flakeidgen/impl/AutoBatcher.java#L50)and if it fails it goes into fetching it again.

does this make sense?

hazelcast/src/hazelcast/client/proxy.cpp
ihsandemir: i did not understand this question. Splitting the auto_batcher allows for it to use a custom `id_batch_supplier` so that we can use during the tests.

hazelcast/src/hazelcast/client/proxy.cpp
ihsandemir: fixed at bbed28b37c4b16dda62f2625e1de0f13b7ac1086

hazelcast/src/hazelcast/client/proxy.cpp
ihsandemir: Why did you think it is not related? One improvement is that we do not want multiple parallel batch requests to the server and also drop most of them but only use one of them. This eliminates it so that the server side batches are not also wasted unnecessarily.

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: winning means to me: The user thread which gets the chance to retrieve from remote the batch range. Hence, the other treads do not win but use whatever that winning thread retrieves from the server. The non-winning threads do not do any fetch from remote. What exactly is your suggestion?

JackPGreen: > proxy.cpp line 1171 uses a [lock](https://github.com/hazelcast/hazelcast-cpp-client/pull/1454/changes#diff-e22ba4a63fce690f9f9fe7efa8fd11261b84f31f799b2fcce4f7155ab3004afaR1171)

Ok, I misunderstood - it's not lock free, only on the "fast path".

I still don't understand how we can guarantee which threads aren't blocked, if were using locks. What mechanism ensures that?

hazelcast/include/hazelcast/client/impl/auto_batcher.h
JackPGreen: > does this make sense?

Sorry, no.

I understand how the implementation achieves this, but I think from an interface/user POV, the fact that it doesn't return `INT64_MIN` _because that's how it signals to retry_ is an implementation detail.

From a usage POV you just need to know what it's expected (and not expected) to return.

hazelcast/src/hazelcast/client/proxy.cpp
JackPGreen: I would've expected the batching implementation to be in the in a separate file if it's been split out to improve testability. Maybe that's my misunderstanding.

hazelcast/src/hazelcast/client/proxy.cpp
JackPGreen: Maybe "related" was the wrong word - this PR changes several behaviours of the batcher simultaneously, which ultimately leads to a more complex PR to review.

It would've been easier if it was split.

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: It does not matter which user thread triggers getting the new batch from the server. Hence, whichever wins triggers the process making the remote call. It is also off-loaded to user executor thread. The lock is to protect who does the off-loading. Once it is offloaded, the other threads getting the lock will observe that the shared_future exists and they do not need to do the remote call but work on that shared_future.

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: From user API view, they both return the same, there is no difference. Java also does not return Long.MIN_VALUE, right?

hazelcast/include/hazelcast/client/impl/auto_batcher.h
JackPGreen: Understood.

hazelcast/include/hazelcast/client/impl/auto_batcher.h
JackPGreen: Im happy with it not returning INT64_MIN. I just think the documentation could be improved.

hazelcast/src/hazelcast/client/proxy.cpp
ihsandemir: we keep the .cpp file count to minimum because the more of them the compilation times are too high. Hence, we combine related impl. into the same file. see the overall code base. You can have as many header files, no problem but the cpp is the compilation unit.

hazelcast/src/hazelcast/client/proxy.cpp
ihsandemir: item 4 could have been separated but since it was only 558 lines I thought it was not needed. but i agree the more tiny PRs the easier to review.

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: do you mean the documentation at the header file to be improved?

hazelcast/src/hazelcast/client/proxy.cpp
JackPGreen: > only 558 lines 

Still a lot in an unfamiliar language / codebase. A nit tho.

hazelcast/include/hazelcast/client/impl/auto_batcher.h
JackPGreen: Yes, just documentation.

Maybe the `INT64` return should be separate.

ihsandemir: fixed by a86bb5663fa5a74c0a8f85f4fdc38fa0ccba73cd

hazelcast/include/hazelcast/client/impl/auto_batcher.h
ihsandemir: fixed by a86bb5663fa5a74c0a8f85f4fdc38fa0ccba73cd

JackPGreen (approved):

PR #1458, opened 2026-05-21
Revert back the max boost version to 1.90.0 [HZ-5424]

ihsandemir:
Revert back the max boost version to 1.90.0 as this is the highest that exists at the latest released code base https://github.com/microsoft/vcpkg/releases/tag/2026.04.27. We only use the released vcpkg versions. The higher 1.91.0 exists at vcpkg master but it is not released yet. This caused the nightly builds (e.g. https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/26135100539/job/76868494624) fail, since we had the

```
"builtin-baseline": "56bb2411609227288b70117ead2c47585ba07713",
```
at vcpkg.json file which is 2026.04.27 release commit id and it does not have 1.91 versions of boost ports.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -24,11 +24,11 @@ on:
         description: Enter guest PR verified HEAD commit SHA.
         required: true
 env:
-  BOOST_VERSION: 1.91.0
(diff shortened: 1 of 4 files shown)


JackPGreen (approved):

ihsandemir:
I have to merge without green builds since the fix is for the workflows in verification.

PR #1464, opened 2026-05-22
Fix rfc2818_verification build failure on Windows with Boost 1.83.0 and OpenSSL 3.x [HZ-5424]

### LINKED ISSUE #1463: Windows build with Boost 1.83.0 and OpenSSL 3.x fails: undefined type 'asn1_string_st' in rfc2818_verification.ipp
## Summary

The `nightly-Windows` workflow fails to compile for the `msvc-latest_boost_1830` matrix entry when OpenSSL is enabled. Build output (from [nightly-Windows run](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/26265963834/job/77309148016)):

```
C:\Boost\include\boost-1_83\boost\asio\ssl\impl\rfc2818_verification.ipp(85,36): error C2027:
    use of undefined type 'asn1_string_st'
    [D:\a\hazelcast-cpp-client\hazelcast-cpp-client\build\hazelcast-cpp-client.vcxproj]
    (compiling source file '../hazelcast/src/hazelcast/client/client_impl.cpp')
        C:\Program Files\OpenSSL\include\openssl\types.h(57,16):
        see declaration of 'asn1_string_st'

C:\Boost\include\boost-1_83\boost\asio\ssl\impl\rfc2818_verification.ipp(85,15): error C2660:
    'memcmp': function does not take 2 arguments
    (compiling source file '../hazelcast/src/hazelcast/client/client_impl.cpp')
        while trying to match the argument list '(_Ty *, int)'
            with [_Ty=unsigned char]
```

The Boost 1.90.0 matrix entry compiles fine; only Boost 1.83.0 + OpenSSL 3.x (installed via `choco install openssl` on the Windows runner) is affected.

## Root cause

The client never references `boost::asio::ssl::rfc2818_verification` — it uses the modern `boost::asio::ssl::host_name_verification` (see `hazelcast/src/hazelcast/util/util.cpp`). However, three internal headers include the umbrella `<boost/asio/ssl.hpp>`:

- `hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h`
- `hazelcast/include/hazelcast/client/internal/socket/SSLSocket.h`
- `hazelcast/include/hazelcast/client/internal/socket/SocketFactory.h`

In Boost 1.83.0, `<boost/asio/ssl.hpp>` transitively includes `<boost/asio/ssl/rfc2818_verification.hpp>`, which in Boost.Asio's default header-only mode pulls in `rfc2818_verification.ipp`. That `.ipp` dereferences `ASN1_STRING` fields directly (`domain->data`, `domain->length`). Starting with OpenSSL 3.0, `ASN1_STRING` is an opaque type whose layout is no longer exposed by the public headers, and MSVC refuses to compile the field accesses (hence the `C2027`; the `C2660` is a follow-on because `domain->data` collapses to a non-pointer expression, making `memcmp(domain->data, ...)` look like a two-argument call).

GCC/Clang on Linux/macOS happen to accept the opaque-struct dereference more permissively in this code path, which is why only the Windows + MSVC + OpenSSL 3.x combination breaks.

Upstream Boost removed `rfc2818_verification` from the umbrella `<boost/asio/ssl.hpp>` after 1.83.0 (it is no longer present in 1.90.0), so the Boost 1.90.0 matrix entry is unaffected.

## Reproduction

1. Run the `nightly-Windows` workflow (`.github/workflows/nightly-windows.yml`).
2. Any matrix entry with `boost_folder_name: boost_1_83_0` and `with_openssl.toggle: ON` reproduces the failure when MSVC compiles a translation unit that transitively includes `<boost/asio/ssl.hpp>` (e.g. `client_impl.cpp` via `ClientConnectionManagerImpl.h` -> `SocketFactory.h`).

## Proposed fix

Replace the umbrella include with the specific subheaders the codebase actually uses (`context.hpp`, `error.hpp`, `host_name_verification.hpp`, `stream.hpp`) in the three internal headers above. This is the same approach already used in `hazelcast/src/hazelcast/util/util.cpp`. It avoids dragging in `rfc2818_verification.hpp` entirely and restores the Boost 1.83.0 + OpenSSL 3.x Windows build without affecting any other matrix entry.

Tracking: HZ-5424

ihsandemir:
Closes #1463

## Summary

- The Boost 1.83.0 + OpenSSL 3.x matrix entry of the `nightly-Windows` workflow fails to compile `rfc2818_verification.ipp`: `error C2027: use of undefined type 'asn1_string_st'` (followed by a cascading `C2660` on `memcmp`).
- The client never references `boost::asio::ssl::rfc2818_verification`; it uses the modern `host_name_verification` (`hazelcast/src/hazelcast/util/util.cpp`). The broken `.ipp` was being dragged in transitively through the umbrella `<boost/asio/ssl.hpp>` include in three internal socket headers.
- Replaced the umbrella include in `BaseSocket.h`, `SSLSocket.h`, and `SocketFactory.h` with the specific subheaders the codebase actually uses (`context.hpp`, `error.hpp`, `host_name_verification.hpp`, `stream.hpp`) — matching the include pattern already used in `util.cpp`. This avoids `rfc2818_verification.hpp` entirely.

## Root cause

In Boost 1.83.0, `<boost/asio/ssl.hpp>` lists `<boost/asio/ssl/rfc2818_verification.hpp>` among its includes. In Boost.Asio's default header-only mode that pulls in `rfc2818_verification.ipp`, whose hostname-matching code dereferences `ASN1_STRING` fields directly (`domain->data`, `domain->length`). OpenSSL 3.0 made `ASN1_STRING` an opaque type whose layout is no longer exposed by the public headers, so MSVC refuses to compile the field accesses (hence `C2027`; the `C2660` is a follow-on because `domain->data` collapses to a non-pointer expression). GCC/Clang on Linux/macOS happen to accept the opaque-struct dereference more leniently on this path, which is why only the Windows + MSVC + OpenSSL 3.x combination breaks.

Upstream Boost dropped `rfc2818_verification` from `<boost/asio/ssl.hpp>` after 1.83.0; it is no longer present in 1.90.0. That is why only the 1.83.0 matrix entry is affected.

## Test plan

- [x] Local build of the `hazelcast-cpp-client` library with `HZ_BUILD_WITH_SSL=ON` succeeds on macOS arm64 (vcpkg toolchain, Boost 1.90.0, OpenSSL 3.x).
- [x] Local build of the `client_test` binary with `HZ_BUILD_WITH_SSL=ON` succeeds (all `boost::asio::ssl::context` / `verify_peer` / `verify_callback` uses in test files compile cleanly).
- [ ] CI `nightly-Windows` workflow with `msvc-latest_boost_1830` + `with_openssl: ON` now compiles (verified by re-running the workflow on this branch).
- [ ] CI `nightly-Windows` workflow with `msvc-latest_boost_1900` + `with_openssl: ON` still compiles (no regression).
- [ ] CI `build-pr` Windows / Ubuntu / macOS matrices remain green.

Tracking: HZ-5424

### DIFF
--- a/hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
+++ b/hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
@@ -19,7 +19,15 @@
 #include <boost/format.hpp>
 #include <boost/lockfree/queue.hpp>
 #ifdef HZ_BUILD_WITH_SSL
-#include <boost/asio/ssl.hpp>
+// Avoid <boost/asio/ssl.hpp> because in Boost <= 1.83 it transitively
+// includes rfc2818_verification, whose .ipp dereferences private
+// ASN1_STRING fields that became opaque in OpenSSL 3.x and breaks the
+// MSVC build. We use the modern host_name_verification instead, so pull
+// in only the specific subheaders we actually need.
+#include <boost/asio/ssl/context.hpp>
+#include <boost/asio/ssl/error.hpp>
+#include <boost/asio/ssl/host_name_verification.hpp>
+#include <boost/asio/ssl/stream.hpp>
 #endif
 
 #include "hazelcast/client/socket.h"
--- a/hazelcast/include/hazelcast/client/internal/socket/SSLSocket.h
+++ b/hazelcast/include/hazelcast/client/internal/socket/SSLSocket.h
@@ -18,7 +18,13 @@
 #ifdef HZ_BUILD_WITH_SSL
 
 #include <boost/asio.hpp>
-#include <boost/asio/ssl.hpp>
+// See BaseSocket.h: <boost/asio/ssl.hpp> in Boost <= 1.83 transitively
+// pulls in rfc2818_verification, which fails to compile against
+// OpenSSL 3.x (opaque ASN1_STRING). Use specific subheaders instead.
+#include <boost/asio/ssl/context.hpp>
+#include <boost/asio/ssl/error.hpp>
+#include <boost/asio/ssl/host_name_verification.hpp>
+#include <boost/asio/ssl/stream.hpp>
 
 #include "hazelcast/client/internal/socket/BaseSocket.h"
 
--- a/hazelcast/include/hazelcast/client/internal/socket/SocketFactory.h
+++ b/hazelcast/include/hazelcast/client/internal/socket/SocketFactory.h
@@ -22,7 +22,13 @@
 #ifdef HZ_BUILD_WITH_SSL
 
 #include <boost/asio.hpp>
-#include <boost/asio/ssl.hpp>
+// See BaseSocket.h: <boost/asio/ssl.hpp> in Boost <= 1.83 transitively
+// pulls in rfc2818_verification, which fails to compile against
+// OpenSSL 3.x (opaque ASN1_STRING). Use specific subheaders instead.
+#include <boost/asio/ssl/context.hpp>
+#include <boost/asio/ssl/error.hpp>
+#include <boost/asio/ssl/host_name_verification.hpp>
+#include <boost/asio/ssl/stream.hpp>
 
 #endif // HZ_BUILD_WITH_SSL
 


hazelcast/include/hazelcast/client/internal/socket/BaseSocket.h
```diff
@@ -19,7 +19,15 @@
 #include <boost/format.hpp>
 #include <boost/lockfree/queue.hpp>
 #ifdef HZ_BUILD_WITH_SSL
-#include <boost/asio/ssl.hpp>
+// Avoid <boost/asio/ssl.hpp> because in Boost <= 1.83 it transitively
```
JackPGreen, line 22: Could we bump the min supported boost version instead?

ihsandemir: We already have the check at [here](https://github.com/hazelcast/hazelcast-cpp-client/blob/3144c2994a301da58144464ae9c442b95942f114/CMakeLists.txt#L82). Is this sufficient? 

We put the comment here for future maintainers just in case they try to use the header `<boost/asio/ssl.hpp>` . does this make sense?

JackPGreen: What I meant was - could we make the minimum Boost `1.84` instead?

ihsandemir: no, we can support 1.83 as well but just use a different header include and it works. No need to make the min required bumped up to 1.84. At least not needed for this specific case, does it make sense?

JackPGreen: I understand we can make it work with `1.83` as this PR demonstrates, but I wasn't sure of the implication of that (i.e. if this PR adds some negative trade off for that support). As we recently bumped the min version for other reasons, I wasn't sure if bumping one more version would be an issue.

ihsandemir: I do not thinking bumping one minor would be an issue since highest currently is 1.91.0 but still since the need here is different from the other reason why we bumped it to 1.83.0 (because we wanted to use a new capability) versus here there is an already a workaround and it is not an ugly workaround if you ask me, hence, I thought we do not bump for this reason. Do you feel strong about bumping the min version? By the way, it is only due to the windows builds also, and most users we expect to be on non-windows, e.g. linux/unix.

JackPGreen: > Do you feel strong about bumping the min version? By the way, it is only due to the windows builds also, and most users we expect to be on non-windows, e.g. linux/unix.

No, just wanted to check it was considered.

JackPGreen (approved):

PR #1471, opened 2026-06-01
Fix SqlTest.select Windows flake: move bounds-check EXPECT_THROWs outside the row loop [HZ-5424]

### LINKED ISSUE #1470: SqlTest.select flakes on Windows: ~12 000 exception throws inside the row loop cause ProcDump debugger overhead → connection timeout [API-2383]
## Symptom

`SqlTest.select` fails intermittently on **Windows only**. The [CI log](https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/26276607372/job/77342307199?pr=1464) shows a flood of lines such as

```
[08:39:05]Exception: E06D7363.?AVindex_out_of_bounds@exception@client@hazelcast@@
[08:39:05]Exception: E06D7363.?AVillegal_argument@exception@client@hazelcast@@
```

followed by the Java server reporting `Client cannot be reached` for the SQL cursor.

### Windows C++ exceptions are SEH events

On Windows, every C++ `throw` is implemented as an SEH (Structured Exception Handling) exception with code **`0xE06D7363`** (the Windows magic constant for C++ exceptions, encoding "msc" in ASCII). The exception record also carries the mangled RTTI type name (`?AVindex_out_of_bounds@...`).

### ProcDump attaches as a debugger

The CI Windows test step runs:

```powershell
procdump.exe -accepteula -e -ma -w client_test.exe crash.dmp
```

The `-w` flag causes ProcDump to call `DebugActiveProcess()`, making it the **Windows debugger** of the test process. A Windows debugger receives **all** exception events via `WaitForDebugEvent()`, including *first-chance* notifications (exceptions that haven't been handled yet). ProcDump reads the SEH record, prints the `[HH:MM:SS]Exception: E06D7363.?AV...` line to stdout, then calls `ContinueDebugEvent(DBG_EXCEPTION_NOT_HANDLED)` so execution can continue. The C++ runtime's SEH filter then finds the `catch` block and handles the exception normally.

### The cost of each throw under a debugger

Without a debugger, `throw` → `catch` is entirely in-process — zero kernel transitions.  
With ProcDump attached, each throw requires:

1. The test process's threads are **suspended** by the kernel.  
2. An `EXCEPTION_DEBUG_EVENT` is delivered to ProcDump across the kernel debug port (inter-process IPC).  
3. ProcDump wakes from `WaitForDebugEvent()`, prints the line, calls `ContinueDebugEvent()`.  
4. The kernel resumes the test process.  
5. The C++ runtime finds the `catch` handler.

That is **two mandatory kernel-mode round-trips per throw**.

### Why it breaks this test

The `select` test iterates over **4 096 SQL rows**. Inside the loop, three `EXPECT_THROW` calls verify that `get_object` throws on an invalid negative index, an out-of-range index, and an unknown column name:

```cpp
for (auto itr = res->iterator(); itr.has_next();) {
    auto page = itr.next().get();
    for (const auto& row : page->rows()) {
        // ... value checks + 4096 map->get() RPC calls ...
        EXPECT_THROW(row.get_object<int>(-1), exception::index_out_of_bounds);
        EXPECT_THROW(row.get_object<int>(row.row_metadata().column_count()), exception::index_out_of_bounds);
        EXPECT_THROW(row.get_object<int>("unknown_field"), exception::illegal_argument);
    }
}
```

3 throws × 4 096 rows = **12 288 exception round-trips through ProcDump** = 24 576 kernel transitions. Combined with the 4 096 synchronous `map->get()` server round-trips, the loop runs slow enough that the server-side Hazelcast connection health check fires and the SQL cursor is cancelled, causing `Client cannot be reached` and the test to fail.

The `Client cannot be reached` messages for *other* UUIDs in the log are background noise — the `QueryStateRegistryUpdater` cleaning up cursors from already-shutdown clients of earlier tests.

## Fix

The three `EXPECT_THROW` calls verify fixed API behaviour that does not change row-to-row. They only need to run **once**, on a single representative row, before the main iteration loop. Fetching the first page ahead of the loop and running the checks there reduces the exception count from 12 288 to 3.

See the accompanying PR for the one-file change to `hazelcast/test/src/sql_test.cpp`.

ihsandemir:
Fixes #1470

## Summary

- `SqlTest.select` was iterating over 4 096 SQL rows and calling `EXPECT_THROW` three times per row to verify `get_object()` raises `index_out_of_bounds` / `illegal_argument`.  That is **~12 000 C++ exception throws** per test run.
- On Windows the CI step runs ProcDump as a debugger (`procdump -e -ma -w client_test.exe`).  Every C++ `throw` raises SEH code `0xE06D7363`; the OS delivers this as a *first-chance debug event* to ProcDump **before** the C++ `catch` handler can run.  ProcDump logs `[HH:MM:SS]Exception: E06D7363.?AVindex_out_of_bounds@...`, then calls `ContinueDebugEvent()` — **two mandatory kernel-mode round-trips per throw**.  12 000 throws × 2 = 24 000 kernel transitions, combined with 4 096 synchronous `map->get()` server calls, made the loop slow enough for the server-side connection health-check to fire and cancel the SQL cursor (`Client cannot be reached`).
- Fix: fetch the first page before the main loop, run the three `EXPECT_THROW` checks on its first row only, then iterate all pages in a `for(;;)` loop starting from the already-fetched first page.  Exception count drops from **12 000 to 3**.  The `bounds_checked` flag is eliminated.

## Why it was Windows-only

Linux/macOS CI does not use ProcDump.  Without a debugger attached, `throw` → `catch` is entirely in-process with zero kernel transitions, so the loop completes well within any timeout.

## Root cause detail

On Windows, C++ exceptions are built on SEH.  When a debugger is attached via `DebugActiveProcess()`, the kernel suspends the target process's threads and delivers an `EXCEPTION_DEBUG_EVENT` to the debugger through the kernel debug port.  The debugger must call `WaitForDebugEvent()`, process the event, and call `ContinueDebugEvent()` before the target resumes.  This happens for *every* throw, even ones that are immediately caught — ProcDump cannot skip them.  Under heavy exception load (12 000 throws), the cumulative stall is enough to trigger server-side timeouts.

## Test plan

- [x] The three `EXPECT_THROW` boundary checks still execute (on the first row of the first page).
- [x] All 4 096 rows are still visited and verified through the restructured `for(;;)` loop.
- [ ] CI `build-pr` Windows jobs pass without the `SqlTest.select` failure.

### DIFF
--- a/hazelcast/test/src/sql_test.cpp
+++ b/hazelcast/test/src/sql_test.cpp
@@ -1870,8 +1870,34 @@ TEST_F(SqlTest, select)
 
     std::unordered_set<int64_t> unique_keys;
 
-    for (auto itr = res->iterator(); itr.has_next();) {
-        auto page = itr.next().get();
+    // Fetch the first page up front so we can verify the bounds-checking API
+    // on a single representative row before the main loop.
+    // Background: on Windows, C++ exceptions are built on SEH. When ProcDump
+    // is attached as a debugger (CI step: procdump -e -ma -w client_test.exe),
+    // every throw raises SEH code 0xE06D7363, which the OS delivers as a
+    // first-chance debug event to ProcDump before the C++ catch handler runs.
+    // ProcDump logs the line "[HH:MM:SS]Exception: E06D7363.?AV..." and then
+    // calls ContinueDebugEvent so execution can proceed. This is a mandatory
+    // two-kernel-mode-transition round-trip for every single throw.
+    // With the EXPECT_THROW calls inside the 4096-row loop that is ~12 000
+    // round-trips, stalling the loop long enough to trigger connection
+    // timeouts on the server side (Windows-only flaky failure).
+    auto itr = res->iterator();
+    ASSERT_TRUE(itr.has_next());
+    auto page = itr.next().get();
+    ASSERT_FALSE(page->rows().empty());
+    {
+        const auto& first_row = page->rows().front();
+        EXPECT_THROW(first_row.get_object<int>(-1),
+                     exception::index_out_of_bounds);
+        EXPECT_THROW(
+          first_row.get_object<int>(first_row.row_metadata().column_count()),
+          exception::index_out_of_bounds);
+        EXPECT_THROW(first_row.get_object<int>("unknown_field"),
+                     exception::illegal_argument);
+    }
+
+    for (;;) {
         for (const auto& row : page->rows()) {
             ASSERT_EQ(row_metadata, res->row_metadata());
 
@@ -1903,14 +1929,10 @@ TEST_F(SqlTest, select)
               sql_column_type::varchar, val.varchar_val, row, "varcharVal");
 
             unique_keys.emplace(*key0);
-
-            EXPECT_THROW(row.get_object<int>(-1),
-                         exception::index_out_of_bounds);
-            EXPECT_THROW(row.get_object<int>(row.row_metadata().column_count()),
-                         exception::index_out_of_bounds);
-            EXPECT_THROW(row.get_object<int>("unknown_field"),
-                         exception::illegal_argument);
         }
+        if (!itr.has_next())
+            break;
+        page = itr.next().get();
     }
 
     EXPECT_THROW(res->iterator(), exception::illegal_state);


dmitriyrazboev (approved):

hazelcast/test/src/sql_test.cpp
```diff
@@ -1870,8 +1870,34 @@ TEST_F(SqlTest, select)
...
+    // Background: on Windows, C++ exceptions are built on SEH. When ProcDump
+    // is attached as a debugger (CI step: procdump -e -ma -w client_test.exe),
+    // every throw raises SEH code 0xE06D7363, which the OS delivers as a
+    // first-chance debug event to ProcDump before the C++ catch handler runs.
+    // ProcDump logs the line "[HH:MM:SS]Exception: E06D7363.?AV..." and then
+    // calls ContinueDebugEvent so execution can proceed. This is a mandatory
+    // two-kernel-mode-transition round-trip for every single throw.
+    // With the EXPECT_THROW calls inside the 4096-row loop that is ~12 000
+    // round-trips, stalling the loop long enough to trigger connection
+    // timeouts on the server side (Windows-only flaky failure).
```
JackPGreen, line 1884: Nit - this is very verbose and could probably be simplified.

JackPGreen (approved):

PR #1475, opened 2026-06-04
Do not warn on benign double completion of backup-aware invocations [HZ-5459]

### LINKED ISSUE #1474: Spurious "Failed to set the response for invocation" WARNING logs for backup-aware operations (5.7.0) [HZ-5459]
### Summary

Since 5.7.0, a stable C++ client connected to a multi-member cluster continuously emits `WARNING` log lines like:

```
WARNING: [...] hz.client_1[dev] [5.7.0] Failed to set the response for invocation. Dropping the response. The state of the promise has already been set., ClientInvocation{ requestMessage = ClientMessage{... operation=map.put ...}, ... backup_acks_expected_ = 1, backup_acks_received = 1, pending_response: 0x... } Response: ClientMessage{...}
```

These did not appear with the 5.6.0 client. They are emitted while the client is healthy and under steady load — not only during shutdown — and can flood the logs (tens of thousands of lines in a single soak run).

Reproduced with the [soak test](https://github.com/hazelcast/hazelcast-cpp-client/tree/master/examples/soak-test) against a multi-member cluster. In one 10-client run we observed 47,664 such warnings, every one of them for a backup-aware operation (`map.put` / `map.executeonkey`) with `backup_acks_expected_ == backup_acks_received`, and **zero** real failures (no timeouts, no indeterminate-state, no lost responses).

### Root cause

For a backup-aware operation the invocation completes only once both the primary response and the backup ack(s) have arrived. Two paths can deliver that final completion:

- `ClientInvocation::notify_response()` — the primary response
- `ClientInvocation::notify_backup()` — the last backup ack

The handoff is coordinated through `pending_response_` and the ack counters so that only one path calls `complete()`. There is, however, an inherent race window (expected backups = 1):

1. `notify_response`: `expected(1) > received(0)` → stores `pending_response_`
2. `notify_backup` (another thread): `++received` → 1; sees `pending_response_` set; `expected == received` → `complete()` → promise set
3. `notify_response` resumes its recheck: `received(1) == expected(1)` → falls through → `complete()` again → `boost::promise_already_satisfied` → logged as `WARNING`

Both calls complete with the *same* response, so the future is completed exactly once with the correct value and the duplicate is harmlessly dropped. The only defect is the noisy log line.

This matches the Java client logic exactly (`BaseInvocation.notifyResponse` has the same double-check). The difference is purely in reporting:

- Java's `complete()` → `AbstractInvocationFuture.complete0()` → `warnIfSuspiciousDoubleCompletion()` only warns when the already-set value differs from the offered one (`s0 != s1`). The backup-ack race always completes with the same object, so Java stays silent.
- The C++ `complete()` catches `boost::promise_already_satisfied` and logs **every** occurrence at `WARNING`.

Notably the C++ `set_exception()` path already treats `boost::promise_already_satisfied` as benign and logs it at `finest`; only the value path (`complete()`) is loud.

### Why it is new in 5.7.0

The multi-threaded pipeline rearchitecture (#1412, HZ-5387: IO thread pool / response threads) made the primary response and the backup ack genuinely run on different threads, so the race is now hit constantly. It also introduced the `notify_response` recheck (correctly mirroring Java) to fix a lost-wakeup hang; that recheck is what makes both paths reach `complete()`. In 5.6.0 the per-connection processing was effectively serialized and the old `notify()` returned unconditionally after staging the pending response, so a double completion could not occur.

### Impact

Cosmetic / log-noise only. No operation fails, no response is lost, the user always receives the correct result. But the spurious `WARNING` volume is alarming to operators and masks genuine warnings.

### Proposed fix

Make `complete()` mirror Java's `warnIfSuspiciousDoubleCompletion`: only warn when the offered response differs from the canonical `pending_response_`. A double completion with the same response (the benign backup-ack race) is logged at `finest`; a genuinely different value still warns. This also aligns the value path with how `set_exception()` already handles the same condition.

ihsandemir:
Fixes #1474

### Problem

Since 5.7.0 a healthy C++ client connected to a multi-member cluster continuously emits `WARNING` lines like:

```
Failed to set the response for invocation. Dropping the response. The state of the promise has already been set., ClientInvocation{ ... operation=map.put ... backup_acks_expected_ = 1, backup_acks_received = 1, ... }
```

A 10-client soak run produced 47,664 of these, all for backup-aware operations with `backup_acks_expected == backup_acks_received`, with no actual failures.

### Root cause

For a backup-aware operation the invocation completes once the primary response and the backup ack(s) have arrived. Both `notify_response()` and `notify_backup()` can win the final completion, and there is an inherent race in which both reach `complete()` with the **same** pending response:

1. `notify_response` stages `pending_response_`
2. `notify_backup` (other thread) increments the ack count and completes the promise
3. `notify_response`'s recheck now sees `received == expected`, falls through and calls `complete()` again → `boost::promise_already_satisfied`

The future is completed exactly once with the correct value; the duplicate is harmlessly dropped. Only the log line is wrong. The race became routine after the multi-threaded pipeline work (#1412), which made the response and the backup ack run on different threads.

This mirrors the Java client (`BaseInvocation.notifyResponse` has the same double-check). The difference is reporting: Java's `AbstractInvocationFuture.warnIfSuspiciousDoubleCompletion` only warns when the already-set value differs from the offered one, so the same-value backup-ack race is silent. The C++ `complete()` warned on every `promise_already_satisfied`.

### Fix

`complete()` now mirrors Java: a duplicate completion with the same response (the benign backup-ack race) is logged at `finest`, while a genuinely conflicting value still warns. This also aligns the value path with how `set_exception()` already handles the same condition.

### Tests

Added `client_invocation_test` with two cases driving the private completion path via friend access:

- benign double completion with the same response → no `WARNING`, one `finest`
- conflicting double completion with a different response → one `WARNING`

Verified red/green: the benign test fails before the fix (logs at `WARNING`) and passes after. Both tests pass against a running cluster.

### DIFF
--- a/hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
+++ b/hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
@@ -41,7 +41,7 @@ namespace client {
 class address;
 
 namespace test {
-// White-box test fixture; granted access to private completion internals to
+// Exposed for testing only; granted access to private completion internals to
 // deterministically reproduce the response/backup-ack completion race.
 class client_invocation_test;
 } // namespace test
--- a/hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
+++ b/hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
@@ -40,6 +40,12 @@ class logger;
 namespace client {
 class address;
 
+namespace test {
+// White-box test fixture; granted access to private completion internals to
+// deterministically reproduce the response/backup-ack completion race.
+class client_invocation_test;
+} // namespace test
+
 namespace connection {
 class Connection;
 }
@@ -139,6 +145,8 @@ public:
 
     friend std::ostream& operator<<(std::ostream& os,
                                     const ClientInvocation& invocation);
+
+    friend class ::hazelcast::client::test::client_invocation_test;
 
     boost::promise<protocol::ClientMessage>& get_promise();
 
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -2119,6 +2119,30 @@ ClientInvocation::complete(const std::shared_ptr<protocol::ClientMessage>& msg,
     try {
         // TODO: move msg content here?
         this->invocation_promise_.set_value(*msg);
+    } catch (boost::promise_already_satisfied& e) {
+        // The promise is already satisfied. This is an expected, benign race
+        // between notify_response() (primary response) and notify_backup() (the
+        // last backup ack): for a backup-aware operation both paths can reach
+        // complete() with the very same pending_response_. The first one wins
+        // and the duplicate is harmlessly dropped.
+        //
+        // This mirrors Java's AbstractInvocationFuture.warnIfSuspiciousDouble-
+        // Completion, which only warns when the already-set value differs from
+        // the offered one. Here the canonical value is pending_response_, so a
+        // completion with the same message is benign (logged at finest) while a
+        // genuinely different value remains a warning.
+        auto pending = pending_response_.load();
+        auto level = (!pending || *pending != msg)
+                       ? ::hazelcast::logger::level::warning
+                       : ::hazelcast::logger::level::finest;
+        if (logger_.enabled(level)) {
+            logger_.log(
+              level,
+              boost::str(
+                boost::format("Failed to set the response for invocation. "
+                              "Dropping the response. %1%, %2% Response: %3%") %
+                e.what() % *this % *msg));
+        }
     } catch (std::exception& e) {
         HZ_LOG(logger_,
                warning,
--- a/hazelcast/test/src/HazelcastTests5.cpp
+++ b/hazelcast/test/src/HazelcastTests5.cpp
@@ -13,15 +13,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#include <algorithm>
 #include <cmath>
 #include <fstream>
 #include <memory>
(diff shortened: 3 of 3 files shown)


dmitriyrazboev (dismissed):

hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
```diff
@@ -40,6 +40,12 @@ class logger;
 namespace client {
 class address;
 
+namespace test {
+// White-box test fixture; granted access to private completion internals to
```
JackPGreen: Can we use "exposed for testing" or some other terminology in use elsewhere in the codebase?

hazelcast/test/src/client_invocation_test.cpp
```diff
@@ -0,0 +1,185 @@
+/*
+ * Copyright (c) 2008-2025, Hazelcast, Inc. All Rights Reserved.
```
JackPGreen:
```suggestion
 * Copyright (c) 2008-2026, Hazelcast, Inc. All Rights Reserved.
```

Not sure how were managing this.

hazelcast/include/hazelcast/client/spi/impl/ClientInvocation.h
ihsandemir: fixed at 88e26cb4938adb4b1f4c37eac20a56c67994901b

hazelcast/test/src/client_invocation_test.cpp
ihsandemir: moved this file content into existing HazelcastTests5.cpp file.

JackPGreen (approved):

PR #1487, opened 2026-09-02
Make the schema replication executor pool size configurable

ctc-louk:
## Problem

`ClientExecutionServiceImpl::start()` created the schema replication thread pool with the no-argument constructor:

```cpp
schema_replication_executor_.reset(new hazelcast::util::hz_thread_pool());
```

`hz_thread_pool()` delegates to `boost::asio::thread_pool()`, which sizes itself at `hardware_concurrency() * 2` — 192 threads on a 96-core host. The pool's only consumer is `ClientInvocation::replicate_schemas`, which performs occasional compact schema replication, and a given schema is replicated at most once per client. The pool was therefore far larger than the workload warrants, sat idle in steady state, and there was no way to change it.

## Change

Adds a `client_properties` entry:

| | |
|---|---|
| Name | `hazelcast.client.schema.replication.pool.size` |
| Default | `3` |
| Resolution | `client_config::set_property()` > environment variable > default |

`start()` resolves it exactly as it already resolves the internal executor pool size, clamping a non-positive value to the default.

```cpp
hazelcast::client::client_config config;
config.set_property("hazelcast.client.schema.replication.pool.size", "2");

auto client = hazelcast::new_client(std::move(config)).get();
```

This follows every other pool knob in the client — `IO_THREAD_COUNT`, `INTERNAL_EXECUTOR_POOL_SIZE`, `RESPONSE_THREAD_COUNT`, `EVENT_THREAD_COUNT` — all of which are properties with small constant defaults. No public API is added, `ClientExecutionServiceImpl`'s constructor is unchanged, and nothing is added to `client_config`, so there is no ABI impact.

The default of `3` replaces `2 * hardware_concurrency` for **every** user, not only those who discover the knob. Tasks on this pool block — `replicate_schema_in_cluster` waits on `invoke().get()` and sleeps for the invocation retry pause on the retry path — so the property documents that a value of 1 makes concurrent writes of distinct unseen schemas serialize behind one another.

## Testing

- `schema_replication_pool_size_test` — offline; asserts the property resolves to the default when unset and that a configured value overrides it.
- `ThreadPoolTest.testSchemaReplicationPoolSizeIsHonoured` — added to the existing `ThreadPoolTest` fixture, so it reuses that suite's member and adds no second `HazelcastServer`. Each parameter configures the pool to `GetParam()` and submits `2 * GetParam()` jobs, asserting *exactly* `GetParam()` distinct thread ids. Every parameter differs from the default of 3, so each case fails if the property is ignored. The test uses a local `hazelcast_client` rather than the fixture's static raw pointer.

### DIFF
--- a/hazelcast/src/hazelcast/client/spi.cpp
+++ b/hazelcast/src/hazelcast/client/spi.cpp
@@ -1629,7 +1629,15 @@ ClientExecutionServiceImpl::start()
           new hazelcast::util::hz_thread_pool(user_pool_size_));
     }
 
-    schema_replication_executor_.reset(new hazelcast::util::hz_thread_pool());
+    int schemaReplicationPoolSize = client_properties_.get_integer(
+      client_properties_.get_schema_replication_pool_size());
+    if (schemaReplicationPoolSize <= 0) {
+        schemaReplicationPoolSize = util::IOUtil::to_value<int>(
+          client_properties::SCHEMA_REPLICATION_POOL_SIZE_DEFAULT);
+    }
+
+    schema_replication_executor_.reset(
+      new hazelcast::util::hz_thread_pool(schemaReplicationPoolSize));
 }
 
 void
--- a/hazelcast/test/src/HazelcastTests8.cpp
+++ b/hazelcast/test/src/HazelcastTests8.cpp
@@ -2767,9 +2767,66 @@ TEST_P(ThreadPoolTest, testEqualThreadAndJobs)
     ASSERT_EQ(expected_thread_num, state->thread_ids.size());
 }
 
+// Submits twice as many jobs as the configured pool size, so the assertion is
+// an exact thread count rather than min(jobs, size). Each parameter differs
+// from SCHEMA_REPLICATION_POOL_SIZE_DEFAULT, so every case fails if the
+// property is ignored.
+TEST_P(ThreadPoolTest, testSchemaReplicationPoolSizeIsHonoured)
+{
+    int32_t num_of_thread = GetParam();
+    int32_t num_of_jobs = num_of_thread * 2;
+
+    client_config config;
+    config.set_property(client_properties::SCHEMA_REPLICATION_POOL_SIZE,
+                        std::to_string(num_of_thread));
+
+    auto local_client = new_client(std::move(config)).get();
+
+    spi::ClientContext ctx(local_client);
+    auto state = std::make_shared<ThreadState>(num_of_jobs);
+    std::mutex mutex_for_thread_id;
+    boost::barrier sync_barrier(num_of_thread);
+
+    ASSERT_EQ(0, state->thread_ids.size());
+    for (int i = 0; i < num_of_jobs; i++) {
+        ctx.get_client_execution_service()
+          .get_schema_replication_executor()
+          .submit([state, &mutex_for_thread_id, &sync_barrier]() {
+              sync_barrier.count_down_and_wait();
+              auto curr_thread_id = boost::this_thread::get_id();
+              {
+                  std::lock_guard<std::mutex> lg(mutex_for_thread_id);
+                  state->thread_ids.insert(curr_thread_id);
+              }
+              state->latch1.count_down();
+          });
+    }
+    ASSERT_OPEN_EVENTUALLY(state->latch1);
+    ASSERT_EQ(static_cast<size_t>(num_of_thread), state->thread_ids.size());
+
+    local_client.shutdown().get();
+}
+
 INSTANTIATE_TEST_SUITE_P(ThreadPoolTestSuite,
                          ThreadPoolTest,
                          ::testing::Values(5, 10, 2));
+
+TEST(schema_replication_pool_size_test, resolves_to_default_when_unset)
+{
+    client_properties properties({});
+
+    ASSERT_EQ(
+      3, properties.get_integer(properties.get_schema_replication_pool_size()));
+}
+
+TEST(schema_replication_pool_size_test, configured_value_overrides_default)
+{
+    client_properties properties(
+      { { client_properties::SCHEMA_REPLICATION_POOL_SIZE, "7" } });
+
+    ASSERT_EQ(
+      7, properties.get_integer(properties.get_schema_replication_pool_size()));
+}
 
 } // namespace thread_pool
 } // namespace test
--- a/hazelcast/include/hazelcast/client/client_properties.h
+++ b/hazelcast/include/hazelcast/client/client_properties.h
@@ -82,6 +82,8 @@ public:
 
     const client_property& get_response_thread_count() const;
 
+    const client_property& get_schema_replication_pool_size() const;
+
     const client_property& get_shuffle_member_list() const;
 
     const client_property& get_max_concurrent_invocations() const;
@@ -215,6 +217,20 @@ public:
      */
     static const std::string RESPONSE_THREAD_COUNT;
     static const std::string RESPONSE_THREAD_COUNT_DEFAULT;
+
+    /**
+     * Number of threads which replicate compact schemas to the cluster.
+     * <p/>
+     * Every task on this pool blocks: replicate_schema_in_cluster() waits on
+     * invoke().get(), and on the retry path it sleeps for the invocation retry
+     * pause between attempts. A schema is replicated at most once per client,
+     * so this pool is idle in steady state and only a handful of threads are
+     * needed; sizing it any larger simply reserves threads that never run.
+     * Because the tasks block, a value of 1 makes concurrent writes of
+     * distinct unseen schemas serialize behind one another.
+     */
+    static const std::string SCHEMA_REPLICATION_POOL_SIZE;
+    static const std::string SCHEMA_REPLICATION_POOL_SIZE_DEFAULT;
 
     /**
      * Client shuffles the given member list to prevent all clients to connect
@@ -352,6 +368,7 @@ private:
     client_property internal_executor_pool_size_;
     client_property io_thread_count_;
     client_property response_thread_count_;
+    client_property schema_replication_pool_size_;
     client_property shuffle_member_list_;
     client_property max_concurrent_invocations_;
     client_property backpressure_backoff_timeout_millis_;
--- a/hazelcast/src/hazelcast/client/client_impl.cpp
+++ b/hazelcast/src/hazelcast/client/client_impl.cpp
@@ -998,6 +998,10 @@ const std::string client_properties::RESPONSE_THREAD_COUNT =
   "hazelcast.client.response.thread.count";
 const std::string client_properties::RESPONSE_THREAD_COUNT_DEFAULT = "2";
 
+const std::string client_properties::SCHEMA_REPLICATION_POOL_SIZE =
+  "hazelcast.client.schema.replication.pool.size";
+const std::string client_properties::SCHEMA_REPLICATION_POOL_SIZE_DEFAULT = "3";
+
 const std::string client_properties::SHUFFLE_MEMBER_LIST =
   "hazelcast.client.shuffle.member.list";
 const std::string client_properties::SHUFFLE_MEMBER_LIST_DEFAULT = "true";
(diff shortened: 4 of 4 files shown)


hazelcast/src/hazelcast/client/config.cpp
```diff
@@ -1280,6 +1281,19 @@ client_config::set_executor_pool_size(int32_t executor_pool_size)
...
 }
 
+int32_t
+client_config::get_schema_replication_pool_size() const
+{
+    return schema_replication_pool_size_;
+}
+
+void
+client_config::set_schema_replication_pool_size(
```
ihsandemir: This stores any `int32_t` unvalidated. Only `<= 0` is special-cased later in `start()`, so a value like 100000 reaches `boost::asio::thread_pool(n)` during client construction. If `create_thread` fails partway, `~thread_group` joins the already-started workers blocked in `run()` with no `stop()`, so startup hangs or throws a raw `std::system_error` instead of `exception::invalid_configuration`.

`set_executor_pool_size` and friends already go through `util::Preconditions::check_positive`; one call here closes the gap.

hazelcast/include/hazelcast/client/client_config.h
```diff
@@ -351,6 +351,22 @@ class HAZELCAST_API client_config
...
+    /**
+     * Pool size for the executor which replicates compact schemas to the
+     * cluster.
+     *
+     * \return int Schema replication pool size, or -1 when left unset.
+     */
+    int32_t get_schema_replication_pool_size() const;
+
+    /**
+     * Sets the pool size for the schema replication executor. A value of 0 or
```
ihsandemir: Worth stating the cost of a small value here. Every task on this pool blocks: `replicate_schema_in_cluster` calls `invoke().get()` and, in the retry branch, sleeps `retry_pause_millis_` (1000 ms default) up to `max_put_retry_count_` (100) times while holding the thread. The `.then(boost::launch::sync)` in `ClientInvocation::invoke` then runs `invoke_on_selection()` on that same thread. With size 1, two threads writing distinct unseen compact schemas serialize on each other, and in a split-brain case unrelated invocations can stall for up to ~100 s. Either document that or enforce a floor in the setter.

hazelcast/include/hazelcast/client/client_config.h
```diff
@@ -351,6 +351,22 @@ class HAZELCAST_API client_config
...
+     * less leaves the pool at its default size.
+     *
+     * \param schema_replication_pool_size pool size
+     */
+    void set_schema_replication_pool_size(int32_t schema_replication_pool_size);
```
ihsandemir: See the top-level comment: I would prefer a `client_properties` entry over a typed setter here. All sibling pool sizes are properties with small constant defaults, `ClientExecutionServiceImpl` already holds `client_properties_`, and this avoids a permanent public API and the constructor signature change. It also lets us pick a sane default (1-3) in the same PR rather than leaving `2 * hardware_concurrency` in place.

hazelcast/include/hazelcast/client/spi/impl/ClientExecutionServiceImpl.h
```diff
@@ -38,6 +38,7 @@ class HAZELCAST_API ClientExecutionServiceImpl
     ClientExecutionServiceImpl(const std::string& name,
                                const client_properties& properties,
                                int32_t user_pool_size,
+                               int32_t schema_replication_pool_size,
```
ihsandemir: This is a `HAZELCAST_API` exported class and the new parameter is inserted mid-signature with no default, so the old 4-argument constructor disappears in a minor release. Two adjacent `int32_t` parameters also mean a transposed `get_executor_pool_size()` / `get_schema_replication_pool_size()` at the call site in `client_impl.cpp` compiles silently and mis-sizes both pools.

If the setter stays, appending `int32_t schema_replication_pool_size = -1` as the last parameter keeps source compatibility for free. If this becomes a property, the constructor does not need to change at all.

hazelcast/src/hazelcast/client/spi.cpp
```diff
@@ -1629,7 +1631,13 @@ ClientExecutionServiceImpl::start()
           new hazelcast::util::hz_thread_pool(user_pool_size_));
     }
 
-    schema_replication_executor_.reset(new hazelcast::util::hz_thread_pool());
+    if (schema_replication_pool_size_ <= 0) {
```
ihsandemir: This is now the third copy of the `if (size <= 0) reset(new hz_thread_pool()) else reset(new hz_thread_pool(size))` block in `start()`. The sentinel policy belongs in `hz_thread_pool`, which already owns the default/explicit constructor split. A single `hz_thread_pool(int32_t n)` that treats `n <= 0` as default collapses each site to one line and removes the negative-to-`size_t` footgun for any caller that forgets the guard.

hazelcast/test/src/HazelcastTests7.cpp
```diff
@@ -1571,6 +1571,22 @@ TEST(AwsConfigTest, testInvalidAwsMemberPortConfig)
                  exception::invalid_configuration);
 }
 
+TEST(schema_replication_pool_size_test, defaults_to_unset)
+{
+    client_config config;
+
+    ASSERT_EQ(-1, config.get_schema_replication_pool_size());
```
ihsandemir: This freezes the `-1` sentinel into the public contract. We now have four descriptions of one state that already disagree: the field comment says `-1 means unset`, the setter doc says `0 or less`, the consumer checks `<= 0`, and this test asserts exactly `-1`. The resolved default (`2 * hardware_concurrency`) is stated nowhere. Moving to a concrete default later becomes an observable API and test change.

Also, these plain `TEST()`s sit between the AWS tests. Every other `client_config` accessor is tested in the `ClientConfigTest` fixture in `HazelcastTests2.cpp`; they belong there.

hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -2771,6 +2771,84 @@ INSTANTIATE_TEST_SUITE_P(ThreadPoolTestSuite,
                          ThreadPoolTest,
                          ::testing::Values(5, 10, 2));
 
+class SchemaReplicationThreadPoolTest
```
ihsandemir: This fixture is a verbatim copy of `ThreadPoolTest` directly above, differing only in the config setter and the executor getter, and it boots a second `HazelcastServer`. The two copies of `ThreadState`, `SetUpTestCase`/`TearDownTestCase`, the static server/client and the barrier/latch body will drift independently, and CI pays an extra member start plus three more client connects.

Parametrizing the existing fixture over `{setter, executor getter}` (or adding a second `TEST_P` to `ThreadPoolTest`) reuses the server and keeps one harness to maintain.

hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -2771,6 +2771,84 @@ INSTANTIATE_TEST_SUITE_P(ThreadPoolTestSuite,
...
+{
+    int32_t num_of_thread = 5;
+    int32_t num_of_jobs = GetParam();
+    client_config config;
+    config.set_schema_replication_pool_size(num_of_thread);
+
+    if (client != nullptr) {
+        client->shutdown().get();
+    }
+    client = new hazelcast_client{ new_client(std::move(config)).get() };
```
ihsandemir: `client` is a static raw pointer. On the second and third parametrized runs the previous object is shut down and then the pointer is overwritten with a fresh `new` without a `delete`, so only the last of the three clients is freed in `TearDownTestCase`. `shutdown()` stops threads and connections but the `hazelcast_client_instance_impl` graph it owns stays allocated until the wrapper is destroyed, so two full impl graphs leak per suite run.

The pattern is inherited from `ThreadPoolTest`, so this PR doubles the count rather than introducing it. CI only misses it because `scripts/test-unix.sh` sets `ASAN_OPTIONS=detect_leaks=0`. A local `auto client = new_client(std::move(config)).get();` in the test body removes the static, the teardown bookkeeping and the leak in one go.

hazelcast/test/src/HazelcastTests8.cpp
```diff
@@ -2771,6 +2771,84 @@ INSTANTIATE_TEST_SUITE_P(ThreadPoolTestSuite,
...
+                  state->thread_ids.insert(curr_thread_id);
+              }
+              state->latch1.count_down();
+          });
+    }
+    ASSERT_OPEN_EVENTUALLY(state->latch1);
+    ASSERT_EQ(expected_thread_num, state->thread_ids.size());
+}
+
+INSTANTIATE_TEST_SUITE_P(SchemaReplicationThreadPoolTestSuite,
```
ihsandemir: With `num_of_thread` hardcoded to 5, the `jobs = 5` and `jobs = 2` cases assert `min(jobs, 5)` distinct threads, which any pool of size `>= jobs` satisfies, including the unconfigured default of `2 * hardware_concurrency`. If `set_schema_replication_pool_size` were silently ignored, the exact regression this test exists to catch, those two cases would still pass on any multi-core machine. Only `jobs = 10` discriminates, and each vacuous case costs a full client connect. One case with `jobs > threads` asserting exactly 5 distinct thread ids covers the change.

ihsandemir (commented):
Thanks for the PR. A few points, roughly in priority order.

**Commit message**: the commit `d95f43e365` carries a `Co-Authored-By: Claude ...` trailer. Please amend and force-push before this is merged; the PR body itself is fine.

**Mechanism**: the Java client has no schema replication executor at all. `ClientSchemaService.put` replicates synchronously on the calling thread via `invoke().joinInternal()`, so there is nothing there to mirror. Every other pool knob in this client (`IO_THREAD_COUNT`, `INTERNAL_EXECUTOR_POOL_SIZE`, `RESPONSE_THREAD_COUNT`, `EVENT_THREAD_COUNT`) is a `client_properties` entry with a small constant default. A `hazelcast.client.schema.replication.pool.size` property with a default of 1-3 threads would need no constructor change, no `client_config` member, no ABI change, and would fix the `2 * hardware_concurrency` default for every user instead of only those who discover the setter. A typed setter is public API we cannot remove during 5.x.

For what it is worth, I looked at whether the pool could be dropped entirely in favour of `user_executor_` or `internal_executor_`. Neither is safe: `complete_call_id_sequence()` runs every invocation's completion continuation on `user_executor_`, and invocation retries are posted to `internal_executor_` via `ClientInvocation::execute()`. `replicate_schema_in_cluster` blocks on `invoke().get()`, so running it on either pool can deadlock once the pool is full. The dedicated pool is justified; it just wants a small default.

Inline comments follow for the rest.

ctc-louk:
We (CTC) will need this feature sooner than the major version (6.x), ideally a minor version. I pivoted to the recommended approach of adding hazelcast.client.schema.replication.pool.size property

hazelcast/src/hazelcast/client/config.cpp
ctc-louk: Thanks for raising the issue of setting unrealistic large number of threads. The new approach follows existing pattern now.

hazelcast/include/hazelcast/client/client_config.h
ctc-louk: the new approach reverts the change on client_config.h

hazelcast/include/hazelcast/client/client_config.h
ctc-louk: the new approach reverts the change on client_config.h

hazelcast/include/hazelcast/client/spi/impl/ClientExecutionServiceImpl.h
ctc-louk: the new approach reverts the change

hazelcast/src/hazelcast/client/spi.cpp
ctc-louk: fixed

hazelcast/test/src/HazelcastTests7.cpp
ctc-louk: the new approach reverts the change

hazelcast/test/src/HazelcastTests8.cpp
ctc-louk: added a second TEST_P to ThreadPoolTest

hazelcast/test/src/HazelcastTests8.cpp
ctc-louk: fixed

hazelcast/test/src/HazelcastTests8.cpp
ctc-louk: fixed

ihsandemir (approved):
Thanks. looks great. I will also send a documentation update PR after the PR is merged.

ihsandemir:
Hello @ctc-louk can you rebase and push?

ctc-louk:
Appreciate the timely review. I've rebased to the tip of master (b4dd58f2).

ihsandemir (approved):

PR #1489, opened 2026-09-04
Fix macOS CI: install openssl@3 and derive OPENSSL_ROOT_DIR from Homebrew

ihsandemir:
## Problem

`macos-latest` now resolves to the macOS 26 arm64 image. Every macOS job in `build-pr.yml` and the macOS nightly has been failing in the dependency step since at least 2026-08-30:

```
Warning: No available formula with the name "openssl@1.1". Did you mean openssl@3.5, openssl@3.0, openssl@4 or openssl@3?
Error: Process completed with exit code 1.
```

Two issues:

- Homebrew no longer ships the `openssl@1.1` formula.
- `OPENSSL_ROOT_DIR` was hard-coded to `/usr/local/opt/openssl/`, the Intel Homebrew prefix. On arm64 runners Homebrew lives under `/opt/homebrew`, so CMake would not have found OpenSSL even with a working install.

Example: https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/33849953437

## Change

- Install `openssl@3` instead of `openssl@1.1`.
- Export `OPENSSL_ROOT_DIR` from `brew --prefix openssl@3` in the composite action, so it is correct on both Intel and arm64 runners.
- Remove the static `OPENSSL_ROOT_DIR` env from `build-pr.yml` and `nightly-macos-x86_64.yml`.

The action and workflow names still say x86_64. Left as is to keep the diff to the fix.

## Testing

Nightly macOS workflow dispatched on this branch; link in the comments.

### DIFF
--- a/.github/actions/build-test/macos-x86_64/action.yml
+++ b/.github/actions/build-test/macos-x86_64/action.yml
@@ -31,7 +31,8 @@ runs:
       shell: bash
       run: |
         sudo ./scripts/install-boost.sh ${{ inputs.BOOST_VERSION }}
-        brew install openssl@1.1 thrift curl
+        brew install openssl@3 thrift curl
+        echo "OPENSSL_ROOT_DIR=$(brew --prefix openssl@3)" >> $GITHUB_ENV
 
     - name: Setup JDK
       uses: actions/setup-java@v4
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -214,8 +214,6 @@ jobs:
       id-token: write
 
     name: macOS-(${{ matrix.build_type }}, ${{ matrix.shared_libs.name }}, ${{ matrix.with_openssl.name }})
-    env:
-      OPENSSL_ROOT_DIR: /usr/local/opt/openssl/
 
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/nightly-macos-x86_64.yml
+++ b/.github/workflows/nightly-macos-x86_64.yml
@@ -34,9 +34,6 @@ jobs:
       macOS-x86_64
       (${{ matrix.build_type }}, ${{ matrix.shared_libs.name }}, ${{ matrix.with_openssl.name }}, ${{matrix.boost.version}})
 
-    env:
-      OPENSSL_ROOT_DIR: /usr/local/opt/openssl/
-
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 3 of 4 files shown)


ihsandemir:
Verification run of the macOS nightly on this branch: https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/33851216456

ihsandemir:
The first verification run got past the dependency step but Debug builds then failed because the Xcode 26 clang no longer accepts `-Wno-enum-constexpr-conversion`, which `-Werror` turns into a hard error. Pushed a second commit that probes the flag with `check_cxx_compiler_flag`. New verification run: https://github.com/hazelcast/hazelcast-cpp-client/actions/runs/33851579238

ihsandemir:
Status of the second verification run so far: the dependency step, OpenSSL detection (`/opt/homebrew/opt/openssl@3`) and the Debug build all pass now, and jobs reach the test stage.

One job, `Debug, Static, SSL, 1.88.0`, then aborted in `AutoBatcherTest.concurrencySmokeTest` with an AddressSanitizer stack overflow: a deep recursion of `boost::detail::future_unwrap_shared_state::launch_continuation` -> `do_continuation` -> `mark_finished_internal` in the auto batcher's chained futures. The same test passes locally on an arm64 Debug ASan build against Boost 1.90. This is a product/test issue unrelated to the CI changes here and should be handled separately.

.github/actions/build-test/macos-x86_64/action.yml
```diff
@@ -31,7 +31,8 @@ runs:
       shell: bash
       run: |
         sudo ./scripts/install-boost.sh ${{ inputs.BOOST_VERSION }}
-        brew install openssl@1.1 thrift curl
+        brew install openssl@3 thrift curl
+        echo "OPENSSL_ROOT_DIR=$(brew --prefix openssl@3)" >> $GITHUB_ENV
```
JackPGreen, line 35:
```suggestion
        brew install openssl thrift curl
        echo "OPENSSL_ROOT_DIR=$(brew --prefix openssl)" >> $GITHUB_ENV
```

Avoids the problem for next time.

JackPGreen (changes requested):

.github/actions/build-test/macos-x86_64/action.yml
ihsandemir: do we really want to test on the latest or on a minimum supported version?

JackPGreen: I guess there's pros/cons to both.

JackPGreen (approved):

.github/actions/build-test/macos-x86_64/action.yml
ihsandemir: Well, may be we leave it to fixed like this at this PR and may be we can introduce nightly runs for latest versions if we think they are really needed later on.

